PR #1, opened 2019-10-31
Update code from the internal repo.

lazaroclapp:
* This open sources the PluginCleaner checker
* Batch pulls a number of changes from our internal repo
* Gets us generally ready to making Piranha OSS the source of truth
  overall (including internally).


lazaroclapp:
I did my best to check for references to internal stuff and add the licenses (and the tests run :) ). But a second pair of eyes is always helpful, @mkruber

lazaroclapp:
> Can we skip open sourcing PluginCleaner for now

Done.

piranha/src/main/java/com/uber/piranha/PiranhaUtils.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.uber.piranha;
+
+public class PiranhaUtils {
```
mkruber: This class also needs to be deleted as it is used for internally diff processing.

piranha/src/main/java/com/uber/piranha/UsageCounter.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Originally from com.uber.errorprone.checker.dagger.UsageCounter
+ */
+public final class UsageCounter {
```
mkruber: EP was already planning on a usage checker for Dagger. However, no harm in releasing this.

piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -97,9 +141,17 @@
...
 
+  /** State used to track usage counts and delete corresponding declarations if needed. */
+  private TreePath cuPath = null;
+
+  private boolean countsCollected = false;
+  private ImmutableMap<Symbol, UsageCounter.CounterData> usageCounts = null;
+  private Map<Symbol, Integer> deletedUsages = null;
+
   /**
    * Copied from NullAway comment. Error Prone requires us to have an empty constructor for each
```
mkruber: This line can be deleted :)

piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -298,13 +440,80 @@ private Description updateCode(
...
+      if (fsym.getKind().equals(ElementKind.ENUM_CONSTANT)
+          && fsym.name.toString().toLowerCase().equals(treatmentGroup)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public Description matchClass(ClassTree classTree, VisitorState visitorState) {
```
mkruber: this will work well internally, not sure how this is appropriate for external use. We need to think of a better flow here.

piranha/src/main/java/com/uber/piranha/PiranhaUtils.java
lazaroclapp: That's not exactly practical, though. This class is used by Piranha to mark code for file deletion, so we need to reference it from the EP check. We are not (yet?) providing the scripts to remove files marked with these comments, but we probably should.

We could make the comment text a parameter to Piranha (maybe in a follow up PR?), but I am not sure that would be any cleaner than this.

piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -102,7 +189,91 @@ public void positiveControl() throws IOException {
...
       pe.printStackTrace();
-      fail("Incorrect parameters passed to the checker");
+      assert_().fail("Incorrect parameters passed to the checker");
+    }
+  }
+
+  private BugCheckerRefactoringTestHelper addHelperClasses(BugCheckerRefactoringTestHelper bcr)
+      throws IOException {
+    return bcr.addInputLines(
+            "XPTest.java",
```
mkruber: Can we place them as part of the refactoring files? It is hard to parse them manually and also putting them as part of the refactoring files will ensure consistency and list the respective use cases.

piranha/src/main/java/com/uber/piranha/PiranhaUtils.java
mkruber: providing a script to delete the files with this comment addresses my concern.

lazaroclapp: Ok, will do in a follow up diff, opening an issue for now.

piranha/src/main/java/com/uber/piranha/UsageCounter.java
lazaroclapp: They do have a checker, but it's not implemented with a modular usage counter we can easily call, for now it's better to keep this class around, if not the corresponding `UsageChecker` :)

piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: Seems useful for documenting why we have that constructor, though. I can remove the reference from NullAway if you want, but the comment itself seems useful.

piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: I think I agree, although I am also not sure the external flow needs to be too different (except for using gradle instead of buck and integrations with whatever the users' CI is). For now, I think moving the development to the open can be done first and then we can improve the UX for external adopters :)

PR #5, opened 2019-11-27
Push the source code for PiranhaObjC

mkruber:
- Added code for Piranha ObjC
- Moved code related to PiranhaJava into java


lazaroclapp:
Hey Murali, this is obviously a hard change to review in detail, but I trust you with the Objective-C code ;)

At a glance, though, we are missing the license information on each file :)

On the Java side: I assume you are just `mv`-ing everything from root (other than the `LICENSE.txt` and `.gitignore`) to `java/`? Also, we don't plan to have a single build system for all langs, right? 

If so, how about a global `README.md` which says something like:

> This repository contains three independent versions of Pirahana, one for each of the three supported languages: Objective-C, Swift, and Java.
> 
> To use/build each version, look under the corresponding `[lang]/` directory and follow instructions in the corresponding `[lang]/README.md` file. Make sure to `cd` into that directory to build any related code.
> 
> For example, to build Piranha Java, run:
> 
> ```
> cd java/
> ./gradlew build
> ```

PR #7, opened 2020-01-02
Source for Piranha Swift

mkruber:


swift/src/CleanupStaleFlags/StaleFlagCleanerLauncher.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  Copyright © Uber Technologies, Inc. All rights reserved.
```
lazaroclapp: Shouldn't this be the OSS license instead? And also have the year/year range?

mkruber: will create a new pull request.

PR #15, opened 2020-03-23
Replace leftover fields with values

elsh:


swift/.gitignore
```diff
@@ -0,0 +1,8 @@
+.DS_Store
```
mkruber: Some of this is already ignored at the root level `.gitignore`.

swift/Package.swift
```diff
@@ -1,19 +1,3 @@
-/**
- *    Copyright (c) 2019 Uber Technologies, Inc.
```
elsh: this confuses parsing toolchain version when running swift build

swift/tests/control.swift
```diff
@@ -201,7 +202,7 @@ class SwiftExamples {
     }
 
     // tests for task T2191251
```
mkruber: This comment about task name can be removed.

swift/Package.swift
mkruber: Since this is a generic file, not having the license should be okay.

ketkarameya:
Since there has been no action on this PR for the past 2 years I am closing it.

PR #16, opened 2020-03-23
Fix overlap checking for nodes with same end position.

lazaroclapp:
This solves an annoying off-by-one error in our overlap checking where a parent node and its child can both have the same end position. 

Previously, overlap checking was strict less-than (`<`). Thus, for example `x || !y` will analyze `y` twice, since based on that checking `x || !y` does not overlap with `!y` (they have the same end position, rather than `!y` having an earlier one).

After this chance, we check less-or-equal (`<=`) and thus catch these types of overlaps.


lazaroclapp:
This was literally an afternoon of debugging for a single character fix.

Additionally, we seem to have an issue where `cond()` simplifies to `true` in a conditional like:

```
if (...) {
   ...
} else if (cond()) {
   ...
} else {
   ...
}
```

where the final unreachable branch is not removed.

PR #19, opened 2020-03-25
Fix swift-tools-version in Package.swift

commscheck:
When running `swift build` in the 'swift' subdirectory, I get the following error:

```
/Users/ben/src/third-party/piranha/swift: error: package at '/Users/ben/src/third-party/piranha/swift' is using Swift tools version 3.1.0 which is no longer supported; consider using '// swift-tools-version:5.1' to specify the current tools version
```

The `// swift-tools-version:5.1` declaration needs to be the first line of the Package.swift file to be picked up correctly by Swift tools. I've moved it above the licensing notice which has fixed it for me!

### DIFF
--- a/swift/Package.swift
+++ b/swift/Package.swift
@@ -1,3 +1,6 @@
+// swift-tools-version:5.1
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
 /**
  *    Copyright (c) 2019 Uber Technologies, Inc.
  *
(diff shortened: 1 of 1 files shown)


mkruber:
> When running `swift build` in the 'swift' subdirectory, I get the following error:
> 
> ```
> /Users/ben/src/third-party/piranha/swift: error: package at '/Users/ben/src/third-party/piranha/swift' is using Swift tools version 3.1.0 which is no longer supported; consider using '// swift-tools-version:5.1' to specify the current tools version
> ```
> 
> The `// swift-tools-version:5.1` declaration needs to be the first line of the Package.swift file to be picked up correctly by Swift tools. I've moved it above the licensing notice which has fixed it for me!

Thanks for pointing this out and not sure why the PR was closed. I have included the suggested changes.

PR #33, opened 2020-04-20
Set up travis CI for PiranhaJava and PiranhaSwift

lazaroclapp:
This is just the initial change-set enabling travis for the Java and Swift.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,6 @@ matrix:
       before_script:
         - cd swift/
       script:
-        - swift build
-        - swift package generate-xcodeproj 
         - sh package.sh 
         - sh test.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,8 @@ matrix:
           - $HOME/.gradle
 
     - language: swift
-      osx_image: xcode11.2
+      os: osx
+      osx_image: xcode11.3
       env: SWIFT=5.1
       before_script:
         - cd swift/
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+          - $HOME/.gradle
+
+    - language: swift
+      os: osx
+      osx_image: xcode11.3
+      env: SWIFT=5.1
+      before_script:
+        - cd swift/
+      script:
+        - swift build && swift build && swift package generate-xcodeproj && sh package.sh && sh test.sh
```
mkruber, line 28: this can be changed to `sh package.sh && sh test.sh`? Is there a place to show the console output for this?

mkruber: Discard the second part. I can see the output from the CI logs.

lazaroclapp: It's already running `sh package.sh && sh test.sh` after build and `swift package`. For the output, just click the travis job below: show all checks -> details on the PR one -> The build -> check the Swift one.

mkruber: Correct. But, we don't want the first two expressions. 

`swift build && swift build && swift package generate-xcodeproj && sh package.sh && sh test.sh` -> `sh package.sh && sh test.sh`. 

This is because `swift build` is called from `package.sh` and we don't need to generate a Xcodeproj for running it on CI.

lazaroclapp: Let me try to make the command steps more readable :)

mkruber:
thanks for taking care of this @lazaroclapp

PR #36, opened 2020-04-30
Adding support for control methods with more than 1 argument

sivacoder:
Control methods with more than 1 argument
`getldvalue(STALE_FLAG_NAME, paramter1, parameter 2)`

### DIFF
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -286,7 +286,7 @@ public class XPFlagCleaner extends BugChecker
         return API.UNKNOWN;
       }
 
-      if (mit.getArguments().size() == 1 || mit.getArguments().size() == 2) {
+      if (mit.getArguments().size() >= 1) {
         ExpressionTree arg = mit.getArguments().get(0);
         Symbol argSym = ASTHelpers.getSymbol(arg);
         if (isLiteralTreeAndMatchesFlagName(arg)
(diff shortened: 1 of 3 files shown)


java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -286,7 +286,7 @@ private API getXPAPI(ExpressionTree et) {
         return API.UNKNOWN;
       }
 
-      if (mit.getArguments().size() == 1 || mit.getArguments().size() == 2) {
+      if (mit.getArguments().size() >= 1) {
```
mkruber, line 289: Let us switch the search for arguments and methods i.e., search for the method names first and if they match, check for flag name. Otherwise, I am afraid this operation will become quite expensive. 

Also, better to move all the method HashSets (`controlMethods`, `treatedMethods`, etc) into a HashMap<String,API>.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -396,6 +396,50 @@ public void positiveCaseWithFlagNameAsVariable() throws IOException {
     }
   }
 
+  @Test
+  public void positiveCaseWithMultipleControlArguments() throws IOException {
```
mkruber, line 400: Can you add a negative case also? One for flag name not being stale and another for not matching the method name.

mkr-plse:
@sivacoder Closing this PR as the underlying problem has been resolved in #39. Let me know if it doesn't address your use case.

Issue #38, opened 2020-05-15
PiranhaJava - remove flag methods that have no arguments (that don't use enum)


pranavsb:
My codebase defines toggles like this:
```
public class MockToggle {
    public static boolean isToggleEnabled() {
        return true;
    }
}
```

And they are used like this:
```
public class ApplicationCode {
   if (MockToggle.isToggleEnabled()) {
       // do something
   } else {
       // do something else 
   }
}
```

What's the easiest way to make Piranha work? Should I modify Piranha to consider empty method calls that match the method specified in the properties file? (is it just a matter of changing an `if` condition somewhere?)

I'd like to contribute a PR if I get this working!

mkr-plse:
In the [properties](https://github.com/uber/piranha/blob/master/java/piranha/config/piranha.properties) file, if you include your toggle method in the `treatedMethods`, then depending on the input to Piranha (treated/control), the `then` or `else` branch will be kept.

pranavsb:
Hi! 
I put all possibilities in piranha.properties
`isToggleEnabled,MockToggle.isToggleEnabled,MockToggle`

but it doesn't work.

But I know piranha is configured correctly because dummy branches `if (true)` are getting removed.

mkr-plse:
Ah, I see the problem. There is no flagname for these APIs. You will need to modify the [code](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L289) for the no flag name case. 

A broader fix for this is discussed [here](https://github.com/uber/piranha/issues/6) and [here](https://github.com/uber/piranha/pull/36#pullrequestreview-403458940). A PR contribution for this will be appreciated.

pranavsb:
I got it working by adding a `mit.getArguments().size() == 0` check to `getXPAPI`, will raise a PR soon.

I'm wondering if it makes sense to extend this so that it checks the invocation to make sure its the correct toggle class and not just match by method name. 

Something like this:
piranha.properties
`treatedMethods=MockToggle1.isEnabled`

Application.java
```
if (MockToggle1.isEnabled() {
    // this if should be removed
}
...
if (MockToggle2.isEnabled()) {
   // this one should not be removed
}

```

mkr-plse:
Instead of an interim fix, I suggest creating a json properties file with the following attributes:
1. method name 
2. argument index (0, 1, .. or don't care)
3. return type (boolean, void, ..)
[...]

Issue #6, opened 2019-12-02
Support specifying method and argument index pairs in piranha.properties


lazaroclapp:
This is needed, e.g. to remove methods under `emptyMethods=...` for which the experiment symbol is not the first parameter.

jan25:
Hey @lazaroclapp!
Just to clarify, do you mean like when we have
```
if(expt.flagEnabled(SOME_OTHER_ARG, TestExperimentName.SAMPLE_STALE_FLAG)) {
        System.out.println("Hello World");
}
```
We could specify `emptyMethods=(flagEnabled,1),others..` ?

mkruber:
`emptyMethods` do not return a boolean value. Otherwise, your understanding is correct. 

Currently, the expectation is that the first argument for all the flag APIs correspond to the flag name. This can be relaxed by providing this method, argument index pairs in piranha.properties and updating the implementation to respect this.

lazaroclapp:
Closed by #39

PR #39, opened 2020-05-24
PiranhaJava // config: add returnType, receiverType and argumentIndex; move to properties.json

pranavsb:
This is an enhancement and a fix for #38 and #6 . 
The config file is moved to `properties.json`, removing the necessity for `piranha.properties`.

Return type and receiver type regexes can be specified in `properties.json` corresponding to the `flagType` and `methodName`. Argument index to check against xpFlagName can also be specified. All 3 are optional fields.

*Please see `TODO` in `XPFlagCleanerTest`. I have changed `getXPAPI` to return `API.UNKNOWN` if  the specified method's receiver or return type don't match what is specified in `properties.json`, but I think there is some change required in some other part of the code.*
UPDATE: Have raised #40 for any further discussion related to this.

~NOTE: Argument index is not yet implemented, so #6 is still open.~


java/README.md
```diff
@@ -49,17 +49,36 @@ In the `tasks.withType(JavaCompile)` section, we pass some configuration options
...
+  "piranhaMethodProperties":
+    [
+      {
+        "methodName": "isToggleEnabled",
+        "flagType": "treated",
+        "returnType": "boolean",
+        "receiverType": "com.uber.piranha.XPTest"
+      },
+      {
+        "methodName": null,
```
mkr-plse, line 62: why the presence of `nulls` here?

java/piranha/config/properties.json
```diff
@@ -0,0 +1,46 @@
...
+    {
+      "methodName": "includeEvent",
+      "flagType": "empty"
+    },
+    {
+      "methodName": "isToggleInGroup",
+      "flagType": "treatmentGroup"
+    },
+    {
+      "methodName": null,
```
mkr-plse, line 35: don't follow the need for the set of `null` here.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -139,12 +149,14 @@
...
-  /* information provided in the config file */
-  private final HashSet<String> treatedMethods = new HashSet<String>();
-  private final HashSet<String> controlMethods = new HashSet<String>();
-  private final HashSet<String> deleteMethods = new HashSet<String>();
-  private final HashSet<String> treatmentGroupMethods = new HashSet<String>();
-  private final HashSet<String> handledAnnotations = new HashSet<String>();
+  /** Information provided in the properties.json config file */
+  private final Set<Map<String, Object>> methodProperties = new HashSet<>();
+
+  private final Map<String, List<Map<String, Object>>> treatedMethods = new HashMap<>();
```
mkr-plse, line 155: Can you reduce the number of maps here to one? The earlier code also had these maps unnecessarily.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -396,13 +396,120 @@ public void positiveCaseWithFlagNameAsVariable() throws IOException {
     }
   }
 
+  @Test
+  public void caseWithFlagNameAsVariableWithReturnType() throws IOException {
+    /* todo - piranha deletes the line
```
mkr-plse, line 401: given the flexibiity with the APIs, Piranha will need one more pass to delete the variable declarations [here](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L710). While this is a good test to check for return type based deletion, this may be not happen in practice.

java/piranha/src/test/resources/config/properties_test_receive.json
```diff
@@ -0,0 +1,19 @@
...
+  "piranhaMethodProperties": [
+    {
+      "methodName": "isToggleDisabled",
+      "flagType": "control",
+      "receiverType": "com\\.uber\\.piranha\\.XPTest"
+    },
+    {
+      "methodName": "isFlagTreated",
+      "flagType": "treated",
+      "receiverType": "com.uber.*"
```
mkr-plse, line 11: where is this used?

java/piranha/config/properties.json
pranavsb: The last set is just for illustrative purposes. Do you think it's better to replace `null` with empty strings `""` or just remove that set altogether?

java/README.md
pranavsb: See my comment on `properties.json`: https://github.com/uber/piranha/pull/39#discussion_r430159803

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
pranavsb: 👍 it will make the data structure one level deeper but makes sense.

java/piranha/src/test/resources/config/properties_test_receive.json
pranavsb: `properties_test_receive.json` and `properties_test_return.json` are used in `XPFlagCleanerTest`

If you were asking about this particular testcase:
`caseWithFlagNameAsStringLiteralWithReceiverTypeRegex`

java/piranha/build.gradle
```diff
@@ -30,6 +30,7 @@ dependencies {
     annotationProcessor deps.apt.autoService
 
     compileOnly deps.build.errorProneCheckApi
+    implementation deps.build.googleJsonSimple
```
pranavsb, line 33: I'm new to gradle. Can you double check that `implementation` is the correct thing to use here?

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: Hmm, I'm not sure I followed. In this case, Piranha is deleting a variable declaration that is still in use. Is there an easy way to fix this?

pranavsb:
> 
> 
> Thanks for an excellent enhancement PR. Overall, looks great.
> Can you clarify some of the questions raised in the review?

Thank you. I've replied to your comments and added one of my own. 

Will change:
1) all maps to one map - for holding config data
2) `null` to `""`

java/piranha/config/properties.json
mkr-plse: I suggest that it be changed to `...` from line 34, as there is no requirement to have a `null` or empty string at the end of the array.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
mkr-plse: In the original implementation, all flag API invocations independent of their receiver/return types would be deleted and therefore deleting the flag deletion was not a problem. This is likely to be the case in real codebases also. 

However, for the test case here, given the differentiating functionality is being tested, the variable decl (which doesn't look at the uses) is incorrectly being deleted. I am okay with this change. Can you file an issue for this TODO? It has to be fixed by doing a two pass analysis. In the first pass, delete all the uses and if there are no uses for the variable, delete the variable in the second pass.

java/piranha/src/test/resources/config/properties_test_receive.json
mkr-plse: i was asking for this specific case. thanks.

java/sample/config/properties.json
```diff
@@ -0,0 +1,28 @@
+{
+  "piranhaMethodProperties": [
+    {
+      "methodName": "flagDisabled",
```
mkr-plse, line 4: how is the `flagDisabled("some_flag")` differentiated from `flagDisabled()`? Even if argument indices are not handled in this PR, I think it is important to handle the differential when  the flagname is one of the arguments and when it is not.

java/piranha/config/properties.json
pranavsb: :+1: changing to `...` in README.md and deleting in the actual `properties.json` as it has to be JSON formatted.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: :+1:

java/sample/config/properties.json
pranavsb: I've added a check under `if (mit.getArguments().size()==0)` (it comes around here: https://github.com/uber/piranha/blob/dd050f86315ab3e719fa5d22d7f0039ad844b039/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L289) 

that does this:
```
if (mit.getArguments().isEmpty()) {
          if (xpFlagName.equals(EMPTY)) {
            return getXPAPIFromMIT(mit);
          } else {
            return API.UNKNOWN;
          }
   }
```

So a user would have to explicitly run Piranha twice (with different flag arguments) to delete both`flagDisabled("some_flag")` and `flagDisabled()`

I've updated the relevant testcase as well.

java/piranha/build.gradle
pranavsb: Oh and don't forget this one!

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -285,30 +369,105 @@ private API getXPAPI(ExpressionTree et) {
...
+              || isSymbolAndMatchesFlagName(argSym)) {
+            return getXPAPIFromMIT(mit);
+          }
+        }
+      }
+    }
+    return API.UNKNOWN;
+  }
+
+  private API getXPAPIFromMIT(MethodInvocationTree mit) {
```
lazaroclapp, line 393: I very well might be missing something, but this looks both like more code _and_ a worse runtime hit than just keeping the separate maps.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -396,13 +396,112 @@ public void positiveCaseWithFlagNameAsVariable() throws IOException {
...
+              "     else { return \"Y\";}",
+              " }",
+              "}")
+          .addOutputLines(
+              "XPFlagCleanerSinglePositiveCase.java",
+              "package com.uber.piranha;",
+              "class XPFlagCleanerSinglePositiveCase {",
+              " private XPTest experimentation;",
+              " public String evaluate() {",
+              "  if (experimentation.isToggleEnabled(STALE_FLAG_CONSTANTS)) { int a = 1; }",
```
lazaroclapp, line 434: Probably worth a comment as to why this isn't simplified. I get that the `returnType` is not marked as `bool.*` (also: any reason for that vs `boolean`?), but I am not sure I'll remember that when looking at this test case in isolation years down the line ;)

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -526,4 +625,199 @@ public void negativeCaseWithFlagNameAsVariable() throws IOException {
...
+              "     else { return \"Y\";}",
+              " }",
+              "}")
+          .addOutputLines(
+              "XPFlagCleanerSinglePositiveCase.java",
+              "package com.uber.piranha;",
+              "class XPFlagCleanerSinglePositiveCase {",
+              " private XPTest experimentation;",
+              " public String evaluate() {",
+              "  if (experimentation.isToggleEnabled()) { int a = 1; }",
```
lazaroclapp, line 764: Same as above in the "yes flag" case.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -526,4 +625,199 @@ public void negativeCaseWithFlagNameAsVariable() throws IOException {
...
+      bcr.setArgs("-d", temporaryFolder.getRoot().getAbsolutePath());
+
+      bcr = addHelperClasses(bcr);
+      bcr.addInputLines(
+              "XPFlagCleanerSinglePositiveCase.java",
+              "package com.uber.piranha;",
+              "class XPFlagCleanerSinglePositiveCase {",
+              " private XPTest experimentation;",
+              " public String evaluate() {",
+              "  // BUG: Diagnostic contains: Cleans stale XP flags",
```
lazaroclapp, line 698: Should this produce an error here if it's not going to refactor anything?

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -526,4 +625,199 @@ public void negativeCaseWithFlagNameAsVariable() throws IOException {
...
+              "     else { return \"Y\";}",
+              " }",
+              "}")
+          .addOutputLines(
+              "XPFlagCleanerSinglePositiveCase.java",
+              "package com.uber.piranha;",
+              "class XPFlagCleanerSinglePositiveCase {",
+              " private XPTest experimentation;",
+              " public String evaluate() {",
+              "  // BUG: Diagnostic contains: Cleans stale XP flags",
```
lazaroclapp, line 713: Same.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -526,4 +625,199 @@ public void negativeCaseWithFlagNameAsVariable() throws IOException {
...
+      bcr.setArgs("-d", temporaryFolder.getRoot().getAbsolutePath());
+
+      bcr = addHelperClasses(bcr);
+      bcr.addInputLines(
+              "XPFlagCleanerSinglePositiveCase.java",
+              "package com.uber.piranha;",
+              "class XPFlagCleanerSinglePositiveCase {",
+              " private XPTest experimentation;",
+              " public String evaluate() {",
+              "  // BUG: Diagnostic contains: Cleans stale XP flags",
```
lazaroclapp, line 751: And, again, should this raise an error here? In any case it would be the conditional below and not this one, wouldn't it?

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -526,4 +625,199 @@ public void negativeCaseWithFlagNameAsVariable() throws IOException {
...
```
lazaroclapp, line 798: Same.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -139,12 +153,12 @@
...
-  private final HashSet<String> treatedMethods = new HashSet<String>();
-  private final HashSet<String> controlMethods = new HashSet<String>();
-  private final HashSet<String> deleteMethods = new HashSet<String>();
-  private final HashSet<String> treatmentGroupMethods = new HashSet<String>();
-  private final HashSet<String> handledAnnotations = new HashSet<String>();
+  /** Information provided in the properties.json config file */
+  private final Set<Map<String, Object>> methodProperties = new HashSet<>();
+
+  private final Map<String, Map<String, List<Map<String, Object>>>> configMethodsMap =
+      new HashMap<>();
```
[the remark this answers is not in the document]
mkr-plse: I think this can be a map from methodname to a set of attributes. Something like: 

methodX -> {(treated, bool, com.uber.*)}
methodY -> {(control, bool, com.uber.*), (control, bool, some_other_receiver), ...}

The idea is to do one hashmap lookup instead of four different lookups for each method call. In most cases, there won't be a match and so this will be faster.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
mkr-plse: See my comment above on the structure of the HashMap. This will simplify to one `containsKey` on `configMethodsMap`.

java/sample/config/properties.json
mkr-plse: This will be problematic. Why not simply add an additional argument index to the json file, and if the argument index is not provided, flagname need not be checked. Otherwise, for the given index, the flag name is checked.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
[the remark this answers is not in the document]
lazaroclapp: This is not currently true. The first key to `configMethodsMap` is the flag type. See [these calls](https://github.com/uber/piranha/pull/39/files#diff-ddde662a04227b8ea9bef9fc02110791R327) and the target method implementation.

If we do go with the approach you are proposing (which makes sense), then: a) the top level key would need to be the method, b) I'd change `Map<String, List<Map<String, Object>>>` to `ImmutableList<PiranhaMethodRecord>` or something like that, where `PiranhaMethodRecord` is some sort of typed config record with fields like `.flagType`, `.receiverType`, and so on.

Btw, extra pedantic nit: I'd prefer `ImmutableMap` and `ImmutableList` here and builders. But since the original code was also using mutable collections, I am absolutely fine leaving that for a follow up issue you two can assign to me 😅

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: That's not the current structure, though, check the four constants being looked up below.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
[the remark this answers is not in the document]
mkr-plse: +1.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -180,14 +194,23 @@ private void init(ErrorProneFlags flags) throws ParseException {
...
       try {
-        Properties prop = new Properties();
-        prop.load(Files.newBufferedReader(Paths.get(configFile), Charset.defaultCharset()));
-        updateConfig(prop, "treatedMethods", treatedMethods);
-        updateConfig(prop, "controlMethods", controlMethods);
-        updateConfig(prop, "emptyMethods", deleteMethods);
-        updateConfig(prop, "treatmentGroupMethods", treatmentGroupMethods);
-        updateConfig(prop, "annotations", handledAnnotations);
-        linkURL = prop.getProperty("linkURL");
+        JSONParser jsonParser = new JSONParser();
```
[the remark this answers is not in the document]
mkr-plse: After discussions with @lazaroclapp, let us skip on this for now. We can just have one json file for configuration.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
mkr-plse: Agreed. let us revamp the map structure as discussed above. 

For a majority of the functions, the `containsKey` will be a false. For the few FF APIs that match, we can evaluate their attributes (methodtype, return type, receiver type, etc) and take appropriate action.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
mkr-plse: +1. useful to have additional comments for all the tests that are added here.

pranavsb:
@mkr-plse @lazaroclapp 

All good points, will get back to this over the weekend and wrap it up.

Wanted to bring one thing to your attention though.
I think some gradle changes are required to ensure that json jar is contained within the piranha jar generated by `gradle jar`. I tried copying the 0.0.4-SNAPSHOT jar to my local maven repository and compiled a dummy springboot project and got this error. Used to compile and have in place piranha updates before.


```
Error: An unhandled exception was thrown by the Error Prone static analysis plugin.
[ERROR] Please report this at https://github.com/google/error-prone/issues/new and include the following:
[ERROR]
[ERROR] error-prone version: 2.3.4
[ERROR] BugPattern: Piranha
[ERROR] Stack Trace:
[ERROR] java.lang.NoClassDefFoundError: org/json/simple/parser/JSONParser
[ERROR] -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
```

lazaroclapp:
> @mkr-plse @lazaroclapp
> 
> All good points, will get back to this over the weekend and wrap it up.
> 
> Wanted to bring one thing to your attention though.
> I think some gradle changes are required to ensure that json jar is contained within the piranha jar generated by `gradle jar`. I tried copying the 0.0.4-SNAPSHOT jar to my local maven repository and compiled a dummy springboot project and got this error. Used to compile and have in place piranha updates before.
> 
> ```
> Error: An unhandled exception was thrown by the Error Prone static analysis plugin.
> [ERROR] Please report this at https://github.com/google/error-prone/issues/new and include the following:
> [ERROR]
> [ERROR] error-prone version: 2.3.4
> [ERROR] BugPattern: Piranha
> [ERROR] Stack Trace:
> [ERROR] java.lang.NoClassDefFoundError: org/json/simple/parser/JSONParser
> [ERROR] -> [Help 1]
> [ERROR]
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
> [ERROR]
> ```

How are you bringing in Piranha? Are you making it a dependency for your project? (If so, I believe the build system would grab its dependencies from maven central) Or are you just sticking the jar somewhere? (Piranha is not supposed to be a fat jar, afaik)

Note that the sample projects in this repo build by simply depending on Piranha and we'd be sure to test this with our internal (buck) build system before cutting a release.

What happens if you replace `-SNAPSHOT` with `.1-LOCAL`, do `./gradlew install` (on the `java/` directory) then import in your project from the local maven cache?

pranavsb:
> How are you bringing in Piranha? Are you making it a dependency for your project? (If so, I believe the build system would grab its dependencies from maven central) Or are you just sticking the jar somewhere? (Piranha is not supposed to be a fat jar, afaik)

Yes, Piranha is added as a dependency. Basically the same POM config that's given in the [Java README](https://github.com/uber/piranha/blob/master/java/README.md). I figured out that adding `json-simple` to `annotationProcessorPaths` (that already contains piranha and errorprone) works. 


> Note that the sample projects in this repo build by simply depending on Piranha and we'd be sure to test this with our internal (buck) build system before cutting a release.

 :+1: 

> What happens if you replace `-SNAPSHOT` with `.1-LOCAL`, do `./gradlew install` (on the `java/` directory) then import in your project from the local maven cache?

This works too! Must have been some issue on my end.

lazaroclapp:
> > How are you bringing in Piranha? Are you making it a dependency for your project? (If so, I believe the build system would grab its dependencies from maven central) Or are you just sticking the jar somewhere? (Piranha is not supposed to be a fat jar, afaik)
> 
> Yes, Piranha is added as a dependency. Basically the same POM config that's given in the [Java README](https://github.com/uber/piranha/blob/master/java/README.md). I figured out that adding `json-simple` to `annotationProcessorPaths` (that already contains piranha and errorprone) works.
> 
> > Note that the sample projects in this repo build by simply depending on Piranha and we'd be sure to test this with our internal (buck) build system before cutting a release.
> 
> 👍
> 
> > What happens if you replace `-SNAPSHOT` with `.1-LOCAL`, do `./gradlew install` (on the `java/` directory) then import in your project from the local maven cache?
> 
> This works too! Must have been some issue on my end.

I think what is happening is, because you are moving the `SNAPSHOT` jar around, it was trying to use Piranha after this patch but only fetching the dependencies of the latest Piranha release, which doesn't include `json-simple` (you shouldn't need to add it to `annotationProcessorPaths` once this PR becomes part of mainline Piranha, and I should test that that is true before cutting a release :) )

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: Done! 

>  I get that the returnType is not marked as bool.* (also: any reason for that vs boolean?)

Since I've used regex matching for both return and receiver type, just thought a test making it explicit is worth having.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: Looking at the code, seems like it's supposed to 🤔 
But locally I don't get any error, irrespective of that line (this is true for all of my tests btw)

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: > In any case it would be the conditional below and not this one, wouldn't it?

Yup

java/sample/config/properties.json
pranavsb: Added argumentIndex to properties.json and discarded this temporary fix.

lazaroclapp:
> Anything else remaining?
> Not sure about the testcases with `//BUG: `
> They pass for me, with and without the line, though I see from the code that maybe that's not supposed to happen.

Haven't done the full review yet, but two things:

1. Let's remove the `//BUG: ` lines on the _new_ refactoring test cases (no need to change the existing ones), if they aren't being checked. As long as we are checking input and output code, it should be good. Does that make sense to you as well, @mkr-plse ?
2. Please `git pull` to rebase atop master and resolve the conflicts with other recent PRs.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -285,30 +336,110 @@ private API getXPAPI(ExpressionTree et) {
...
+                mit, RETURN_TYPE_STRING, (String) currentMethodProperties.get(RETURN_TYPE_STRING))
+            && checkMethodTreeMatchesMethodProperty(
+                mit,
+                RECEIVER_TYPE_STRING,
+                (String) currentMethodProperties.get(RECEIVER_TYPE_STRING))) {
+          return matchAPIToFlagTypeString((String) currentMethodProperties.get(FLAG_TYPE_KEY));
+        }
+      } else if (returnSpecified || receiverSpecified) {
+        if (returnSpecified) {
+          if (checkMethodTreeMatchesMethodProperty(
```
mkr-plse, line 391: Can we avoid the additional method call here and inline the code.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -285,30 +336,110 @@ private API getXPAPI(ExpressionTree et) {
...
+        }
+      } else {
+        return matchAPIToFlagTypeString((String) currentMethodProperties.get(FLAG_TYPE_KEY));
+      }
     }
     return API.UNKNOWN;
   }
 
+  private API matchAPIToFlagTypeString(String flagType) {
+    if (TREATED_KEY.equalsIgnoreCase(flagType)) {
```
mkr-plse, line 411: `HashMap<flagType, enum>` will simplify this method implementation.

mkr-plse:
> 1. Let's remove the `//BUG: ` lines on the _new_ refactoring test cases (no need to change the existing ones), if they aren't being checked. As long as we are checking input and output code, it should be good. Does that make sense to you as well, @mkr-plse ?

Agree.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -139,12 +157,11 @@
...
-  /* information provided in the config file */
-  private final HashSet<String> treatedMethods = new HashSet<String>();
-  private final HashSet<String> controlMethods = new HashSet<String>();
-  private final HashSet<String> deleteMethods = new HashSet<String>();
-  private final HashSet<String> treatmentGroupMethods = new HashSet<String>();
-  private final HashSet<String> handledAnnotations = new HashSet<String>();
+  /** Information provided in the properties.json config file */
+  private final Set<Map<String, Object>> methodProperties = new HashSet<>();
+
+  private final Map<String, List<Map<String, Object>>> configMethodsMap = new HashMap<>();
```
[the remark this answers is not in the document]
pranavsb: Done.

```
/** Information provided in the properties.json config file.
   *
   * configMethodsMap is a map where key is method name and value is a list where
   * each item in the list is a map that corresponds to each method property from properties.json.
   * In most cases, the list would have only one element.
   * But if someone reuses the same method name with different returnType/receiverType/argumentIndex,
   * the list would have each method property map as one element.
   * */
```
I know it's not ideal but it's a start.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -918,4 +1049,23 @@ private boolean endsWithReturn(StatementTree stmt) {
...
+
+  private boolean checkValueStringExistsAndNotEmpty(Map<String, Object> map, String key) {
+    return map.containsKey(key)
+        && map.get(key) != null
+        && map.get(key) instanceof String
+        && !map.get(key).equals(EMPTY);
+  }
+
+  private boolean checkValueIntOrLongExistsAndIsValid(Map<String, Object> map, String key) {
+    boolean nonNullValue = map.containsKey(key) && map.get(key) != null;
```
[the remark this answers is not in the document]
pranavsb: Simplified.

The JSONParser returns Long, so we need to `((Long) value).intValue() >= ZERO;`
Checking for Integer just to be on the safe side.
```
private boolean checkArgumentIndexExistsAndIsValid(Map<String, Object> map) {
    Object value = map.get(ARGUMENT_INDEX_STRING);
    if (value instanceof Long) {
      return ((Long) value).intValue() >= ZERO;
    } else if (value instanceof Integer) {
      return ((Integer) value) >= ZERO;
    }
    return false;
  }
```

java/piranha/config/properties.json
```diff
@@ -0,0 +1,45 @@
...
+      "argumentIndex": 0
+    },
+    {
+      "methodName": "putToggleDisabled",
+      "flagType": "empty",
+      "argumentIndex": 0
+    },
+    {
+      "methodName": "includeEvent",
+      "flagType": "empty"
```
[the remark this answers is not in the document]
pranavsb: Is there a "flagName" somewhere?

I've only considered the following
mandatory: "flagType" and "methodName"
optional: "returnType", "receiverType" and "argumentIndex"

Any key other than these will not be considered.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -285,30 +336,110 @@ private API getXPAPI(ExpressionTree et) {
...
+      return API.UNKNOWN;
+    }
+  }
+
+  private boolean checkMethodTreeMatchesMethodProperty(
+      MethodInvocationTree mit, String propKey, String propValue) {
+    MemberSelectTree mst = ((MemberSelectTree) mit.getMethodSelect());
+    if (RETURN_TYPE_STRING.equals(propKey)) {
+      String mReturn = ASTHelpers.getReturnType(mst).toString();
+      return Pattern.compile(propValue, Pattern.CASE_INSENSITIVE).matcher(mReturn).find();
```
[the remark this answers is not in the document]
pranavsb: Good catch. But I'm not really sure how much we can pre-compute it.

**Why this case is rare (enough)**
It is triggered only when `returnType` or `receiverType` is specified - for method names that are specified in properties.json

The `for` loop with currentMethodProperties is only triggered once in the vast majority of cases. With the `continue` refactoring, `checkMethodTreeMatchesMethodProperty` is only called once if either return or receiver are specified and twice if both are.

**Why it's difficult to easily speedup**
The `mReturn/mReceive` can be obtained only when mit/mst is passed in. That means only the `Pattern.compile(propValue, Pattern.CASE_INSENSITIVE)` can be pre-computed, but even this depends on the value given for `returnType` or `receiverType`, which may or may not be unique for a given config file. We might get some speedup by precomputing all `Pattern.compile("boolean", Pattern.CASE_INSENSITIVE)` for example.

To do that, we would have to iterate over the config file after parsing it and compute all `Pattern.compile` if return or receiver is specified. (These would then be stored in a Map<String, Pattern.Compile> somewhere?)

But I think this may not provide that much of a speedup and seems more confusing code-wise.

Let me know if I've missed something or if you disagree.

pranavsb:
>    Let's remove the //BUG: lines on the new refactoring test cases
>  Please git pull to rebase atop master and resolve the conflicts with other recent PRs.

Done!

I also closed almost all review comments. Have left a few open in case there's further discussion required.

One thing worth noting, I've changed the `Files.newBufferedReader` to use `StandardCharsets.UTF_8` instead of the `Charset.defaultCharset()` call earlier.

java/piranha/config/properties.json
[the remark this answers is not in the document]
mkr-plse: I meant since the argumentIndex is not specified here, the flagName that is provided as input will not be considered for comparison.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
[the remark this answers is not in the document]
mkr-plse: Thanks for the detailed response. This just got me thinking that we really don't need a pattern for either returnType or receiverType. These can be exact fully qualified names that can be specified in the json and the comparison can be with that. The flexibility of having a pattern makes the implementation (and documentation) complex but doesn't yield any additional benefits given the user has to specify methodName and argumentIndex precisely. Can we get rid of the pattern for these types and simplify the code, and update the documentation for this?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -96,11 +103,26 @@
       ImmutableSet.of("control", "enabled", "disabled", "treatment", "treated");
 
   private static final int DONTCARE = -1;
```
mkr-plse, line 105: delete unused field?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -920,4 +1042,34 @@ private boolean endsWithReturn(StatementTree stmt) {
...
+
+  /**
+   * Utility method. Checks whether the value associated to a given map and given key is a non-empty
+   * string.
+   *
+   * @param map - map corresponding to a method property
+   * @param key - key to check the corresponding value
+   * @return true if value is a non-empty string
+   */
+  private boolean checkValueStringExistsAndNotEmpty(Map<String, Object> map, String key) {
```
mkr-plse, line 1054: If we change the return type of this method from `boolean` to `String`, and return `null` when `false` is being returned, we can avoid more calls to the map subsequently. This can be `getValueStringFromMap`.  
The callers can then be simplified.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -278,6 +316,29 @@ private void updateConfig(Properties prop, String key, HashSet<String> hs) {
...
   }
 
+  private void addMethodPropertiesToConfigMethodProperties(
+      Set<Map<String, Object>> methodProperties) {
+    Iterator<Map<String, Object>> methodPropsIterator = methodProperties.iterator();
+    while (methodPropsIterator.hasNext()) {
+      Map<String, Object> methodProperty = methodPropsIterator.next();
+      if (checkValueStringExistsAndNotEmpty(methodProperty, FLAG_TYPE_KEY)
+          && checkValueStringExistsAndNotEmpty(methodProperty, METHOD_NAME_KEY)) {
+        addMethodPropertyToConfigMethodProperties(methodProperty);
```
mkr-plse, line 326: given the update to `checkValueStringExistsAndNotEmpty`, this code can be simplified further. Also, 
`addMethodPropertyToConfigMethodProperties` is only called here and can be confusing with the current method `addMethodPropertiesToConfigMethodProperties`. Can we inline `addMethodPropertyToConfigMethodProperties`?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -278,6 +316,29 @@ private void updateConfig(Properties prop, String key, HashSet<String> hs) {
...
+      }
+    }
+  }
+
+  private void addMethodPropertyToConfigMethodProperties(Map<String, Object> methodProperty) {
+    String methodNameKey = (String) methodProperty.get(METHOD_NAME_KEY);
+    if (configMethodProperties.containsKey(methodNameKey)) {
+      configMethodProperties.get(methodNameKey).add(methodProperty);
```
mkr-plse, line 334: avoid `containsKey` as `get` should provide the required data.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +348,92 @@ private API getXPAPI(ExpressionTree et) {
       if (!mit.getMethodSelect().getKind().equals(Kind.MEMBER_SELECT)) {
         return API.UNKNOWN;
       }
+      MemberSelectTree mst = (MemberSelectTree) mit.getMethodSelect();
+      String methodName = mst.getIdentifier().toString();
+      List<Map<String, Object>> methodPropertiesForMethodName =
+          configMethodProperties.get(methodName);
+      if (methodPropertiesForMethodName != null) {
+        return getAPIFromMethodPropsAndMIT(mit, methodPropertiesForMethodName);
```
mkr-plse, line 356: This can also be `getXPAPI` but with different parameters.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +348,92 @@ private API getXPAPI(ExpressionTree et) {
...
-            return API.IS_TREATED;
-          } else if (deleteMethods.contains(methodName)) {
-            return API.DELETE_METHOD;
-          } else if (treatmentGroupMethods.contains(methodName)) {
-            return API.IS_TREATMENT_GROUP_CHECK;
+  private API getAPIFromMethodPropsAndMIT(
+      MethodInvocationTree mit, List<Map<String, Object>> methodPropertiesForMethodName) {
+    for (Map<String, Object> currentMethodProperties : methodPropertiesForMethodName) {
+      boolean returnSpecified =
+          checkValueStringExistsAndNotEmpty(currentMethodProperties, RETURN_TYPE_STRING);
```
mkr-plse, line 366: return will be a String and will require updates.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +348,92 @@ private API getXPAPI(ExpressionTree et) {
...
+      boolean receiverSpecified =
+          checkValueStringExistsAndNotEmpty(currentMethodProperties, RECEIVER_TYPE_STRING);
+      boolean argumentIndexSpecified = checkArgumentIndexExistsAndIsValid(currentMethodProperties);
```
mkr-plse, line 369: instead of `checkArgumentIndex` returning `boolean`, return the appropriate object. Then, we can avoid line 374.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +348,92 @@ private API getXPAPI(ExpressionTree et) {
...
-          } else if (treatedMethods.contains(methodName)) {
-            return API.IS_TREATED;
-          } else if (deleteMethods.contains(methodName)) {
-            return API.DELETE_METHOD;
-          } else if (treatmentGroupMethods.contains(methodName)) {
-            return API.IS_TREATMENT_GROUP_CHECK;
+  private API getAPIFromMethodPropsAndMIT(
+      MethodInvocationTree mit, List<Map<String, Object>> methodPropertiesForMethodName) {
+    for (Map<String, Object> currentMethodProperties : methodPropertiesForMethodName) {
+      boolean returnSpecified =
```
mkr-plse, line 365: move the two calls to 365 And 367 to after 384

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +348,92 @@ private API getXPAPI(ExpressionTree et) {
...
+        } else {
+          continue;
         }
       }
+      MemberSelectTree mst = ((MemberSelectTree) mit.getMethodSelect());
+      // when returnType regex is specified, check if mst's return type matches it
+      // if it's not a match, skip to next method property map
+      if (returnSpecified) {
+        if (!isMethodTreeMatchesMethodProperty(
+            mst, RETURN_TYPE_STRING, (String) currentMethodProperties.get(RETURN_TYPE_STRING))) {
```
mkr-plse, line 390: `get` can be avoided once we have the exact string.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +348,92 @@ private API getXPAPI(ExpressionTree et) {
...
+          continue;
+        }
+      }
+      // when receiverType regex is specified, check if mst's receiver type matches it
+      // if it's not a match, skip to next method property map
+      if (receiverSpecified) {
+        if (!isMethodTreeMatchesMethodProperty(
+            mst,
+            RECEIVER_TYPE_STRING,
+            (String) currentMethodProperties.get(RECEIVER_TYPE_STRING))) {
```
mkr-plse, line 400: `get` can be avoided here too.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -920,4 +1042,34 @@ private boolean endsWithReturn(StatementTree stmt) {
...
+  }
+
+  /**
+   * Utility method. Checks whether the argumentIndex key of a method property map is a non-negative
+   * integer.
+   *
+   * @param map - map corresponding to a method property
+   * @return true if argument index is a non-negative integer
+   */
+  private boolean checkArgumentIndexExistsAndIsValid(Map<String, Object> map) {
```
mkr-plse, line 1066: return the object instead of boolean?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -920,4 +1042,34 @@ private boolean endsWithReturn(StatementTree stmt) {
...
+    Object value = map.get(ARGUMENT_INDEX_STRING);
+    if (value instanceof Long) {
+      return ((Long) value).intValue() >= ZERO;
+    } else if (value instanceof Integer) {
+      return ((Integer) value) >= ZERO;
```
mkr-plse, line 1071: Still don't see the need for checking `Integer`.

java/piranha/src/test/resources/config/properties_test_argument.json
```diff
@@ -0,0 +1,18 @@
+{
+  "piranhaMethodProperties": [
+    {
+      "methodName": "isToggleDisabled",
+      "flagType": "control"
+    },
+    {
+      "methodName": "isFlagTreated",
+      "flagType": "treated",
+      "argumentIndex": 0
```
mkr-plse, line 10: can we add a check where argumentIndex is either 1 or 2?

java/README.md
```diff
@@ -49,17 +49,35 @@ In the `tasks.withType(JavaCompile)` section, we pass some configuration options
...
 
 ```
-treatedMethods=treated,flagEnabled
-controlMethods=flagDisabled
-emptyMethods=enableFlag,disableFlag
-treatmentGroupMethods=isToggleInGroup
-annotations=FlagTesting
-linkURL=<provide_your_url>
+{
+  "piranhaMethodProperties":
```
mkr-plse, line 53: Let us call this `methodProperties`.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
pranavsb: It's an existing field that's used. I didn't introduce it.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
pranavsb: Might be required if different parser is used eg. when config support is added for YAML.
Removing it for now.

java/piranha/src/test/resources/config/properties_test_argument.json
pranavsb: Good catch.

pranavsb:
> Thanks for the changes. Left a few more comments, but at the high level, these are the main issues:
> a) We can get rid of the pattern for receiverType and returnType and simplify the code

Done.

> b) We can change the return type of `checkValue...` to the appropriate Object. Will avoid redundant `get` calls.

Done.

> c) The tests may need to be enhanced to check for a few more combinations. Some cases that I can think of:

 > for each of the ~above~ below cases, there are likely two tests -- with or without method matching in the input source

### Testcases

- [ ] when the required properties are not provided in the json,
- [ ] when receiverType alone is specified
- [ ] when returnType alone is specified
- [ ] when argumentIndex alone is specified
- [ ] when argumentIndex and receiverType are specified
- [ ] when argumentIndex and returnType are specified
- [ ] when all the three are specified

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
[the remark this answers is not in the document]
pranavsb: Sounds good. I think there may be some use to having pattern match for receiver, but its a rare enough that it may not be worth the complexity.

mkr-plse:
> > Thanks for the changes. Left a few more comments, but at the high level, these are the main issues:
> > a) We can get rid of the pattern for receiverType and returnType and simplify the code
> 
> Done.

@pranavsb Can you push your changes? It looks like the files have not been updated yet.

pranavsb:
> 
> 
> > > Thanks for the changes. Left a few more comments, but at the high level, these are the main issues:
> > > a) We can get rid of the pattern for receiverType and returnType and simplify the code
> > 
> > 
> > Done.
> 
> @pranavsb Can you push your changes? It looks like the files have not been updated yet.

Was planning to finish tests before pushing. Have added a few. Let me push for now.

mkr-plse:
> Was planning to finish tests before pushing. Have added a few.
> ~Let me push for now.~ Pushed.

I thought you forgot to push because of the done comment. No hurry. Let me know when the code is available for review again.

pranavsb:
Added (lots of) tests and pushed. Guess it's in decent shape now.

java/README.md
```diff
@@ -49,17 +49,35 @@ In the `tasks.withType(JavaCompile)` section, we pass some configuration options
...
+
+The `flagType` with `empty` specifies the APIs which need to be discarded from the code. For example, a statement `enableFlag(SAMPLE_STALE_FLAG);` will be deleted from the code. 
+
+The optional field `argumentIndex` specifies where to look for the flag name (given by `-XepOpt:Piranha:FlagName`) in the method's arguments. We follow 0 based indexing.
+If your toggle method takes no arguments, or if you want to delete all occurrences of a given `methodName` irrespective of their arguments, `argumentIndex` need not be specified.
 
-The `emptyMethods` specify the APIs which need to be discarded from the code. For example, a statement `enableFlag(SAMPLE_STALE_FLAG);` will be deleted from the code. 
+For `returnType` and `receiverType`, types should be written as `boolean` or `void` for inbuilt types, and fully qualified for custom defined types. eg: `com.uber.piranha.XPFlagCleanerPositiveCases.XPTest`
```
mkr-plse, line 80: inbuilt -> primitive.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -287,30 +338,85 @@ private API getXPAPI(ExpressionTree et) {
...
+  private static Map<String, API> initializeFlagTypeToAPIMap() {
+    Map<String, API> map = new HashMap<>();
+    map.put("treated", API.IS_TREATED);
+    map.put("control", API.IS_CONTROL);
+    map.put("empty", API.DELETE_METHOD);
+    map.put("treatmentGroup", API.IS_TREATMENT_GROUP_CHECK);
+    return Collections.unmodifiableMap(map);
+  }
+
+  private boolean isMethodTreeMatchesMethodProperty(
```
mkr-plse, line 402: Can we get rid of this function? It is adding an unnecessary string comparison and there is nothing common between the two branches either.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -551,4 +556,1697 @@ public void runPiranhaWithFewProperties() throws IOException {
...
+      bcr.doTest();
+    } catch (ParseException pe) {
+      pe.printStackTrace();
+      assert_().fail("Incorrect parameters passed to the checker");
+    }
+  }
+
+  /*
+   * This test passes flag name as a Piranha argument.
+   * Whether or not flag name is passed, the noFlagCase simplification should not be affected.
```
mkr-plse, line 616: Nice.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -551,4 +556,1697 @@ public void runPiranhaWithFewProperties() throws IOException {
...
+    } catch (ParseException pe) {
+      pe.printStackTrace();
+      assert_().fail("Incorrect parameters passed to the checker");
+    }
+  }
+
+  /*
+   * Uses "properties_test_invalid.json" instead of "properties.json".
+   * When "methodProperties" is not specified,
+   * raise ParseException with appropriate exception message.
```
mkr-plse, line 2201: Great!

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -551,4 +556,1697 @@ public void runPiranhaWithFewProperties() throws IOException {
       assert_().fail("Incorrect parameters passed to the checker");
     }
   }
+
+  /*
+   * Test for the case when xpFlagName is ""
+   * Methods with no arguments are simplified, provided argumentIndex is not specified
+   * (assuming that returnType and receiverType - if specified - are a match)
+   * Uses "properties_test_noFlag.json" as the config file.
+   * Note that "Piranha:FlagName" is not considered when argumentIndex is not specified.
```
lazaroclapp, line 565: Out of curiosity: will this work only for versions of the method that take zero args?

My worry here is that a misconfiguration, where `Piranha:FlagName` is empty, and the `argumentIndex` is missing, it will go and delete half the code. We probably don't want that behavior. @murali ??

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -551,4 +556,1697 @@ public void runPiranhaWithFewProperties() throws IOException {
...
+      bcr.setArgs("-d", temporaryFolder.getRoot().getAbsolutePath());
+
+      bcr = addHelperClasses(bcr);
+      bcr.addInputLines(
+              "XPFlagCleanerSinglePositiveCase.java",
+              "package com.uber.piranha;",
+              "class XPFlagCleanerSinglePositiveCase {",
+              " private XPTest experimentation;",
+              " public String evaluate() {",
+              "  if (experimentation.isToggleEnabled()) { int a = 1; }",
```
lazaroclapp, line 640: What does this code do about `experimentation.isToggleEnabled(SOME_FLAG) { ... }`?

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -551,4 +556,1697 @@ public void runPiranhaWithFewProperties() throws IOException {
...
+    }
+  }
+
+  /*
+   * Uses "properties_test_invalid.json" instead of "properties.json".
+   * In it, the required top-level "methodProperties" is not present.
+   * As a result, refactoring will not be carried out, except for unreachable code.
+   * */
+  @Test
+  public void dontRefactorWhenInvalidConfigFile() throws IOException {
```
lazaroclapp, line 2135: 👍

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
lazaroclapp: I think we'd want any code that would delete `experimentation.isToggleEnabled(SOME_FLAG) { ... }` for every possible value of `SOME_FLAG` at once to error out by default. If someone really needs to do this, we can add a `Piranha:AllowMassRemoval=true` argument or something. It just feels like the kind of thing that might scare new users from trying piranha if they don't configure it right. 

Unless I am mistaken and this case will only delete treatment methods with zero args.

mkr-plse: My understanding is that this will only delete methods listed in the properties.json without checking for flagname in argument. At worst, this will delete all calls to an API if the argumentIndex is not specified. Your point on getting the configuration right is valid too. 

Let us leave this functionality as it is in the current version of the PR and if it becomes a source of confusion, let us revisit this.

lazaroclapp: I feel that worst case scenario is really really bad. Consider the effects of trying out Piranha in Uber's monorepos under those conditions, it would locally rewrite maybe hundreds of thousands of lines of code... and scare many people from ever trying it again.

I am fine with that check being a follow-up task, but I'd want to have it in place before we cut any release with this (it's already tricky enough that we are switching the config file format)

lazaroclapp: Also for a follow up task (I can do it): Print a message whenever the config points to a file ending in `.properties` explaining the format change and linking the docs in case someone is just upgrading Piranha.

mkr-plse: How about this: 
a) By default, argumentIndex is a required parameter in json. 
b) Using a configuration option (as you point above), this can be made optional. 

(a) will be good for most users and in cases, where a user wants to delete all calls to the API, (b) can be used.

lazaroclapp: Yup. That's exactly what I'd want. And we can add (a) to this PR and make (b) a follow up, to make this easier to land. But I'd rather have that small guardrail there than the extra functionality of deleting methods without arguments 😄

lazaroclapp: ^ @pranavsb Everything else is a nit and could be done later, but I'd like this one change before landing:

> a) By default, argumentIndex is a required parameter in json.

pranavsb:
> 
> Mostly just the not allowing index to be missing change (see discussion above). You don't need to add the option to re-enable the empty index mode, and you can decide whether or not to address the other nits, but this one is kinda important :/

Alright I see your point about the mass deletion being confusing for new users.
But if I make `argumentIndex` mandatory and not implement "enable empty index", piranha would no longer work for my codebase :/

How about I make `argumentIndex` mandatory, but it allows values from `-1,0,1...`?
`-1` would correspond to the noFlagCase, that is, a toggle method that takes no arguments.
Later the Piranha options can be modified and `argumentIndex` can go back to being 0 upwards if needed.

mkr-plse:
> > Mostly just the not allowing index to be missing change (see discussion above). You don't need to add the option to re-enable the empty index mode, and you can decide whether or not to address the other nits, but this one is kinda important :/
> 
> Alright I see your point about the mass deletion being confusing for new users.
> But if I make `argumentIndex` mandatory and not implement "enable empty index", Piranha would no longer work for my codebase :/
> 
> How about I make `argumentIndex` mandatory, but it allows values from `-1,0,1...`?
> `-1` would correspond to the noFlagCase, that is, a toggle method that takes no arguments.

Instead of `-1`, see the suggestion [here](https://github.com/uber/piranha/pull/39#discussion_r436969690) which references [this](https://github.com/uber/piranha/pull/39#discussion_r436955078). We can add an option `ArgumentIndexOptional` to Piranha and it will be `false` by default. This will avoid the confusion related to misconfiguration. For your codebase (and similar scenarios), the option can be set to true when Piranha is invoked. 

By adding this option, you can also leave most of the `noFlagTestCases` as is and simply update the configuration options passed to the tests. 
 
> P.S. I'll need a few days before starting on this, but I'll be happy to work on the nits too. Let me know if you catch anything else. :)

pranavsb:
> Instead of `-1`, see the suggestion [here](https://github.com/uber/piranha/pull/39#discussion_r436969690) which references [this](https://github.com/uber/piranha/pull/39#discussion_r436955078). We can add an option `ArgumentIndexOptional` to Piranha and it will be `false` by default. This will avoid the confusion related to misconfiguration. For your codebase (and similar scenarios), the option can be set to true when Piranha is invoked.
> 
> By adding this option, you can also leave most of the `noFlagTestCases` as is and simply update the configuration options passed to the tests.

Yes, that's what I meant by "enable empty index". Ok then, new Piranha option is the way. `ArgumentIndexOptional` is a good name. Will send updated PR.

lazaroclapp:
> > Instead of `-1`, see the suggestion [here](https://github.com/uber/piranha/pull/39#discussion_r436969690) which references [this](https://github.com/uber/piranha/pull/39#discussion_r436955078). We can add an option `ArgumentIndexOptional` to Piranha and it will be `false` by default. This will avoid the confusion related to misconfiguration. For your codebase (and similar scenarios), the option can be set to true when Piranha is invoked.
> > By adding this option, you can also leave most of the `noFlagTestCases` as is and simply update the configuration options passed to the tests.
> 
> Yes, that's what I meant by "enable empty index". Ok then, new Piranha option is the way. `ArgumentIndexOptional` is a good name. Will send updated PR.

Sorry for that extra work, but it's a better long term solution and, compared with the rest of this PR, just adding that option will be a fairly small change, I think. I would not be comfortable releasing a new version of Piranha with "empty index" allowed by default, even if we could land this to master.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -139,12 +159,18 @@
...
+  /**
+   * Information provided in the properties.json config file.
+   *
+   * <p>configMethodsMap is a map where key is method name and value is a list where each item in
+   * the list is a map that corresponds to each method property from properties.json. In most cases,
+   * the list would have only one element. But if someone reuses the same method name with different
+   * returnType/receiverType/argumentIndex, the list would have each method property map as one
+   * element.
+   */
+  private final Map<String, List<Map<String, Object>>> configMethodProperties = new HashMap<>();
```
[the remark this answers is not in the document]
pranavsb: I'd love to! But I won't be able to get back to this for a few weeks.

Have raised #48, will get back to it if you haven't already :)

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -328,8 +434,7 @@ private boolean isSymbolAndMatchesFlagName(Symbol argSym) {
    * @return True if matches. Otherwise false
    */
   private boolean isVarSymbolAndMatchesFlagName(Symbol argSym) {
-    return argSym != null
-        && argSym instanceof Symbol.VarSymbol
+    return argSym instanceof Symbol.VarSymbol
```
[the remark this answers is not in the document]
pranavsb: Yup, `instanceof` handles the null case. 
https://stackoverflow.com/a/2950415

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: >Also for a follow up task (I can do it): Print a message whenever the config points to a file ending in .properties explaining the format change and linking the docs in case someone is just upgrading Piranha.

Raised #49 for this. Clearly, I have issues :P

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
pranavsb: And as you know by now, argumentIndex not being specified would lead to all occurrences of `isToggleEnabled`, irrespective of its argument(s) count/value.

I'm in the process of adding `ArgumentIndexOptional` Piranha argument to fix this

java/README.md
pranavsb: Also added java.lang.string and a java.lang.boolean (which returns `Boolean` instead of `boolean`) test

pranavsb:
Pushed support for `ArgumentIndexOptional` and - brace yourself - more tests!

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -311,20 +318,27 @@ public String linkUrl() {
...
+      Set<Map<String, Object>> methodProperties, boolean isArgumentIndexOptional)
+      throws ParseException {
     for (Map<String, Object> methodProperty : methodProperties) {
-      String flagType = getValueStringFromMap(methodProperty, FLAG_TYPE_KEY);
       String methodName = getValueStringFromMap(methodProperty, METHOD_NAME_KEY);
-      if (flagType != null && methodName != null) {
+      if (methodName != null && getValueStringFromMap(methodProperty, FLAG_TYPE_KEY) != null) {
+        if (!isArgumentIndexOptional && getArgumentIndexFromMap(methodProperty) == null) {
+          throw new ParseException(
+              "methodProperty did not have argumentIndex. Use Piranha:ArgumentIndexOptional if required. Check:"
```
lazaroclapp, line 328: This should be a more explanatory error message:

"methodProperty did not have argumentIndex. By default, Piranha requires an argument index for flag APIs, to which the flag name/symbol will be passed. This is to avoid over-deletion of all occurrences of a flag API method. If you are sure you want to delete all instances of the method below, consider using Piranha:ArgumentIndexOptional=true to override this behavior. Check:"

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -311,20 +318,27 @@ public String linkUrl() {
...
-        } else {
-          List<Map<String, Object>> methodPropertiesForNewMethodName = new ArrayList<>();
-          methodPropertiesForNewMethodName.add(methodProperty);
-          configMethodProperties.put(methodName, methodPropertiesForNewMethodName);
+        if (methodPropertiesForMethodName == null) {
+          methodPropertiesForMethodName = new ArrayList<>();
+          configMethodProperties.put(methodName, methodPropertiesForMethodName);
         }
+        methodPropertiesForMethodName.add(methodProperty);
+      } else {
```
lazaroclapp, line 339: For readability, consider reversing this conditional, so you have:

```
if (methodName == null || getValueStringFromMap(methodProperty, FLAG_TYPE_KEY) == null) {
   throw ...;
}
if (!isArgumentIndexOptional && getArgumentIndexFromMap(methodProperty) == null) {
   throw ...;
}
List<Map<String, Object>> methodPropertiesForMethodName = ...;
...
```

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -372,15 +386,17 @@ private API getXPAPI(
       // if it's not a match, skip to next method property map
       String returnType = getValueStringFromMap(currentMethodProperties, RETURN_TYPE_STRING);
       if (returnType != null) {
-        if (!isMethodTreeMatchesMethodProperty(mst, RETURN_TYPE_STRING, returnType)) {
+        String mReturn = ASTHelpers.getReturnType(mst).toString();
+        if (!returnType.equalsIgnoreCase(mReturn)) {
```
lazaroclapp, line 390: Why `equalsIgnoreCase` instead of case sensitive equals?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
pranavsb: Done.
Also replaced the add methodProperty/new ArrayList logic with a one-liner using `computeIfAbsent`

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
pranavsb: Just to allow users the flexibility of wrongly capitalizing stuff. Like `java.lang.string` or the fully qualified name of their class.

lazaroclapp: I say let's compare case-sensitive, since Java itself is case sensitive, I think, e.g. you can have both a type `com.foo.Bar` and one called `com.foo.bar` at the same time in the same JVM (you probably shouldn't, but still). That said, I have no super strong opinion here. cc: @mkr-plse Any thoughts?

pranavsb: >...you can have both a type com.foo.Bar and one called com.foo.bar ...

Oh, didn't know this. Then a case sensitive match is a better default for now - just in case someone does this. 

Updated and fixed testcase.

pranavsb:
Pushed case-sensitive match for returnType and receiverType (and some refactoring)

lazaroclapp:
Congrats and thank you for the massive contribution @pranavsb !

pranavsb:
> Congrats and thank you for the massive contribution @pranavsb !

Thanks for all your help. It was good fun! :)
@lazaroclapp @mkr-plse

mkr-plse:
Indeed, this is a massive contribution. Thanks a lot @pranavsb. Looking forward to more such contributions ;)

Issue #43, opened 2020-06-01
PiranhaJava - Unable to remove stale Unit Test code 


Samhithgb:
Hi, 

I'm defining an annotation : 

```
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.METHOD})
public @interface FlagTesting {
    MyClass.TestExperimentName[] treated();
}

```
and using it in my unit test : 

```
@Test
@FlagTesting(treated = MyClass.TestExperimentName.SAMPLE_STALE_FLAG)
public void addition_isCorrect() throws Exception {
...
}

```
When I build this, although the flag and its corresponding methods in the product code get removed as expected, unit test build fails with the error : 

```
Execution failed for task ':app:compileInternalChromeBeta64DebugUnitTestJavaWithJavac'.
> Piranha is not a valid checker name
```

Any pointers on why this might happen will be really great. 

PS : Using this on Android code that uses gradle.

mkr-plse:
Have you updated the piranha.properties file?

Samhithgb:
@mkr-plse Yes, this is how it looks : 

> treatedMethods=IsFeatureEnabled,flagEnabled
> controlMethods=flagDisabled
> emptyMethods=enableFlag,disableFlag
> annotations=FlagTesting
> linkURL=<provide_your_url>
> 

The methods listed get removed out as expected.

lazaroclapp:
This sounds like Piranha itself is not being loaded (but Error Prone is and is configured to run piranha). Do you have a sample gradle file showing the issue? 

`:app:compileInternalChromeBeta64DebugUnitTestJavaWithJavac` sounds like it potentially uses a custom `javac` command line. Make sure Piranha is on the classpath for it.

Samhithgb:
@lazaroclapp Thanks for you response. 

Adding Piranha to errorprone configuration resolved the issue seems like.  

However, there is a new issue : The stale flag gets removed from the product code initially and then the unit test build task fails comlaining : 

> /Users/samhith/RfAndroid/UI/app/src/test/java/com/citrix/client/Receiver/ExampleUnitTest.java:28: error: cannot find symbol
> @FlagTesting(treated = MyClass.TestExperimentName.SAMPLE_STALE_FLAG)
> 

My gradle file currently looks something like this : 

```

plugins {
    id "com.github.sherter.google-java-format" version "0.7.1"
    id "net.ltgt.errorprone" version "0.6"
}

...
android {
    ...
    compileOptions {
[...]

Issue #42, opened 2020-05-30
PiranhaJava // Potential string compare bug in matchIf


pranavsb:
In `XPFlagCleaner`, the `public Description matchIf(IfTree ifTree, VisitorState visitorState)` method uses the equality operator `==` for String comparison instead of `.equals`

`if (update) {
      if (replacementPrefix != EMPTY) {
        replacementString = replacementPrefix + replacementString;
      }`

https://github.com/uber/piranha/blob/dd050f86315ab3e719fa5d22d7f0039ad844b039/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L853

mkr-plse:
Fixed in #45

PR #45, opened 2020-06-02
Fix PiranhaJava issue when not all properties are specified

mkr-plse:
Fixes issues [43](https://github.com/uber/piranha/issues/43) and [42](https://github.com/uber/piranha/issues/42)


lazaroclapp:
Stamped, as `treatmentGroupMethods` definitely should be optional. Not sure I agree it ought to be removed.

Issue #54, opened 2020-06-15
Tidy up internal data structures (map) generated from properties.json


lazaroclapp:
`Map<String, List<Map<String, Object>>>`

to something like:

`MultiMap<String, PiranhaMethod>`

Issue #49, opened 2020-06-13
PiranhaJava // print warning when .properties is used after v0.0.4


pranavsb:
Version 0.0.4 moves to a JSON format for the Piranha config. 
When users upgrade, they would have their old `piranha.properties` files lying around.
Piranha should print a warning and point them to the documentation so they can move to a json config.

For context:
https://github.com/uber/piranha/pull/39#discussion_r436967528

PR #55, opened 2020-06-16
Refactor PiranhaJava configuration internals.

lazaroclapp:
This change:
a) Switches `Map<String, List<Map<String, Object>>>` for an immutable map
   of structured records (PiranhaMethodRecord) [#54]
b) Performs ALL configuration parsing at init() time.
c) Removes confused mixed uses of two ParseException classes (one from
  `java.text.` and the other from `org.json.simple.parser.` and
   introduces our own `PiranhaConfigurationException` to encapsulate
   everything that can go wrong during Piranha configuration/initialization.
d) Adds a WARNING when seeing an old-format `piranha.properties` file. It
   adds a test for this. [#49]
e) Removes `throws` clause from the `XPFlagCleaner` constructor, as that
   exception can no longer happen there (it must happen during the latter
   call to `init()` and Piranha just silently disables itself in that case)

The change to `XPFlagCleanerTest` is huge, but mostly just removing redundant
try{...}catch{...} blocks for the never thrown exception in (e).


java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -2237,16 +2044,34 @@ public void exceptionWhenConfigFileInvalid() throws Exception {
...
+
+    XPFlagCleaner flagCleaner = new XPFlagCleaner();
+    flagCleaner.init(b.build());
+  }
+
+  /*
+   * Check that the user is warned when using an old style piranha.properties file
+   * */
+  @Test
+  public void warningWhenConfigFileOld() throws Exception {
```
lazaroclapp, line 2058: This test is new, everything else in this file is removing old `try{...}catch{...}` blocks and updating error messages.

java/piranha/src/test/resources/config/invalid/piranha.properties
```diff
@@ -0,0 +1,6 @@
+treatedMethods=isToggleEnabled,isFlagTreated
+controlMethods=isToggleDisabled
+emptyMethods=putToggleEnabled,putToggleDisabled,includeEvent
+treatmentGroupMethods=isToggleInGroup
+annotations=ToggleTesting
+linkURL=<provide_your_url>
```
lazaroclapp, line 6: Will fix this missing new line.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -349,23 +335,20 @@ private API getXPAPI(ExpressionTree et) {
...
-      MethodInvocationTree mit, List<Map<String, Object>> methodPropertiesForMethodName) {
-    for (Map<String, Object> currentMethodProperties : methodPropertiesForMethodName) {
-      Integer argumentIndex = getArgumentIndexFromMap(currentMethodProperties);
-
+      MethodInvocationTree mit, ImmutableCollection<PiranhaMethodRecord> methodRecordsForName) {
+    for (PiranhaMethodRecord methodRecord : methodRecordsForName) {
       // when argumentIndex is specified, if mit's argument at argIndex doesn't match xpFlagName,
       // skip to next method property map
-      if (argumentIndex != null) {
+      if (methodRecord.getArgumentIdx().isPresent()) {
```
mkr-plse, line 350: Can we store the `methodRecord.getArgumentIdx()` locally and re-use it to avoid verbosity?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -298,7 +305,7 @@ public Description matchCompilationUnit(
     if (!initialized && !disabled) {
       try {
         init(flags);
```
mkr-plse, line 307: just noticing this even though not part of this change. Why was this under `matchCompilationUnit`? Ideally, `init` at the beginning right after construction would be better. Is this to handle the silent exception problem in the constructor?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -379,37 +362,26 @@ private API getXPAPI(
       MemberSelectTree mst = ((MemberSelectTree) mit.getMethodSelect());
       // when returnType is specified, check if mst's return type matches it
       // if it's not a match, skip to next method property map
-      String returnType = getValueStringFromMap(currentMethodProperties, RETURN_TYPE_STRING);
-      if (returnType != null) {
+      if (methodRecord.getReturnType().isPresent()) {
         String mReturn = ASTHelpers.getReturnType(mst).toString();
-        if (!returnType.equals(mReturn)) {
+        if (!methodRecord.getReturnType().get().equals(mReturn)) {
```
mkr-plse, line 367: Store `methodRecord.getReturnType()` to avoid verbosity?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -379,37 +362,26 @@ private API getXPAPI(
...
         }
       }
       // when receiverType is specified, check if mst's receiver type matches it
       // if it's not a match, skip to next method property map
-      String receiverType = getValueStringFromMap(currentMethodProperties, RECEIVER_TYPE_STRING);
-      if (receiverType != null) {
+      if (methodRecord.getReceiverType().isPresent()) {
         String mReceive = ASTHelpers.getReceiverType(mst).toString();
-        if (!receiverType.equals(mReceive)) {
+        if (!methodRecord.getReceiverType().get().equals(mReceive)) {
```
mkr-plse, line 375: verbosity:?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: I don't think it saves too much. We'd be replacing a call to `.getArgumentIdx()` with adding a new `Optional<Integer>` local. Either way we need to check for `isPresent()` and then do `get()` to get the actual value, so this seemed clearer to me. That said, I don't feel strongly either way, can rewrite to introduce the local.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: See above.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: See above.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: This is exactly to avoid the silent swallowing of the error inside the constructor and use of the zero-arg constructor. I think it is safer and easier to debug this way, but it is a bit of a workaround for the EP's "feature" of silent fallback of constructors :/ 

It's also because of this that I couldn't make the `ImmutableMultimap` also `final`. But I do think it's the best we can do right now.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
mkr-plse: I am not referring to the performance (the compiler will likely take care of this) but to the readability. Feel free to update (or not).

lazaroclapp: I am also not talking about performance. I am just not sure it's more readable/less verbose/less idiomatic to go `var.getField.isPresent()` and then `T field = var.getField.get()` vs `Optional<T> maybeField = var.getField()` then `maybeField.isPresent` then `T field = maybeField.get()`.

But I have no strong opinion here, so let me update.

PR #59, opened 2020-06-17
update docs on how to run the sample app

swayamraina:


mkr-plse:
This [section](https://github.com/uber/piranha/blob/master/java/README.md#example-refactoring) along with the text in the [beginning](https://github.com/uber/piranha/blob/master/java/README.md#overview) of the README covers the textual description in this PR. Is there anything extra that you had to do to handle IntelliJ for Piranha beyond the standard setup instructions for ErrorProne? If so, that aspect may be useful. Otherwise, we don't need to update the doc yet.

swayamraina:
I guess we can only add a link to the error-prone installation page on how to setup error-prone.
Other than that I don't see much for now.

mkr-plse:
Thanks @swayamraina. We already have the link to ErrorProne installation at the top of the README. Closing this PR as per your suggestion.

Issue #35, opened 2020-04-23
Additional paranthesis in PiranhaJava


mkr-plse:
Given input:
```
if(cond1 || (cond2 && experiments.isToggleEnabled(STALE_FLAG)) {
  // do something
}
``` 
the output of PiranhaJava is 

```
if(cond1 || (cond2)) {
  // do something
}
```

The additional paranthesis around `cond2` should be eliminated.

jitendrs:
can you please assign to me?

ketkarameya:
This issue has been addressed in the latest release (0.1.6). Maybe we can close this issue?

mkr-plse:
Thanks @ketkarameya. Closing this issue.

PR #60, opened 2020-06-18
handle redundant parenthesis

swayamraina:
fixes https://github.com/uber/piranha/issues/35

### DIFF
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -702,7 +702,7 @@ public class XPFlagCleaner extends BugChecker
       boolean loneCondition =
           (checkLoneCondition(tree.getLeftOperand(), state, remainingSubTree.toString())
               || checkLoneCondition(tree.getRightOperand(), state, remainingSubTree.toString()));
-      if (loneCondition) {
+      if (loneCondition && isTreated) {
         // strip parenthesis
         int end = state.getEndPosition(tree);
         while (state.getSourceCode().charAt(end) != ')') end++;
@@ -735,8 +735,7 @@ public class XPFlagCleaner extends BugChecker
   private boolean checkLoneCondition(ExpressionTree tree, VisitorState state, String remaining) {
     return (state.getSourceForNode(tree).equals(remaining)
         && (!tree.getKind().equals(Kind.CONDITIONAL_OR)
-            && !tree.getKind().equals(Kind.CONDITIONAL_AND))
-        && isTreated);
+            && !tree.getKind().equals(Kind.CONDITIONAL_AND)));
   }
 
   @Override
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -705,8 +705,10 @@ public class XPFlagCleaner extends BugChecker
       if (loneCondition) {
         // strip parenthesis
         int end = state.getEndPosition(tree);
+        while (state.getSourceCode().charAt(end) != ')') end++;
         int start = end - state.getSourceForNode(tree).length();
-        fixBuilder.replace(start - 1, end + 1, remainingSubTree.toString());
+        while (state.getSourceCode().charAt(start) != '(') start--;
+        fixBuilder.replace(start, end + 1, remainingSubTree.toString());
       } else {
         fixBuilder.replace(tree, remainingSubTree.toString());
       }
@@ -733,7 +735,8 @@ public class XPFlagCleaner extends BugChecker
   private boolean checkLoneCondition(ExpressionTree tree, VisitorState state, String remaining) {
     return (state.getSourceForNode(tree).equals(remaining)
         && (!tree.getKind().equals(Kind.CONDITIONAL_OR)
-            && !tree.getKind().equals(Kind.CONDITIONAL_AND)));
+            && !tree.getKind().equals(Kind.CONDITIONAL_AND))
+        && isTreated);
   }
 
   @Override
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -15,6 +15,7 @@ package com.uber.piranha;
 
 import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
 
+import com.facebook.infer.annotation.Initializer;
 import com.google.auto.service.AutoService;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableCollection;
@@ -65,12 +66,29 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 import javax.lang.model.element.ElementKind;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
 import org.json.simple.parser.ParseException;
 
-/** @author murali@uber.com (Murali Krishna Ramanathan) */
+/**
+ * This is the core PiranhaJava checker and code rewriting class.
+ *
+ * <p>This checker iterates over the AST of each compilation unit, performing the following steps:
+ * a) Replaces occurrences of flag checking APIs with boolean constants based on Piranha's
+ * configuration and flag name arguments, and removes outdated flag setting operations. b)
+ * Simplifies boolean expressions including these new constants. c) Deletes code that has become
+ * unreachable due to conditional guards which can no longer evaluate to true. d) Deletes outdated
+ * enums and classes representing the removed flags and/or their treatment conditions.
+ *
+ * <p>In some cases, Piranha/XPFlagCleaner will want to delete a whole file. Since EP has no
+ * facilities for creating a patch that removes an entire file, it will instead replace its contents
+ * by an empty enum/class and add a special comment indicating it must be removed. Automated scripts
+ * using this code can then perform the actual file deletion previous to creating a PR or diff.
+ *
+ * @author murali@uber.com (Murali Krishna Ramanathan)
+ */
 @AutoService(BugChecker.class)
 @BugPattern(
     name = "Piranha",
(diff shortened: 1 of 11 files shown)


swayamraina:
sure I am working on that but had a few questions on running Test cases.
When I run a test case (lets say `refactorUnreachableWhenInvalidConfigFile`), I get the below error

```
java.lang.NoSuchMethodError: com.google.errorprone.ErrorProneInMemoryFileManager.getJavaFileObjects([Ljava/nio/file/Path;)Ljava/lang/Iterable;

	at com.google.errorprone.ErrorProneInMemoryFileManager.forResource(ErrorProneInMemoryFileManager.java:89)
	at com.google.errorprone.ErrorProneInMemoryFileManager.forResource(ErrorProneInMemoryFileManager.java:109)
	at com.google.errorprone.BugCheckerRefactoringTestHelper.addInput(BugCheckerRefactoringTestHelper.java:151)
	at com.uber.piranha.XPFlagCleanerTest.addHelperClasses(XPFlagCleanerTest.java:254)
	at com.uber.piranha.XPFlagCleanerTest.refactorUnreachableWhenInvalidConfigFile(XPFlagCleanerTest.java:1988)
```

mkr-plse:
> sure I am working on that but had a few questions on running Test cases.
> When I run a test case (lets say `refactorUnreachableWhenInvalidConfigFile`), I get the below error

How are you running the test? If you do a `./gradlew test` from the `java` directory, it should run the tests without any problem.

swayamraina:
this error came when running test cases from InitelliJ UI but works fine from the terminal. Seems like an issue in the InitelliJ error-prone plugin

mkr-plse:
> this error came when running test cases from InitelliJ UI but works fine from the terminal.

Great. Looks like you forgot to push the tests?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -699,7 +699,19 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
     if (deletedSubTree != null) {
       Description.Builder builder = buildDescription(tree);
       SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
-      fixBuilder.replace(tree, remainingSubTree.toString());
```
mkr-plse, line 702: If the tree kind is `ParenthesizedTree` here, then this utility [call](https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L342) can be used.

swayamraina: the tree kind is `conditional_and` in such cases. 
`parenthesised tree` is a couple levels above from the current node position

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -710,6 +722,23 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
...
+   * @param tree expression tree
+   * @param state visitor state
+   * @param remaining expression after removal
+   * @return true if parenthesis are redundant
+   */
+  private boolean checkLoneCondition(ExpressionTree tree, VisitorState state, String remaining) {
+    return (state.getSourceForNode(tree).equals(remaining)
+        && (!tree.getKind().equals(Kind.CONDITIONAL_OR)
+            && !tree.getKind().equals(Kind.CONDITIONAL_AND))
+        && isTreated);
```
mkr-plse, line 739: why is there a `isTreated` here? This code should not be dependent on the treatment value.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
mkr-plse: Ok. Since there is a check for lone condition by getting the leftOperand, presumably, the parent can be checked for that type to refactor appropriately.

swayamraina: since the tree does not have any back pointer to the parent node and these calls happen as the AST node visits happen we cannot directly use this particular method.
If we really want to use this method, we need to somehow get the parent tree pointer. And i guess this can be achieved using recursion based handling.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
swayamraina: to check if the complete expression boils down to a single identifier expression there is a dependency on the treatment value also. This is because it totally depends on which branch one is selecting. This part is somewhat similar to `evalExpr` method call where we check the branch path.

Based on the branch path taken, the child code deletion happens which in turn affects if the parent node is a lone identifier contained in parenthesis.

swayamraina: @mkr-plse I do believe that moving this treatment check out of this method and placing this at the top level does makes sense

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -699,7 +699,19 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
     if (deletedSubTree != null) {
       Description.Builder builder = buildDescription(tree);
       SuggestedFix.Builder fixBuilder = SuggestedFix.builder();
-      fixBuilder.replace(tree, remainingSubTree.toString());
+      boolean loneCondition =
+          (checkLoneCondition(tree.getLeftOperand(), state, remainingSubTree.toString())
+              || checkLoneCondition(tree.getRightOperand(), state, remainingSubTree.toString()));
+      if (loneCondition && isTreated) {
```
mkr-plse, line 705: the fix should be independent on `isTreated`.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
msridhar: FYI `state.getPath()` gives you the `TreePath` to the current AST node, from which you can retrieve all the ancestor nodes

msridhar:
Also may want to see this Error Prone check:

https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryParentheses.java

The `ASTHelpers.requiresParentheses` routine they use could be helpful.

swayamraina:
> Also may want to see this Error Prone check:
> 
> https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryParentheses.java
> 
> The `ASTHelpers.requiresParentheses` routine they use could be helpful.

Thanks @msridhar for helping here. Though this particular way will not help out here but this example did helped me in figuring out another approach which I believe is the correct one to fix this.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -746,14 +747,61 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
...
     return Description.NO_MATCH;
   }
 
+  @Override
+  public Description matchParenthesized(ParenthesizedTree tree, VisitorState state) {
+    if (overLaps(tree, state)) {
+      return Description.NO_MATCH;
+    }
+
+    ExpressionTree expression = tree.getExpression();
```
mkr-plse, line 763: The `evalExpr` method (see [here](https://github.com/uber/piranha/blob/eb73d95a36b4a8cb92e8db4e1a3cb7bf3dc27b1a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L498)) performs the evaluation of the expression. Also, there is some overlap of code in `matchBinary`  and in this method. Can you please consider refactoring this change to see: 
a) whether it could be handled within `evalExpr` itself? 
b) if not, have a common helper for `matchBinary` and `matchParenthesized` and update `matchBinary` too? 

Also, to check the generalization of the solution, can you add more tests with parenthesis? examples: 
1. `if(a && (b && (isTreated(...) || c)`
2. `if((a && !isControl(...)) && c)`
3. `if(a && !(b && isTreated(...))`
4. `if((a && isTreated(...)) || (b && isTreated(...))`
5. ...
...

msridhar:
Another thought is that if multi-pass analysis for Java is on the horizon, one could maybe just run the existing `UnnecessaryParentheses` check in autofix mode alongside Piranha after the first pass and that would just solve this problem?  It may be better to integrate the logic into Piranha, just might be good to understand the tradeoffs

mkr-plse:
> Another thought is that if multi-pass analysis for Java is on the horizon, one could maybe just run the existing `UnnecessaryParentheses` check in autofix mode alongside Piranha after the first pass and that would just solve this problem? It may be better to integrate the logic into Piranha, just might be good to understand the tradeoffs

That is a good idea. My initial plan for the multi-pass analysis was to use the first pass as a derivation pass to identify style/locations of change and then apply the cumulative changes in the second pass. This will overcome any compilation issues that are introduced by the refactorings. But, there is no harm in trying this out as there are couple of benefits: 
1. Good prototype for the multi-pass analysis. 
2. The UnnecessaryParentheses [checker](https://errorprone.info/bugpattern/UnnecessaryParentheses) should comprehensively handle all cases related to parentheses. 
3. We can identify the boundaries of current Piranha refactorings with this and fix any potential compilation issues that are introduced.

swayamraina:
> Also, to check the generalization of the solution, can you add more tests with parenthesis?

I have added a few more test cases to test the fix and all pass. However there is one case that results in compilation issue and the same is observed for master code.

```
public void foo(boolean x, boolean y) {
  if ((x && expt.flagEnabled(TestExperimentName.SAMPLE_STALE_FLAG)) || (y && !expt.flagEnabled(TestExperimentName.SAMPLE_STALE_FLAG))) {
    System.out.println("Hello World");
  }
}
```

The above example code IMO is very very rare and is converted to 
```
public void foo(boolean x, boolean y) {
  if ((y && !expt.flagEnabled(TestExperimentName.SAMPLE_STALE_FLAG))) {
    System.out.println("Hello World");
  }
}
```
when `Piranha:IsTreated=false` and `Piranha:FlagName=SAMPLE_STALE_FLAG`
We can fix this but IMO this is a rare case and should be fixed only if someone encounters this issue while using this tool and we can create a marker issue reminding us about its existence. 

> Also, there is some overlap of code in matchBinary and in this method

There is some (very less) overlap which I believe cannot be carved out to a common method.

swayamraina:
Any updates @mkr-plse ?

mkr-plse:
> Any updates @mkr-plse ?

I would like the redundancy in the code to be removed. Couple of possibilities:
1. Update `evalExpr` to return a custom object containing `Value` and a simplified expression. 
2. Have a common utility method that does the generic processing. The parenthesis related code should process parenthesis and leave the binary expression simplification to the utility method. 

For this PR, I recommend the second option.

mkr-plse:
Closing PR given this issue is handled. @ketkarameya please re-open otherwise.

PR #62, opened 2020-06-18
Enable NullAway for the PiranhaJava core.

lazaroclapp:
This adds NullAway to our build for PiranhaJava code, excluding test code and sample projects for now.

No NPEs found, but there are some preconditions that was good to document/encode.

### DIFF
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -15,6 +15,7 @@ package com.uber.piranha;
 
 import static com.google.errorprone.BugPattern.SeverityLevel.SUGGESTION;
 
+import com.facebook.infer.annotation.Initializer;
 import com.google.auto.service.AutoService;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableCollection;
@@ -65,6 +66,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 import javax.lang.model.element.ElementKind;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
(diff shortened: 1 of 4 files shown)


lazaroclapp:
There is saying in Spanish that basically translates to: "In the blacksmith home: wooden tooling.", and I think we are basically in that case here 😅 . 

Let's dogfood our own tooling outside of the large Uber repos.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -202,7 +209,10 @@ void init(ErrorProneFlags flags) throws PiranhaConfigurationException {
         JSONObject propertiesJson =
             (JSONObject)
                 parser.parse(Files.newBufferedReader(configFilePath, StandardCharsets.UTF_8));
-        linkURL = (String) propertiesJson.get("linkURL");
+        final String linkURLKey = "linkURL";
```
lazaroclapp, line 212: Annoying NullAway bug/gotcha/quirk: https://github.com/uber/NullAway/issues/398

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -408,9 +425,11 @@ private boolean isSymbolAndMatchesFlagName(Symbol argSym) {
...
    */
   private boolean isVarSymbolAndMatchesFlagName(Symbol argSym) {
-    return argSym instanceof Symbol.VarSymbol
-        && ((Symbol.VarSymbol) argSym).getConstantValue() != null
-        && ((Symbol.VarSymbol) argSym).getConstantValue().equals(xpFlagName);
+    if (argSym instanceof Symbol.VarSymbol) {
+      Object constantValue = ((Symbol.VarSymbol) argSym).getConstantValue();
+      return constantValue != null && constantValue.equals(xpFlagName);
+    }
+    return false;
```
lazaroclapp, line 432: An example of a method that must be changed to basically equivalent code for NullAway's understanding benefit :)

Issue #61, opened 2020-06-18
PiranhaJava: add extra space (" ") after 'If' block


swayamraina:
When the input. code is 
```
 public void foo (boolean x) {
    if (x) {
      System.out.println("if block");
    } else if (expt.flagEnabled(TestExperimentName.SAMPLE_STALE_FLAG)) {
      System.out.println("else if block");
    } else {
      System.out.println("else block");
    }
  }
```
and `IsTreated=true`

PiranhaJava emits following output,
```
public void foo (boolean x) {
    if (x){
      System.out.println("if block");
    } else {
      System.out.println("else if block");
    }
  }
```

Notice the removed space character in `if (x){`

swayamraina:
This is a very small issue and requires a one-line fix.
This happens when we try to merge the (if) conditions.

PR #64, opened 2020-06-19
added spacing

swayamraina:
fixes https://github.com/uber/piranha/issues/61


lazaroclapp:
> @lazaroclapp Could this have been handled by GJF?

In Uber's code? Yeah. But not every external user is going to use a code auto-formatter, which we kind of assumed for the internal deployment 🙂  

We could recommend it in the docs, since I do feel auto-formatting is the only way to enforce style guidelines when you have automated tools like Piranha rewriting your code. But, otherwise, I am fine with accepting this PR and other contributions improving the default formatting of code spewed out of Piranha as a standalone tool. No sense in generating uglier code than we have to and internally it will all get GJFed either way.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -2161,4 +2161,204 @@ public void exceptionWhenArgumentIndexNotSpecifiedAndNotArgumentIndexOptional()
     XPFlagCleaner flagCleaner = new XPFlagCleaner();
     flagCleaner.init(b.build());
   }
+
+  @Test
+  public void testSpacingForSingleElseIfTreated() throws IOException {
```
lazaroclapp, line 2166: Just to confirm: you checked that these tests do fail if you mess up the spacing, right? I am not 100% sure `BugCheckerRefactoringTestHelper` cares about that, it might be doing AST-level comparisons only. In which case we are fine with the fix above, but the tests might not be useful. If the tests work, then I am fine with them, though.

msridhar:
> We could recommend it in the docs, since I do feel auto-formatting is the only way to enforce style guidelines when you have automated tools like Piranha rewriting your code.

👍 We should try to make the world a better place for code transformation tools 🙂

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
swayamraina: Thanks @lazaroclapp for pointing this out. Yes, the bug-checker indeed only does the AST level comparisons and will not be able to detect the failure here.
IMO we could add string level comparison to check the formatting issues but not sure if having both will add much as string comparison kind of does the AST check (but will not be able to point out the error) 
Note that the tests will still fail if the code produced has some errors. What are your thoughts?

lazaroclapp: My take on this is that I am against string-level comparisons in tests, as it will cause those tests to break every time Piranha changes slightly in things like the whitespace formatting of the code, which I personally don't consider part of the functionality of the tool. AST-level checks is what we want for testing Piranha's functionality.

That said, I get that you'd like it to produce cleaner code by default, and I am not opposed to it where it doesn't break existing tests, so I am happy to accept the patch. I think in this case - and this case _only_ - the patch doesn't need tests. My suggestion is that you remove the changes to `XPFlagCleanerTest` and keep only the one-line change in `XPFlagCleaner` as a matter of practicality. That particular line is certainly exercised by (and passes) our existing suite.

We do strongly recommend to use Piranha with an automated code formatter, though, and can make no promises that there will never be regressions on things like whitespace and line length without one 🙂

swayamraina: Agreed. String comparisons will break tests every time a minor formatting patch is merged.
Will remove the tests as suggested.

PR #69, opened 2020-06-21
Add note recommending automated code formatting

lazaroclapp:
Piranha is better when paired with automated code formatting, and
this note makes it clear that we don't want to commit to produce 
pre-formatted code in all cases, much less to provide coding style related
configuration options.


lazaroclapp:
cc: @msridhar

Issue #44, opened 2020-06-02
PiranhaJava // minor - avoid simplifying when xpFlagName is empty string ("")


pranavsb:
Running Piranha with `-XepOpt:Piranha:FlagName=""` results in the deletion of all static imports.

Two places in `XPFlagCleaner` can cause this, because the code is `.endsWith(xpFlagName)` when comparing with expressions. Since `xpFlagName` is `""`, that always evaluates to true.

The specific places are: 
https://github.com/uber/piranha/blob/dd050f86315ab3e719fa5d22d7f0039ad844b039/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L534

and

https://github.com/uber/piranha/blob/dd050f86315ab3e719fa5d22d7f0039ad844b039/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L726

The second case seems to be rarer.

The can be fixed with another check `!xpFlagName.equals(EMPTY)`

swayamraina:
The second case mentioned by @pranavsb will actually impact users who write test cases using configuration (or profile) based test cases powered using annotations

PR #75, opened 2020-06-30
Handle empty flag name

swayamraina:
fixes https://github.com/uber/piranha/issues/44

### DIFF
--- a/java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
+++ b/java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
@@ -2211,4 +2211,136 @@ public class XPFlagCleanerTest {
 
     bcr.doTest();
   }
+
+  @Test
+  public void testEmptyFlagRemovesAnnotatedMethodsTreatment() throws IOException {
+
+    ErrorProneFlags.Builder b = ErrorProneFlags.builder();
+    b.putFlag("Piranha:FlagName", "");
+    b.putFlag("Piranha:IsTreated", "true");
+    b.putFlag("Piranha:Config", "config/properties.json");
+
+    BugCheckerRefactoringTestHelper bcr =
+        BugCheckerRefactoringTestHelper.newInstance(new XPFlagCleaner(b.build()), getClass());
+
+    bcr.setArgs("-d", temporaryFolder.getRoot().getAbsolutePath());
+
+    bcr = addHelperClasses(bcr);
+    bcr.addInputLines(
+            "EmptyFlagRemovesAnnotatedMethods.java",
+            "package com.uber.piranha;",
+            "class EmptyFlagRemovesAnnotatedMethods {",
+            "  enum TestExperimentName {",
+            "    SAMPLE_STALE_FLAG",
+            "  }",
+            "  public String evaluate(int x) {",
+            "    if (x == 1) { return \"yes\"; }",
+            "    return \"no\";",
+            "  }",
+            "  @ToggleTesting(treated = TestExperimentName.SAMPLE_STALE_FLAG)",
+            "  public void x () {}",
+            "}")
+        .addOutputLines(
+            "EmptyFlagRemovesAnnotatedMethods.java",
+            "package com.uber.piranha;",
+            "class EmptyFlagRemovesAnnotatedMethods {",
+            "  enum TestExperimentName {",
+            "    SAMPLE_STALE_FLAG",
+            "  }",
+            "  public String evaluate(int x) {",
+            "    if (x == 1) { return \"yes\"; }",
+            "    return \"no\";",
+            "  }",
+            "  @ToggleTesting(treated = TestExperimentName.SAMPLE_STALE_FLAG)",
+            "  public void x () {}",
+            "}")
+        .addInputLines(
+            "ToggleTesting.java",
+            "package com.uber.piranha;",
+            "import java.lang.annotation.ElementType;",
+            "import java.lang.annotation.Retention;",
+            "import java.lang.annotation.RetentionPolicy;",
+            "import java.lang.annotation.Target;",
+            "@Retention(RetentionPolicy.RUNTIME)",
+            "@Target({ElementType.METHOD})",
+            "@interface ToggleTesting {",
+            "  EmptyFlagRemovesAnnotatedMethods.TestExperimentName treated();",
+            "}")
+        .addOutputLines(
+            "ToggleTesting.java",
+            "package com.uber.piranha;",
+            "import java.lang.annotation.ElementType;",
+            "import java.lang.annotation.Retention;",
+            "import java.lang.annotation.RetentionPolicy;",
+            "import java.lang.annotation.Target;",
+            "@Retention(RetentionPolicy.RUNTIME)",
+            "@Target({ElementType.METHOD})",
+            "@interface ToggleTesting {",
+            "  EmptyFlagRemovesAnnotatedMethods.TestExperimentName treated();",
+            "}");
+
+    bcr.doTest();
+  }
+
+  @Test
+  public void testEmptyFlagRemovesAnnotatedMethodsControl() throws IOException {
+
+    ErrorProneFlags.Builder b = ErrorProneFlags.builder();
+    b.putFlag("Piranha:FlagName", "");
+    b.putFlag("Piranha:IsTreated", "false");
+    b.putFlag("Piranha:Config", "config/properties.json");
+
+    BugCheckerRefactoringTestHelper bcr =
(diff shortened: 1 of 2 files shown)


pranavsb:
Nice!

Instead of adding these two checks which get called everytime, it might be better to take care of empty flag case in initialisation (`init` method) itself.

There's a `isPresent` check for xpFlagName, under that you can add the not EMPTY check.
(Under it because I feel EMPTY flag name shouldn't cause `PiranhaConfigException`, it should just keep flag name as the default initialisation dummy value, so that Piranha continues to work for people without flag name in their toggles)

swayamraina:
```
Optional<String> s = flags.get("Piranha:FlagName");
```
So 'Piranha' assumes empty string to be a valid input. Keeping dummy value is one way to avoid this side-effect but pivoting a dummy value also needs some discussion (for uniqueness).

pranavsb:
> ```
> Optional<String> s = flags.get("Piranha:FlagName");
> ```
> 
> So 'Piranha' assumes empty string to be a valid input. 

Yes, I was suggesting:
```
Optional<String> s = flags.get("Piranha:FlagName");
    if (s.isPresent()) {
      if (!EMPTY.equal(s.get())) {
         xpFlagName = s.get();
      }
      isTreated = flags.getBoolean("Piranha:IsTreated").orElse(true);
      treatmentGroup = flags.get("Piranha:TreatmentGroup").orElse("").toLowerCase();
    } 
    ...
```

> Keeping dummy value is one way to avoid this side-effect but pivoting a dummy value also needs some discussion (for uniqueness).

Yeah I had this thought too, but xpFlagName is initialized like this, which I think is rare enough.
`xpFlagName = "_xpflag_dummy"`

Thoughts? @mkr-plse @lazaroclapp

mkr-plse:
> Thoughts? @mkr-plse @lazaroclapp

I agree with this approach. If the user doesn't specify the flagname, an exception will be thrown. If an empty flagname is specified, internally, it will be skipped and the code will be processed using `_xpflag_dummy` flag name. For now, we need not worry about the uniqueness of the dummy flag name (highly unlikely that someone is going to name a feature flag `_xpflag_dummy`).

swayamraina:
> Yeah I had this thought too, but xpFlagName is initialized like this, which I think is rare enough.
xpFlagName = "_xpflag_dummy"

I completely missed this init part. 
Yeah, agreed this way we reduce a lot of additional computation also.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -2120,4 +2120,183 @@ public void exceptionWhenArgumentIndexNotSpecifiedAndNotArgumentIndexOptional()
...
+            "@Retention(RetentionPolicy.RUNTIME)",
+            "@Target({ElementType.METHOD})",
+            "@interface ToggleTesting {",
+            "  EmptyFlagRemovesAnnotatedMethods.TestExperimentName treated();",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  public void testEmptyFlagRemovesAnnotatedMethodsControl() throws IOException {
```
mkr-plse, line 2243: this test can be removed.

swayamraina: Not sure why this test need to be removed?
This test checks if the complete method deletion (with the annotation) does not happen in case of empty flag-name

mkr-plse: Because the treatment [check](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L857) is already within the conditional [check](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L851) on flagname.  The control [check](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L860) is not testing anything new and increases the testing time.

swayamraina:
> Overall, looks good. Can you also add a test case where ` b.putFlag("Piranha:FlagName", "");` is not passed so that we can be sure that an exception is thrown?

Sorry, did not get you. 
do you mean where the flag config is not present and `PiranhaConfigurationException` is thrown?

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
swayamraina: with `isTreated=false`, `Piranha:FlagName=""` and fix removed I see `piranha` deletes the annotated method whereas it shouldn't

mkr-plse: > > Overall, looks good. Can you also add a test case where ` b.putFlag("Piranha:FlagName", "");` is not passed so that we can be sure that an exception is thrown?
> 
> Sorry, did not get you.
> do you mean where the flag config is not present and `PiranhaConfigurationException` is thrown?

Yes.  A test without flagname to check an exception is correctly thrown is needed here.

mkr-plse:
> > Overall, looks good. Can you also add a test case where ` b.putFlag("Piranha:FlagName", "");` is not passed so that we can be sure that an exception is thrown?
> 
> Sorry, did not get you.
> do you mean where the flag config is not present and `PiranhaConfigurationException` is thrown?

Yes.  A test without flagname to check an exception is correctly thrown is needed here.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
mkr-plse: This doesn't contradict whatever I mentioned above.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -2120,4 +2121,197 @@ public void exceptionWhenArgumentIndexNotSpecifiedAndNotArgumentIndexOptional()
...
+            "@Retention(RetentionPolicy.RUNTIME)",
+            "@Target({ElementType.METHOD})",
+            "@interface ToggleTesting {",
+            "  EmptyFlagRemovesAnnotatedMethods.TestExperimentName treated();",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  public void testPiranhaConfigErrorWhenFlagNotProvided() {
```
mkr-plse, line 2305: See this [test](https://github.com/uber/piranha/blob/91e25d1d4fae55cc67cdc0db65895cf4cae7cad2/java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java#L2055) for an example for checking exception.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -2120,4 +2120,136 @@ public void exceptionWhenArgumentIndexNotSpecifiedAndNotArgumentIndexOptional()
...
+            "Constants.java",
+            "package com.uber.piranha;",
+            "class Constants {",
+            "  public static int ONE = 1;",
+            "}")
+        .doTest();
+  }
+
+  @Test
+  public void testEmptyFlagRemovesAnnotatedMethods() throws IOException {
```
mkr-plse, line 2174: nit: test needs to be either renamed to `testEmptyFlagOnAnnotatedMethods` or `testEmptyFlagDoesNotRemoveAnnotatedMethods`. The latter is descriptive but more verbose. If you go for the former, please add a comment before the test.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -2120,4 +2120,136 @@ public void exceptionWhenArgumentIndexNotSpecifiedAndNotArgumentIndexOptional()
     XPFlagCleaner flagCleaner = new XPFlagCleaner();
     flagCleaner.init(b.build());
   }
+
+  @Test
+  public void testEmptyFlagRemovesStaticImports() throws IOException {
```
mkr-plse, line 2125: nit: test needs to be either renamed to `testEmptyFlagAndStaticImports` or `testEmptyFlagDoesNotRemoveStaticImports`. The latter is descriptive but more verbose. If you go for the former, please add a comment before the test.

java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
swayamraina: I'll add comments with the latter name.

PR #78, opened 2020-07-16
[PiranhaSwift] log error message when command fails to launch

atrakh:


mkr-plse:
Thanks @atrakh for the fix. Can you please sign the CLA so that I can stamp this?

atrakh:
Sure thing, I just signed it. Thanks for the quick followup!

p.s. For posterity, the error I was getting was `SwiftSyntax parser library isn't compatible` and I had to do some digging to figure out that I needed Swift 5.1 to use piranha. 

I used the instructions in this article to switch to swift 5.1: https://medium.com/xcblog/switching-swift-versions-inside-xcode-using-toolchains-755b28831c43

mkr-plse:
> p.s. For posterity, the error I was getting was `SwiftSyntax parser library isn't compatible` and I had to do some digging to figure out that I needed Swift 5.1 to use piranha.
> 
Indeed. Issue 68 is for upgrading PiranhaSwift to use 5.2. Haven't gotten to it yet.

Issue #77, opened 2020-07-15
PiranhaJava: Piranha versions >=0.1.0 do not error out on piranha.properties as expected


lazaroclapp:
The old configuration format has been deprecated. It was intended that PiranhaJava crashed when seeing such configuration file as opposed to the new `.json` format. See https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java#L262

This warning is currently not triggering when deploying Piranha 0.1.0 with the old configuration format, rather Piranha silently fails to pick up any flag API methods from the configuration.

lazaroclapp:
Never mind, this is a duplicate of #76

lazaroclapp:
Closed by #79

PR #79, opened 2020-07-16
[PiranhaJava] Fail hard in the presence of configuration errors.

### LINKED ISSUE #76: Config file not found exception disables PiranhaJava refactoring silently
I am trying the sample java gradle project and after running clean build piranha only removes the SAMPLE_STALE_FLAG enum but not the code checks. This is the output - 

```
public class MyClass {

  enum TestExperimentName {
    //REMOVED FROM HERE BUT NOT BELOW
  }

  private XPTest expt;

  public void foo() {
    if (expt.flagEnabled(TestExperimentName.SAMPLE_STALE_FLAG)) {
      System.out.println("Hello World");
    }

  }

  public void bar() {
    if (expt.flagDisabled(TestExperimentName.SAMPLE_STALE_FLAG)) {
      System.out.println("Hi World");
    }
  }

  static class XPTest {
    public boolean flagEnabled(TestExperimentName x) {
      return true;
    }

    public boolean enableFlag(TestExperimentName x) {
      return true;
    }

    public boolean disableFlag(TestExperimentName x) {
      return true;
    }

    public boolean flagDisabled(TestExperimentName x) {
      return true;
    }
  }
}
```

Here is my gradle file - https://github.com/vihanparmar/testRepo/blob/master/test/build.gradle

Please let me know what i am doing wrong or if this is an issue.

lazaroclapp:
This makes PiranhaConfigurationException a runtime exception,
and gets rid of the `disabled` checker state. In general, it doesn't
make much sense to run Piranha with invalid configuration and
it's better to show the issue to users as visibly as possible,
since either way the tool behavior will be broken in this case.

This resolves #76 and #77.

### DIFF
--- a/java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
+++ b/java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
@@ -2213,10 +2213,8 @@ public class XPFlagCleanerTest {
   @Test
   public void testPiranhaCrashOnNoConfig() {
     // Error Prone turns Runtime Exceptions inside EP into blocks of text (including the exception
-    // trace)
-    // inside a new AssertionError exception. This is actually looking for
-    // PiranhaConfigurationException
-    // inside XPFlagCleaner.
+    // trace) inside a new AssertionError exception. This is actually looking for
+    // PiranhaConfigurationException inside XPFlagCleaner.
     expectedEx.expect(AssertionError.class);
     expectedEx.expectMessage(
         "An unhandled exception was thrown by the Error Prone static analysis plugin");
@@ -2244,10 +2242,8 @@ public class XPFlagCleanerTest {
   @Test
   public void testPiranhaCrashOnOldConfig() {
     // Error Prone turns Runtime Exceptions inside EP into blocks of text (including the exception
-    // trace)
-    // inside a new AssertionError exception. This is actually looking for
-    // PiranhaConfigurationException
-    // inside XPFlagCleaner.
+    // trace) inside a new AssertionError exception. This is actually looking for
+    // PiranhaConfigurationException inside XPFlagCleaner.
     expectedEx.expect(AssertionError.class);
     expectedEx.expectMessage(
         "An unhandled exception was thrown by the Error Prone static analysis plugin");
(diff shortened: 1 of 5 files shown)


java/piranha/src/test/java/com/uber/piranha/XPFlagCleanerTest.java
```diff
@@ -1928,69 +1928,6 @@ public void variableFlagWithReturnAndReceiverAndArgument() throws Exception {
         .doTest();
   }
 
-  /*
-   * Uses "properties_test_invalid.json" instead of "properties.json".
-   * In it, the required top-level "methodProperties" is not present.
-   * As a result, refactoring will not be carried out, except for unreachable code.
-   * */
-  @Test
-  public void refactorUnreachableWhenInvalidConfigFile() throws IOException {
```
lazaroclapp, line 1937: The behavior being tested by this test is no longer intended behavior.

lazaroclapp:
> This doesn't resolve #76 completely. Needs to fix missing config file in sample.

Are you sure? I see that the [author of that issue says it's resolved in Piranha 0.1.0](https://github.com/uber/piranha/issues/76#issuecomment-655685298). And looking at sample, [this task](https://github.com/uber/piranha/blob/master/java/sample/build.gradle#L49) is succeeding and `properties.json` is being passed. Also, no crash is introduced there by this change. 

Am I missing something?

lazaroclapp:
> > This doesn't resolve #76 completely. Needs to fix missing config file in sample.
> 
> Are you sure? I see that the [author of that issue says it's resolved in Piranha 0.1.0](https://github.com/uber/piranha/issues/76#issuecomment-655685298). And looking at sample, [this task](https://github.com/uber/piranha/blob/master/java/sample/build.gradle#L49) is succeeding and `properties.json` is being passed. Also, no crash is introduced there by this change.
> 
> Am I missing something?

Here are the things being found on `sample` right now: https://travis-ci.com/github/uber/piranha/jobs/361875475#L454

mkr-plse:
> > > This doesn't resolve #76 completely. Needs to fix missing config file in sample.
> > Are you sure? I see that the [author of that issue says it's resolved in Piranha 0.1.0](https://github.com/uber/piranha/issues/76#issuecomment-655685298). 

Yes. They were able to resolve it after specifying the full path (see [here](https://github.com/uber/piranha/issues/76#issuecomment-655667800)). 

And looking at sample, [this task](https://github.com/uber/piranha/blob/master/java/sample/build.gradle#L49) is succeeding and `properties.json` is being passed. Also, no crash is introduced there by this change.
> > Am I missing something?

> Here are the things being found on `sample` right now: https://travis-ci.com/github/uber/piranha/jobs/361875475#L454

https://github.com/uber/piranha/issues/76#issuecomment-655667800

PR #81, opened 2020-08-05
Javascript implementation of Piranha

anag004:
This implements PiranhaJS and the details of usage are in the README.


javascript/README.md
```diff
@@ -0,0 +1,156 @@
+# PiranhaJS
+![Node.js CI](https://github.com/anag004/piranha/workflows/Node.js%20CI/badge.svg) ![Lint](https://github.com/anag004/piranha/workflows/Lint/badge.svg)
```
lazaroclapp, line 2: These badges are for the wrong repo now.

javascript/README.md
```diff
@@ -0,0 +1,156 @@
...
+See the suite of tests in `test/test_refactor.js` and `test/test_integration.js` for more detail on how Piranha works and what kind of refactoring it can handle. 
+
+### Running piranha as a command
+
+It can be convenient to run piranha as a shell command from any location. This can be done by creating an alias.
+
+From the *piranha/javascript* directory, run the following shell commands (for the bash shell).
+
+```
+printf "alias piranha='node `realpath src/piranha.js`'" >> ~/.bashrc
```
lazaroclapp, line 152: maybe worth naming this alias `piranhajs`?

javascript/test/control-expected/sample1-featureFlag.js
```diff
@@ -0,0 +1,22 @@
...
+
+for(var i = 0; i == 0; i++){
+    // Simple flag cleanup in conditional
+    f2();
+}
+
+// Deep clean of flag
+var a = f1();
+
+// Assignment cleanup
```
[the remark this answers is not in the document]
anag004: These files correspond to integration tests used to make sure there are no errors in the refactoring pipeline. Exhaustive unit tests can be found in `test/test_refactor.js` and the other `test/test_*` files.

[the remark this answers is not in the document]
lazaroclapp: Never mind, this is the expected output. I get it now.

anag004:
> Not a full review yet (not sure I can do one for JS code :) ), but: I see GitHub actions for running the tests, yet nothing currently running on this PR on our existing Travis CI checks (see below the PR, all that is running are the tests for Java code). Can we make sure CI is setup for this code? Either using GitHub actions or Travis, I don't care, as long as they run before every PR is landable and on every commit in master.

The GitHub actions are running in my fork of the repo. Maybe they need to be merged to `uber/master` before they can run on pull requests?

javascript/src/config_checker.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = {
+    parseProperties: function (properties_json) {
+        let properties;
+        let properties_abs_path = path.resolve(properties_json);
+        try {
+            properties = require(properties_abs_path);
+        } catch (err) {
+            if (err instanceof SyntaxError) {
+                throw new Error(`${properties_json} does not follow JSON syntax`);
+            } else {
```
lazaroclapp, line 12: Should this be a generic `else` or should we be looking for a specific type of FS Error? Should there be a separate message for "something went wrong, no idea what it was"? I can already see, e.g. file access denied not being accounted for in this logic.

javascript/src/config_checker.js
```diff
@@ -0,0 +1,29 @@
...
+            }
+        }
+        if (!("methodProperties" in properties)) {
+            throw new Error(`The methodProperties property is missing in ${properties_json}`);
+        } else {
+            var missingMethod = properties.methodProperties.find(prop => !("methodName", "flagType", "argumentIndex" in prop));
+            missingMethod = JSON.stringify(missingMethod);
+
+            if (missingMethod != null) {
+                throw new Error(`${missingMethod} in ${properties_json} doesn't have all required keys`);
```
lazaroclapp, line 23: If possible, let's report which key is missing. That said, this can easily be a follow up fix 🙂

javascript/test/control-expected/sample-featureFlag.js
```diff
@@ -0,0 +1,25 @@
...
+   f2()   
+}
+
+// ----------------------------------
+
+// Complex cleanup
+var c = f1();
+
+// Literal conversion
+console.log(false);
```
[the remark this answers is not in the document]
anag004: We plan to eventually run a code formatter on all files. You can ignore minor code styling issues in your review.

javascript/src/piranha.js
```diff
@@ -0,0 +1,138 @@
+// Contributed by Ananye Agarwal (anag004) and Sarthak Behera (s7rthak). The authors are from the Indian Institute of Technology, Delhi (IIT Delhi).
```
lazaroclapp, line 1: Absolutely happy to list this!

That said, as an Uber OSS requirement, every JS file in this PR should have the Uber copyright block and link to the license at the top (above this line). See the Java example [here](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/PiranhaUtils.java#L1)

javascript/src/piranha.js
```diff
@@ -0,0 +1,138 @@
...
+    {
+        help: "If this option is supplied, the flag is treated, otherwise it is control.",
+        nargs: 'OPTIONAL',
+    }
+);
+
+parser.addArgument(
+    ["-n", "--max_cleanup_steps"], 
+    {
+        help: "The number of times literals should be simplified. Runs until fixed point by default.",
```
lazaroclapp, line 63: Cool!

javascript/src/refactor.js
```diff
@@ -0,0 +1,648 @@
...
+            methodHashMap.set(item["methodName"], item);
+        });
+    
+        return methodHashMap;
+    }
+
+    // Verify this is a literal introduced by Piranha. 
+    // We only refactor code associated with literals having this property and leave original code untouched
+    isPiranhaLiteral(node) {
+        return node.type === "Literal" && node.createdByPiranha !== undefined;
```
lazaroclapp, line 64: Cool. Wish I could annotate AST nodes as generated by Piranha like this on the Java side 😉

javascript/src/refactor.js
```diff
@@ -0,0 +1,648 @@
...
+                
+            if (literal.value === false) {
+                return expression;
+            }
+        }
+    }
+
+    // Check if a function with return statement at the end of function body is redundant, i.e returns piranhaLiteral
+    checkAndAddRedundantFunction(func, name, redundantFunctions) {
+        if (func.body === null || func.body.body == null) {
```
lazaroclapp, line 89: Curious here, what does `func.body.body` mean here and in the rest? Why not just `func.body`?

javascript/src/refactor.js
```diff
@@ -0,0 +1,648 @@
...
+            return false;
+        else {
+            var returnNode = func.body.body[returnIndex];
+
+            if (returnNode.type === "ReturnStatement" && 
+                returnNode.argument !== null && 
+                this.isPiranhaLiteral(returnNode.argument) && 
+                typeof(returnNode.argument.value) === "boolean") { 
+                    redundantFunctions[name] = returnNode.argument.value; 
+            }
```
lazaroclapp, line 105: What happens here with code like:

```
foo(...) {
    if(flagOne) {
        return X;
    }

    if(flagTwo) {
        return Y;
    }
}
```

When simplifying `flagTwo`? I feel it will ignore `flagOne` and simplify that to a redundant function, unless I am missing something...

lazaroclapp: Actually. I see below that only functions with a single return statement will get passed to this method, which solves that issue. Maybe worth a comment reminding the reader of that here 🙂

javascript/src/refactor.js
```diff
@@ -0,0 +1,648 @@
...
+
+            this.evalBoolExpressions();
+            this.reduceIfStatements();
+
+            var redundantVarnames = this.getRedundantVarnames();
+            this.pruneVarReferences(redundantVarnames);
+
+            var functionsWithSingleReturn = this.getFunctionsWithSingleReturn();
+            var redundantFunctions = this.getRedundantFunctions(functionsWithSingleReturn)
+            this.pruneFuncReferences(redundantFunctions);
```
lazaroclapp, line 617: So, perhaps a naive question (remember, I don't do much JS coding) but: this version of Piranha operates on one `*.js` file at the time, so, what happens with exported functions that are now reduced to a constant but called from other JS files? Do we do deep clean across files? And if not (which is totally fine!), do we have any check to prevent functions used by other JS files from being removed?

javascript/test/control-expected/sample-featureFlag.js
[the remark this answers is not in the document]
lazaroclapp: Please do run the auto-formatter before landing this PR (probably would have been good to run it before reviews too, for readability, but not sure if it will mess existing PR comments at this point). Also, consider integrating the auto-formatter to run on every commit and check in CI as a follow up PR (not on this one, though, this is already a large code change 🙂 )

javascript/src/config_checker.js
anag004: This was pointed out by @mkr-plse as well. We seem to have missed incorporating the change while merging a PR in our working repo. Will do it now. Thanks for pointing it out.

javascript/src/refactor.js
anag004: Not a naive question at all. In its current form, PiranhaJS ignores `import` statements and refactors each source file independently. That said, refactoring across files is one of several enhancements we plan to introduce in the future.

javascript/src/refactor.js
anag004: This is a quirk of nomenclature. The `body` of a `FunctionDeclaration` refers to the contents of the function. When a function is enclosed within braces `{ ... }`, its `body` is a `BlockStatement` node. The `BlockStatement` itself contains a `body` attribute which is an array of nodes corresponding to each line of code. Therefore, `func.body.body` refers to this array.

javascript/test/control-expected/sample-featureFlag.js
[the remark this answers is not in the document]
mkr-plse: May be https://github.com/prettier/prettier?

javascript/src/refactor.js
lazaroclapp: Actually, it's fine to leave it like that for this PR. Maybe worth adding a small note in the README that this limitation exists and that Piranha JS is fairly experimental right now. 

Cross-file refactoring would be a lot more than most (all?) other versions of Piranha do, though, so that'd be cool! 

Medium term, my only worry here is people running this and it breaking their code because it removed a function called from another file. Specially annoying for JS, I think, since there is a chance such breakage wouldn't be detected except at runtime/prod.

mkr-plse: Good point. Ideally, integration tests of the target codebase should catch this. It will be useful to discuss this point in the README (not as a limitation but by explaining the current features of deep cleaning and the boundaries).

javascript/config/properties.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "methodName": "isFlagTreated",
+      "flagType": "treated",
+      "argumentIndex": 0
+    },
+    {
+      "methodName": "isToggleEnabled",
+      "flagType": "treated",
+      "argumentIndex": 0
+    }
```
mkr-plse, line 17: let us include the `empty` flagType also.

javascript/src/config_checker.js
s7rthak: Okay we will add in the case where error is not identified then user gets the error message generated by machine.

javascript/test/control-expected/sample-featureFlag.js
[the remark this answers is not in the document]
anag004: Done 78adb312677f136e9f7d41027305280007494fd8.

javascript/test/test_refactor.js
```diff
@@ -1,621 +1,971 @@
...
+      engine.finalizeLiterals();
+
+      const refactored_code = recast.print(ast).code;
+
+      assert(engine.changed);
+      assert(
+        expected_code === refactored_code,
+        `\nEXPECTED : ${JSON.stringify(
+          expected_code
+        )}\nREFACTORED : ${JSON.stringify(refactored_code)}`
```
anag004, line 104: Prettier may refactor some code in undesirable ways. For example, this way of writing the output string seems to be unnecessarily confusing.

javascript/README.md
anag004: Done d840b4c6c66b9c0090b2c548971f8a2a93ac2e25

javascript/README.md
anag004: Done 48929a4f680bf1a451432976cfaac1481f6ccc3e

javascript/src/piranha.js
anag004: Done a8ce1614a7d21ea3ca8b542891847a6746ef550d

javascript/src/config_checker.js
anag004: Fixed 7d274de83d26b96acb5de427115b208cfc7db26c

javascript/config/properties.json
anag004: Is it possible to shelve this for the (not too distant) future since in order to implement deep cleaning we will need to modify all functions in `src/refactor.js` and write new tests as well? Alternatively, for this PR we can do a simple replacement `flagAPI(flag) -> null` in `flagAPIToLiteral()` and avoid deep cleaning.

javascript/src/refactor.js
anag004: Done 7d274de83d26b96acb5de427115b208cfc7db26c

javascript/src/config_checker.js
s7rthak: Handled in [329ed90](https://github.com/uber/piranha/pull/81/commits/329ed90f3a53b13a68fa53b97a028aafe6c72abe)

javascript/config/properties.json
mkr-plse: Let us shelve it. No problem.

javascript/package.json
```diff
@@ -0,0 +1,53 @@
...
+        "type": "git",
+        "url": "git+https://github.com/anag004/piranhaJS.git"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "bugs": {
+        "url": "https://github.com/anag004/piranhaJS/issues"
+    },
+    "homepage": "https://github.com/anag004/piranhaJS#readme",
```
mkr-plse, line 36: the hardcoded links need to be updated.

anag004:
@msridhar any particular style of documentation that needs to be followed?

msridhar:
I am not particular. For me regular comments are fine or you can use JSDoc. If others are more particular they can comment

mkr-plse:
> I am not particular. For me regular comments are fine or you can use JSDoc. If others are more particular they can comment

Thanks. Let us go with regular comments for now then.

anag004:
Added docs in 2f7bfe9. Please take a look.

msridhar:
Docs are good!  I'll review again after my other comments (unused esprima depth, realpath) are addressed.

s7rthak:
`recast` internally uses `esprima` as well as other parsers, so `esprima` is not unused but is installed as a dependency for `recast`. Changed the alias  [e27407d](https://github.com/uber/piranha/pull/81/commits/e27407d609ec6f27890e7d346d1707c5ea667545)

mkr-plse:
> LGTM!

Thanks @msridhar for your review.

mkr-plse:
> LGTM

Thanks @lazaroclapp for the review.

anag004:
Thanks @lazaroclapp, @msridhar for your reviews.

PR #82, opened 2020-08-08
Code format

s7rthak:
@mkr-plse Is code formatting before every commit needed? This branch has pre-commit hooks which runs `prettier` on files before every commit.


s7rthak:
To elaborate on the pre-commit hook, it uses `husky` to run scripts before commit. It also uses `lint-staged` to only apply `prettier` on staged files. Only the `.js` files in `src/test` directory are affected by the script if they are modified and staged. `husky`, `prettier` and `lint-staged` are only present in `devDependencies` so they will only be installed with `npm i --dev`.

s7rthak:
merge issue resolution is difficult so moving on to another PR.

PR #84, opened 2020-08-09
[PiranhaJS] Code formatting using pre-commit hook 

s7rthak:
@mkr-plse The other PR became unwieldy due to various resolution issues. This might be better.
To elaborate on the pre-commit hook, it uses `husky` to run scripts before commit. It also uses `lint-staged` to only apply `prettier` on staged files. Only the `.js` files in `src`/`test` directories are affected by the script if they are modified and staged. `husky`, `prettier` and `lint-staged` are only present in `devDependencies` so they will only be installed with `npm i --dev`.


msridhar:
Is there a job on CI that is checking for correct formatting?

s7rthak:
There are no formatting checks as of now. The pre-commit hook maybe alleviates that need as files are automatically formatted before each commit so that they remain in good shape.

msridhar:
Pre-commit hooks cannot provide guarantees as they are easily skipped. If
we want to maintain formatting standards I think there should be a CI job
that checks formatting, with pre-commit hooks as an additional convenience.
/cc @lazaroclapp

s7rthak:
Added in an action in https://github.com/uber/piranha/pull/84/commits/ac91215e2454300d63819ea6443e908f2ae89618 to format code which fails as `package.json` has not been changed but works on fork.

lazaroclapp:
> Added in an action in [ac91215](https://github.com/uber/piranha/commit/ac91215e2454300d63819ea6443e908f2ae89618) to format code which fails as `package.json` has not been changed but works on fork.

Can you clarify here. Is there anything fundamentally preventing that new CI task from succeeding on this PR? If this is landed, will master be clean with respect to that `Format` check?

s7rthak:
So I looked into the issue and found that `actions/checkout@v2` was pointing to the last commit on `uber/piranha` as shown below.
Screenshot from 2020-08-10 23-35-03
But that commit doesn't have the script `prettier` which is needed for pre-commit hooks. So I will push a commit to PR deleting the `with` property of `actions/checkout@v2` and that should fix this issue.

.github/workflows/format.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "14.x"
+      - name: Format
+        run: |
+          npm ci
+          npm run prettier
```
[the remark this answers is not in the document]
msridhar: Re-upping this: I think it will be confusing to have a CI task pushing formatting changes to a PR.  The author will then have to merge from GitHub before adding further changes to the PR and it could lead to weird conflicts.  I think this should just be a check that fails if the code is not formatted.  If that is not possible, probably the CI job should just be removed for now (though then the guarantees about having formatted code are lost).

.github/workflows/format.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4.1.2
+        with:
+          commit_message: Apply formatting changes
+          branch: ${{ github.head_ref }}
```
lazaroclapp, line 28: If the problem before was that using this reference meant we were looking at master? Doesn't this mean the formatting changes will be landed to master? Can we double check this?

In general, it would probably be better not to try to auto-fix the code here, but have two levels:

a) A commit hook to auto-format code LOCALLY (or a script that can be run to do so)
b) A CHECK in CI that will prevent a PR from landing if not formatted correctly, but won't actually attempt to change code.

.github/workflows/format.yml
lazaroclapp: I also vote that this should be a check, without auto-commit. Specially since I am not sure if `github.head_ref` can point to something other than PR (e.g. master)

.github/workflows/format.yml
s7rthak: It was stupid of me to write auto-fix code script in a check. I have added in another script `format` which only checks for formatting error and if there are some issues exits with error code `1`.

.github/workflows/format.yml
s7rthak: @msridhar Done in https://github.com/uber/piranha/pull/84/commits/443a91e133573f9916458ed35c2885914dad8f9a Auto-fixing code was a bad idea as it pushes to master so changed `--write` to `--check`.

javascript/package.json
```diff
@@ -10,7 +10,8 @@
     "scripts": {
         "test": "mocha",
         "only-test": "mocha --inspect-brk test/test_refactor.js",
-        "prettier": "prettier --write './{src,test}/**/*.js'"
+        "prettier": "prettier --write './{src,test}/**/*.js'",
+        "format": "prettier --check './{src,test}/**/*.js'"
     },
```
anag004, line 15: Minor suggestion - can we have more helpful names for these? It is not immediately clear which of `format` and `prettier` is a check and which actually formats code. Maybe `format` can be renamed to `check_format`.

s7rthak: Done https://github.com/uber/piranha/pull/84/commits/eb3b0ae92f2aa59d565d45e5487d9a9512f1724a

.github/workflows/format.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "14.x"
+      - name: Format
```
lazaroclapp, line 20: Can this name also be changed to Check Format (Is this just the display name of the check in the GitHub/PR CI? )

s7rthak: Yes it is the display name of check. I changed the name of its job to `check_format` for consistency.

anag004:
@s7rthak since it's your PR, leaving it to you to merge. Before merging you can optionally add a `code style: prettier` badge to the README.

s7rthak:
@anag004 Okay I will merge.

PR #90, opened 2020-08-12
[PiranhaJS] Upgrade serialize-javascript to 3.1.0

mkr-plse:
This fixes the dependabot alert for [CVE-2020-7660](https://github.com/advisories/GHSA-hxcc-f52p-wc94).

### DIFF
--- a/javascript/package-lock.json
+++ b/javascript/package-lock.json
@@ -1705,7 +1705,7 @@
         "ms": "2.1.2",
         "object.assign": "4.1.0",
         "promise.allsettled": "1.0.2",
-        "serialize-javascript": "3.0.0",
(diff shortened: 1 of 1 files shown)


s7rthak:
@mkr-plse Can we identify dependency issues before commit through a check?

lazaroclapp:
Happy with the contents of the PR, but please change title to "[PiranhaJS] Upgrade serialize-javascript to 3.1.0" and then mention the dependabot alert in the description of the PR/commit, as otherwise the title is not descriptive of the change.

And, @s7rthak, not sure if there is a way to run the dependabot logic on PRs, but worth looking into that.

PR #91, opened 2020-08-13
[PiranhaJS] Add support for string literal arguments

atrakh:
Adds a check for string literal arguments in flag APIs in addition to identifiers. This implementation searches for both, but I can see it making sense to let the user configure which they'd like to search for.

See example added in tests

p.s. I couldn't easily figure out how to clean up my commit history, can we squash it if this PR gets merged?


mkr-plse:
> Adds a check for string literal arguments in flag APIs in addition to identifiers. This implementation searches for both, but I can see it making sense to let the user configure which they'd like to search for.

Thanks @atrakh for the PR. In your experience, what is the popular usage style in JS? We can have PiranhaJS  use that style (literal/identifier) as the default. If necessary, that can be overridden by an option in [properties.json](https://github.com/uber/piranha/blob/master/javascript/config/properties.json)?  

> p.s. I couldn't easily figure out how to clean up my commit history, can we squash it if this PR gets merged?

The commits will be squashed before a merge.

atrakh:
> In your experience, what is the popular usage style in JS? We can have PiranhaJS use that style (literal/identifier) as the default.

Internally we use string literal calls for JS, but I can see both methods being common. Ideally, every Piranha implementation would support both.

> If necessary, that can be overridden by an option in properties.json?

My initial thought was to make this a command line argument, since usage can very on a per-flag or per-file basis (e.g. using an identifier to call a flag API in one file, and a literal in another). Perhaps we can search for both by default, and add a new argument to limit search to one node type?

---
Side note: Is it planned for Piranha to support refactoring/removing the declaration of the identifier argument?
e.g. removing all of the code in this block, instead of just the conditional block:
```javascript
const featureFlag = "feature-flag"
if (isFlagTreated(featureFlag)) {
 // do stuff
}
```

If we had that functionality, we could likely make the assumption that the configured API argument index always refers to a string literal and expect the flag name to be a string literal node. Being able to search across multiple files/imports would likely be required for some of the more complex usages of feature flags

javascript/test/test_refactor.js
```diff
@@ -35,22 +35,37 @@ describe('refactor', () => {
...
+                const engine = new refactor.RefactorEngine(ast, properties, treated, flagname);
+                engine.flagAPIToLiteral();
+                engine.finalizeLiterals();
+
+                const refactoredCode = recast.print(ast).code;
+
+                assert(engine.changed);
+                assert(
+                    spec.expectedCode === refactoredCode,
+                    `\nEXPECTED : ${JSON.stringify(spec.expected_code)}\nREFACTORED : ${JSON.stringify(
```
anag004, line 64: Shouldn't this be `spec.expectedCode` instead of `spec.expected_code`? Actually, I think it's better if we use underscore casing throughout (i.e. rename `spec.expectedCode` to `spec.expected_code` everywhere) since that's what has been used in the rest of the tests.

anag004:
> p.s. I couldn't easily figure out how to clean up my commit history, can we squash it if this PR gets merged?

@mkr-plse How will squashing work given that this PR spans implementations? 

> My initial thought was to make this a command line argument, since usage can very on a per-flag or per-file basis 

@atrakh To be clear, does this imply that the same flag can occur as both a `String` and a `Boolean` in the same codebase? If not, then perhaps it would be better to search for a single flag type at a time and to configure this in `properties.json` as @mkr-plse mentioned since the type of a flag then becomes an immutable property of the flag itself.

javascript/examples/sample.js
```diff
@@ -19,6 +19,15 @@ if (isFlagTreated(featureFlag)) {
     f2();
 }
 
+// ---------------------------------
+
```
s7rthak, line 23: Can you please remove this comment line here? Other than that the code looks good!

s7rthak:
> > In your experience, what is the popular usage style in JS? We can have PiranhaJS use that style (literal/identifier) as the default.
> 
> Internally we use string literal calls for JS, but I can see both methods being common. Ideally, every Piranha implementation would support both.
> 
> > If necessary, that can be overridden by an option in properties.json?
> 
> My initial thought was to make this a command line argument, since usage can very on a per-flag or per-file basis (e.g. using an identifier to call a flag API in one file, and a literal in another). Perhaps we can search for both by default, and add a new argument to limit search to one node type?
> 
> Side note: Is it planned for Piranha to support refactoring/removing the declaration of the identifier argument?
> e.g. removing all of the code in this block, instead of just the conditional block:
> 
> ```js
> const featureFlag = "feature-flag"
> if (isFlagTreated(featureFlag)) {
>  // do stuff
> }
> ```
> 
> If we had that functionality, we could likely make the assumption that the configured API argument index always refers to a string literal and expect the flag name to be a string literal node. Being able to search across multiple files/imports would likely be required for some of the more complex usages of feature flags

We will look into implementing the side note.

anag004:
> Is it planned for Piranha to support refactoring/removing the declaration of the identifier argument?
e.g. removing all of the code in this block, instead of just the conditional block:
>```javascript
>const featureFlag = "feature-flag"
>if (isFlagTreated(featureFlag)) {
> // do stuff
> }
> ```

This is probably a naive question but what is the point of this code pattern? Why not simply use `isFlagTreated('feature-flag')`?

mkr-plse:
> > p.s. I couldn't easily figure out how to clean up my commit history, can we squash it if this PR gets merged?
> 
> @mkr-plse How will squashing work given that this PR spans implementations?

There is a `squash and merge` as an option while merging. 
> 
> > My initial thought was to make this a command line argument, since usage can very on a per-flag or per-file basis
> 
> @atrakh To be clear, does this imply that the same flag can occur as both a `String` and a `Boolean` in the same codebase? If not, then perhaps it would be better to search for a single flag type at a time and to configure this in `properties.json` as @mkr-plse mentioned since the type of a flag then becomes an immutable property of the flag itself.

Correct. We may be able to configure each flagAPI accordingly. We can follow up in a different PR.

> This is probably a naive question but what is the point of this code pattern? Why not simply use isFlagTreated('feature-flag')?

Since developers have the flexibility in using APIs, it is conceivable that pattern suggested by @atrakh is common. For example, definition be present in one file and the API call in a different file and the variable being able to maintain the map between flag names.  Created #92 to track this further.

mkr-plse:
> > @mkr-plse How will squashing work given that this PR spans implementations?
> 
> There is a `squash and merge` as an option while merging.

To clarify further, @anag004 was actually referring to the presence of a commit corresponding to PiranhaSwift in the list of commits here which I missed. Since that commit is already in master, nothing needs to be done here additionally.

javascript/test/test_refactor.js
atrakh: Good catch - I'm not sure how this test passed locally...

My JS instincts kicked and I unconsciously wrote this camel case -- I'll comb through this PR and make sure it's only using snake case.

mkr-plse: This would have been a problem only if the refactoring failed. Since the refactoring is correct, the tests passed. 
Impressive that this was caught in code review 💯

atrakh:
> To be clear, does this imply that the same flag can occur as both a String and a Boolean in the same codebase? If not, then perhaps it would be better to search for a single flag type at a time and to configure this in properties.json as

Sorry, I may not be following completely. To clarify what I meant, when referring to the type, I'm specifically referring to the symbol used to identify the flag in calls to APIs. Would it be reasonable to assume the identifier for a feature flag is almost always string, that may be abstracted in some way?

My concern with using `properties.json` to configure this is that if a codebase does happen to have multiple patterns for using the same flag API, we would need to change the `properties.json` each time we scan a new flag/file combo. A command line option would be a bit more flexible, allowing the consumer to switch types on the fly.

> Since developers have the flexibility in using APIs, it is conceivable that pattern suggested by @atrakh is common. For example, definition be present in one file and the API call in a different file and the variable being able to maintain the map between flag names. Created #92 to track this further.

Spot on. Thank you for clarifying @mkr-plse.

javascript/test/test_refactor.js
atrakh: 🤦  Yup, that typo is in the assertion failure message. Thanks all

mkr-plse:
> > To be clear, does this imply that the same flag can occur as both a String and a Boolean in the same codebase? If not, then perhaps it would be better to search for a single flag type at a time and to configure this in properties.json as
> 
> Sorry, I may not be following completely. To clarify what I meant, when referring to the type, I'm specifically referring to the symbol used to identify the flag in calls to APIs. Would it be reasonable to assume the identifier for a feature flag is almost always string, that may be abstracted in some way?
> 
> My concern with using `properties.json` to configure this is that if a codebase does happen to have multiple patterns for using the same flag API, we would need to change the `properties.json` each time we scan a new flag/file combo. A command line option would be a bit more flexible, allowing the consumer to switch types on the fly.

My initial understanding was that flags used in specific APIs were either identifiers or literals and that needs to be configured. But, if for the same flag API, it can be used either as string or literal depending on the file, then specifying it as a command line option will be helpful. Created #93 to discuss further.

PR #97, opened 2020-09-15
[PiranhaJava] Add DisabledUnlessConfigured option.

lazaroclapp:
This option tells Piranha to explicitly turn itself off unless passed
ANY OTHER Piranha related configuration option. It will still report
an error on partial configuration, but silently disable itself if
this is the only parameter it receives.

This option is required for large multi-targets builds which contain
a global set of targets to be built (G), of which only a subset
loads the Piranha EP checker (L) and only a subset of those get
passed proper Piranha configuration (C).

Without this option, the checker will fail with a
`PiranhaConfigurationException` for all targets in (L - C), since
correct configuration isn't provided. Furthermore, passing
`-Xep:Piranha:OFF` everywhere unless a configuration is passed
and overriding that for C with e.g. `-Xep:Piranha:WARN` is not
a solution. The later will result in an exception, since the
targets in (G - L) won't recognize the name "Piranha" as a checker.

In particular, the Uber build configuration becomes much simpler
using this option.

### DIFF
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -130,6 +130,15 @@ public class XPFlagCleaner extends BugChecker
    */
   private boolean initialized = false;
 
+  /**
+   * When no configuration whatsoever is provided, and -Xep:Piranha:DisabledUnlessConfigured=true,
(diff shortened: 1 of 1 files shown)


lazaroclapp:
@mkr-plse This is a necessary change to get Piranha master to run on our internal build. We need this, plus the fix to the issue from yesterday before we cut the next release. Will be working on a second PR for that other part.

mkr-plse (approved):

PR #100, opened 2020-09-30
[PiranhaJava] Support array expressions in ExperimentTest annotations

lazaroclapp:
This adds support for `@ExperimentTest` style annotations which
take an array expression rather than a single value to their
`treated` parameter.

Before this change, analyzing such annotation would result in an
NPE.

In terms of the semantics/intended refactoring, consider code like:

```
@ExperimentTest(treated = {XP_FLAG_1,XP_FLAG_2,XP_FLAG_3})
public void foo() {
  ...
}
```

And `XP_FLAG_1` being stale.

If the flag is **treated**, Piranha should remove it from the list but
keep the annotation:

```
@ExperimentTest(treated = {XP_FLAG_2,XP_FLAG_3})
public void foo() {
  ...
}
```

If the flag is in **control**, the entire test should be removed
as it requires all three flags to be in treatment mode.

Additionally, this code:

```
@ExperimentTest(treated = {XP_FLAG_1})
public void foo() {
  ...
}
```

Should be treated as equivalent to:

```
@ExperimentTest(treated = XP_FLAG_1)
public void foo() {
  ...
}
```

### DIFF
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -45,6 +45,7 @@ import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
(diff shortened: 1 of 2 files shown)


mkr-plse (approved):
lgtm

PR #107, opened 2020-10-29
[Approach PR] [Not ready for review] - Adding support for mocks in test cleanup

sivacoder:
This PR is to discuss the possibilities of cleaning up mocks in tests. Approach seems very specific to few use-case and want to discuss approaches here with code.

1. Code not prod ready
2. Tests not added

### DIFF
--- a/java/piranha/config/properties.json
+++ b/java/piranha/config/properties.json
@@ -41,5 +41,12 @@
   "linkURL": "<provide_your_url>",
   "annotations": [
     "ToggleTesting"
+  ],
+  "testMethodProperties" : [
+    {
+      "methodName": "when",
+      "flagType": "empty",
+      "argumentIndex": 0
+    }
   ]
 }
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -69,6 +69,7 @@ import java.util.Optional;
 import java.util.Set;
 import javax.annotation.Nullable;
 import javax.lang.model.element.ElementKind;
+
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
 import org.json.simple.parser.ParseException;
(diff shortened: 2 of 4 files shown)


java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -881,6 +899,45 @@ private boolean isCheckedXPFlagName(ExpressionTree tree) {
   public Description matchMethod(MethodTree tree, VisitorState state) {
     if (disabled) return Description.NO_MATCH;
 
+    if (!configTestMethodProperties.isEmpty() && tree != null && tree.getBody() != null && tree.getBody().getStatements() != null) {
```
sivacoder, line 902: @mkr-plse - This entire logic is very specific and I'm not happy with this approach. 
But need your inputs to get this better and generic

java/piranha/config/properties.json
```diff
@@ -41,5 +41,12 @@
   "linkURL": "<provide_your_url>",
   "annotations": [
     "ToggleTesting"
+  ],
+  "testMethodProperties" : [
```
sivacoder, line 45: This will act as the configuration for the test methods that needs to be cleanup. Structure is the same as `methodProperties`

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -881,6 +899,45 @@ private boolean isCheckedXPFlagName(ExpressionTree tree) {
...
+      List<? extends StatementTree> bt = tree.getBody().getStatements();
+      for (StatementTree st : bt) {
+        if (st != null && st.getKind().equals(Kind.EXPRESSION_STATEMENT)) {
+          ExpressionStatementTree est = (ExpressionStatementTree) st;
```
mkr-plse, line 906: 1. Shouldn't the statement count here be just one? Can we avoid the loop? 
2. How about calling `evalExpr(st)`? If not, `evalExpr(mit)` below and use the result of that. 
3. Before calling the`evalExpr`, call `getXPAPI` and see whether the API is `Mock` method ? You may also want to update `getXPAPI` to do the necessary processing to check for `when` and return the appropriate type. 

So, the code will be: 
```
API api = getXPAPI(mit) // may be st too.
if(api.equals(API.MOCK)) {
     Value value = evalExpr(mit) // may be st too and update evalExpr
      if(value != Value.BOT) {
         // code starting at line 922. 
      }
}
```

java/piranha/config/properties.json
mkr-plse: Just change the flagType for the method. So, 
```
{
      "methodName": "when",
      "flagType": "mock",
      "argumentIndex": 0
 }
```

PR #111, opened 2021-02-20
Update SwiftSyntax version to swift-5.3-RELEASE

mkr-plse:


.travis.yml
```diff
@@ -20,8 +20,8 @@ matrix:
 
     - language: swift
       os: osx
-      osx_image: xcode11.6
-      env: SWIFT=5.2
+      osx_image: xcode12.2
+      env: SWIFT=5.3
```
elsh, line 24: you can probably delete this, as specifying xcode version should be sufficient.

swift/Package.swift
```diff
@@ -25,7 +25,7 @@ let package = Package(
         .macOS(.v10_14),
```
elsh, line 25: this needs to be .v10_15

lazaroclapp (approved):
Makes sense to me :)

PR #116, opened 2021-04-10
Integrating Swift Argument Parser

chiragramani:
**Context:** 
The PR aims to integrate [Swift Argument Parser](https://github.com/apple/swift-argument-parser) with Piranha Swift. Swift Argument Parser is now a standard way to accept input from command line and it abstracts all the code to provide the documentation for input commands, custom naming flags, out of box support for flags/arguments/options with transformation that really helps to provide a robust command line input taking away all the pain of manual handling.

Even other Uber swift repos such as CodeSan makes use of the same. This will also help us to have have parity with js piranha and will also help us to easily add new commands or modify existing commands as Piranha evolves.

1. The PR also addresses other minor changes. Keeping the naming of swift directories very much similar to what default SPM packages do.
2. A new target dependency PiranhaKit because test bundle cannot import an executable target and therefore to satisfy the testing needs, this new module was needed.
3. Added unit-tests for the newly added functionality as well as updated the test.sh
4. Also updated the README accordingly. 

Please note: This PR doesn't aim to provide any change in the existing functionality related to AST rewriting.


swift/Sources/PiranhaKit/InputCommand/PiranhaCommand.swift
```diff
@@ -0,0 +1,88 @@
...
+                                                           defaultSubcommand: CleanupStaleFlagsCommand.self)
+}
+
+struct CleanupStaleFlagsCommand: ParsableCommand {
+    
+    static var _commandName: String = "cleanup-stale-flags"
+    
+    @Option(name: [.customShort("s"),
+                   .long],
+            help: "Input Source File that needs to run piranha",
```
mkr-plse, line 35: Input source file for Piranha?

mkr-plse (approved):
Looks good. Minor nit on the help message. Thanks for taking care of this.

swift/Sources/PiranhaKit/InputCommand/PiranhaCommand.swift
chiragramani: Yeah, that does seem better to me too. Thanks for suggesting this - will make the changes!

chiragramani: Updated!

mkr-plse (approved):

PR #125, opened 2021-05-05
[PiranhaJava] Switch Java CI to GitHub Actions.

lazaroclapp:
Travis CI is about to be sunset for Uber OSS projects, we should switch PiranhaJava and PiranhaSwift CI builds to use GitHub Actions, just as PiranhaJS already does (see also #124). This accounts for the PiranhaJava portion.

This required a number of changes to our Gradle build. Most importantly, a long pending upgrade from Gradle 4.9 to 6.8.3, but also configuration to run the sample project on JDK11, fixes to our code coverage reporting, and an update to our Google Java Format plugin.

Issue #124, opened 2021-05-05
Migrate PiranhaJava and PiranhaSwift CI to use GitHub Actions


lazaroclapp:
Travis CI is deprecated for Uber OSS and soon to be sunset, we should switch our CI to be GitHub Actions uniformly (PiranhaJS is already there)

lazaroclapp:
cc: @mkr-plse @rajbarik for visibility

lazaroclapp:
All done.

PR #126, opened 2021-05-06
[PiranhaSwift] Switch Swift CI to GitHub Actions

lazaroclapp:
Travis CI is about to be sunset for Uber OSS projects, we should switch PiranhaJava and PiranhaSwift CI builds to use GitHub Actions, just as PiranhaJS already does (see also #124). This accounts for the PiranhaSwift portion. See #125 for the equivalent PiranhaJava change.


mkr-plse (approved):
lgtm. thanks!

PR #127, opened 2021-05-06
Remove/deprecate Travis CI

lazaroclapp:
No longer needed, due to our migration to GitHub actions. See #125 and #126.

Additionally, Uber OSS no longer supports Travis CI.


mkr-plse (approved):

lazaroclapp:
We get more coverage out running the test suite with JDK 11, but given this is still primarily used on Android code, I think we should track the coverage based on the JDK 8 test suite (hence the apparent "reduction" in coverage, Travis was using JDK 11, GitHub actions is getting the info from JDK 8). Note, this affects only coverage reporting, we are running (and blocking on!) the full test suite for both JDK versions.

PR #128, opened 2021-05-07
[WIP] [PiranhaJava] Support fixing enum fields with trailing semicolon

lazaroclapp:
(There has to be a better way of doing this, just making a PR so we can track this and a possible solution)


ketkarameya:
@lazaroclapp I think this recent [PR](https://github.com/uber/piranha/pull/159) solves a similar problem.

PR #131, opened 2021-05-11
Checking for asString in MemberAccessExprSyntax

chiragramani:
**Context:** The following usage ```cachedExperiments.isTreated(forExperiment: ExperimentNamesSwift.test_experiment.asString)``` is not refactored by Piranha.

This PR basically:

1. analyses the MemberAccessExprSyntax description and checks if its == "asString".
2. If it's equal, then it takes into account the previous token's description which in the above case is the experiment identifier that needs to be considered for refactoring.
3. The InputSampleFiles (testfile.swift, control.swift and treated.swift) have been updated accordingly.


mkr-plse (approved):
lgtm.

mkr-plse:
Thanks @chiragramani and @lazaroclapp.

PR #133, opened 2021-05-13
Scoped label replacements

chiragramani:
**Context:**
```swift
let xpName = ExperimentNames.someStaleFlag
return cachedExperiments.isTreated(forExperiment: xpName)
````
is not considered for refactoring. 

**This PR:**
1. Adds support to enable refactoring for the above and related use-cases.
2. It also considers scope into account. For ex, a local scope variable value(if accessible) should override the higher scope variable value. This makes sure that as FuncDeclSyntax nodes are being entered/existed, the scope resolution provides enough context to fill in the suitable values (considering the closest context to avoid ambiguities).
3. As per the above scope based rules, the sample input files(testfile.swift, treated.swift, control.swift) are updated with suitable test-cases.

### DIFF
--- a/swift/Sources/PiranhaKit/CleanupStaleFlags/StaleFlagCleaner.swift
+++ b/swift/Sources/PiranhaKit/CleanupStaleFlags/StaleFlagCleaner.swift
@@ -45,6 +45,8 @@ class XPFlagCleaner: SyntaxRewriter {
 
     private var previousTrivia: Trivia = []
     private var caseIndex: Int?
+    
+    private var scopes = [Scope]()
 
     private let flagName: String
     private let groupName: String
@@ -120,6 +122,11 @@ class XPFlagCleaner: SyntaxRewriter {
                     return true
                 }
             }
+            
+            if let expr = ExprSyntax.init(Syntax(argument.expression)),
+               labelReplacementAllowed(forIdentifier: expr.description) {
+                return true
+            }
             if let expr = StringLiteralExprSyntax.init(Syntax(argument.expression)) {
                 if name == expr.description.replacingOccurrences(of: "\"", with: "") {
                    return true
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: swift/Sources/PiranhaKit/CleanupStaleFlags/StaleFlagCleaner.swift [hunk 1]
```diff
@@ -822,3 +863,33 @@ private extension TriviaPiece {
         return true
     }
 }
+
+private extension XPFlagCleaner {
+    
+    
+    final class Scope {
+        // True signifies that the replacement can happen and false signifies that replacement shouldn't happen.
+        var labelVariableMap: [String: Bool] = [:]
```
mkr-plse, line 872: Do you think we can update the map to not use Strings? We currently use strings (see [here](https://github.com/uber/piranha/blob/master/swift/Sources/PiranhaKit/CleanupStaleFlags/StaleFlagCleaner.swift#L42) to identify variables which is hacky. We can't identify the equivalence among self.X, X, <classname>.X and have to add specific mappings. Ideally, we will have a variable to bool mapping and the matching will be on the variable as opposed to strings.  

I am fine if you want to handle it in a followup PR too as it touches all parts of the current implementation.

swift/Sources/PiranhaKit/CleanupStaleFlags/StaleFlagCleaner.swift
```diff
@@ -714,6 +751,8 @@ class XPFlagCleaner: SyntaxRewriter {
             }
 
             if let rhs = binding.initializer {
+                updateScope(withIdentifier: binding.pattern.description.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines),
```
mkr-plse, line 754: the comment on Strings will help get rid of passing description and instead can probably operate on ExprSyntax.

mkr-plse (approved):
Discussed offline and the changes will be taken up as part of a separate PR.

PR #134, opened 2021-05-13
If else ladder refactoring

chiragramani:
**Context**: BooleanLiteral expressions in the if-else conditions should be further refactored by updating the if-else tree, wherever applicable.

**This PR:**
1. This PR adds an IfElseStmtRewriter - that basically consumes the output of the default re-writer and takes this as an opportunity to dedicatedly operate on the if-else condition and also refactor the if-else ladder.
It does the following:
a.) removes the conditional block that has the false condition because that will never be executed and also does the related stitching of the next blocks.
b.) If there is a conditional block evaluating to true, then it removes all the following conditions as well as their respective bodies because they will never be executed along with necessary stitching.
2. As per the above specs, the sample input files(testfile.swift, treated.swift, control.swift) are updated with suitable test cases.

Having independent dedicated reducers will prove to be beneficial as Piranha evolves:
1. The default re-writer focuses on experimentation-related replacements etc.
2. Reducers applied post the above re-writing further refactors as per their utility. This way, new reducers can be added without affecting each other though sequencing in which they will be called will depend on their refactoring strategy. 
3. Also, small scoped independent units are always easier to maintain.

### DIFF
--- a/swift/tests/InputSampleFiles/testfile.swift
+++ b/swift/tests/InputSampleFiles/testfile.swift
@@ -600,4 +600,28 @@ class SwiftExamples {
             print("4")
         }
     }
+    
+    private func ifElseLadder_3() {
+        if x {
+            print("1")
+        } else if true {
+            print("2")
+        } else if cachedExperiments.isTreated(forExperiment: ExperimentNamesSwift.test_experiment) {
+            print("3")
+        } else {
+            print("4")
+        }
+    }
+    
+    private func ifElseLadder_4() {
+        if x {
+            print("1")
+        } else if false {
+            print("2")
+        } else if cachedExperiments.isTreated(forExperiment: ExperimentNamesSwift.test_experiment) {
+            print("3")
+        } else {
+            print("4")
+        }
+    }
 }
--- a/swift/tests/InputSampleFiles/testfile.swift
+++ b/swift/tests/InputSampleFiles/testfile.swift
@@ -624,4 +624,28 @@ class SwiftExamples {
             print("4")
         }
     }
+    
+    private func ifElseLadder_5() {
+        if x {
+           print("1")
+        } else if cachedExperiments.isTreated(ExperimentNamesSwift.test_experiment) || y {
+            print("2")
+        } else if cachedExperiments.isTreated(ExperimentNamesSwift.randomExperiment) || z {
+            print("3")
+        } else {
+            print("4")
+        }
+    }
+    
+    private func ifElseLadder_6() {
+        if x {
+           print("1")
+        } else if cachedExperiments.isTreated(ExperimentNamesSwift.test_experiment) && y {
+            print("2")
+        } else if cachedExperiments.isTreated(ExperimentNamesSwift.randomExperiment) || z {
+            print("3")
+        } else {
+            print("4")
+        }
+    }
 }
--- a/swift/tests/InputSampleFiles/testfile.swift
+++ b/swift/tests/InputSampleFiles/testfile.swift
@@ -648,4 +648,14 @@ class SwiftExamples {
             print("4")
         }
     }
+    
+    private func ifElseLadder_7() {
+        if x {
+           print("1")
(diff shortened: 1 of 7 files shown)


mkr-plse (changes requested):
This is a very nice change to the implementation. I can see how this will eventually be used to finesse deep cleaning. 

However, there is one concern that needs to be fixed. The reducer should only be applied on expressions which are a result of a FF cleanup. Otherwise, this can modify code that are unrelated to FFs. 

How about having an option to differentiate between cleaning up _any_ expression vs only modifying expressions that were a result of a previous refactoring? This will let users choose an option that is more appropriate for their codebase.

chiragramani:
> This is a very nice change to the implementation. I can see how this will eventually be used to finesse deep cleaning.
> 
> However, there is one concern that needs to be fixed. The reducer should only be applied on expressions which are a result of a FF cleanup. Otherwise, this can modify code that are unrelated to FFs.
> 
> How about having an option to differentiate between cleaning up _any_ expression vs only modifying expressions that were a result of a previous refactoring? This will let users choose an option that is more appropriate for their codebase.

Hey, @mkr-plse  thanks for pointing this out - valid point. I have updated the code to make sure that only Piranha-related expressions are considered by the reducer.

**Some context about the changes**

1. If the source file has a condition ("if true" or "if false"), then post the Lexical Analysis - the tokenKind of "true" is = .trueKeywork and .falseKeywork for "if false" by default. But for expressions that Piranha reduces, we can have a small change in the code that basically constructs a TokenKind.identifier saying "true/false". Why is this required? Well, the reducer needs a hint - what to reduce and what to allow - reducer uses this hint and only applies respective reductions to Piranha refactored expressions. 
2. **Is there any other hint that we can use?** I tried different expressions, their index, root-wise comparisons, hash values, parent relationships, source ranges, and respective locations - but seems like other hints are not at all reliable because they keep changing because of many refactoring passes happening in Piranha. 
3. **Is there any con with the above approach?** Since we are updating the raw code and are only concerned with updating the file contents, no cons at all here. But if we do plan to analyze the AST in the future (extreme node to node comparisons), then definitely a true keyword has a different semantic meaning compared to an Identifier (having value == "true" || "false"). But this is again not a con, because our reducer is eliminating these true/false nodes so the last SourceSyntax that is generated, will not be aware of this since reducers will take care of this.  
**Pros:** we limit refactoring to Piranha.

swift/tests/InputSampleFiles/testfile.swift
```diff
@@ -576,4 +576,52 @@ class SwiftExamples {
...
+        } else if true {
+            print("2")
+        } else if cachedExperiments.isTreated(forExperiment: ExperimentNamesSwift.test_experiment) {
+            print("3")
+        } else {
+            print("4")
+        }
+    }
+    
+    private func ifElseLadder_4() {
```
mkr-plse, line 616: Can you also add couple of ladder tests that 
1. test compound expressions reducing to a boolean as part of this set of tests? 
2. test ending with a else if instead of else. 

e.g.,
```
if x {
   ...
} else if cachedExperiments.isTreated(...stale) || y  {
   ... 
} else if cachedExperiments.isTreated(...nonstale) || z { 
   ... 
} else {
   ...
}

mkr-plse (commented):
Let us add a few more tests to this PR.

chiragramani:
> Let us add a few more tests to this PR.

Yeah, sounds good. Added more test-cases. Let me know if more are to be added. Thanks!

swift/tests/InputSampleFiles/testfile.swift
chiragramani: Yeah, added for both the cases
1. ifElseLadder_5 - having "or" expression.
2. ifElseLadder_6 - having "and" expression.
3. ifElseLadder_6 - if else terminating with "else-if"..

mkr-plse (approved):
Looks great. Thanks for taking care of this.

PR #135, opened 2021-05-19
[PiranhaSwift] Fix handling of trivia in enum declaration

mkr-plse:
The current version of PiranhaSwift resulted in refactorings where two flag declarations appear on the same line due to an unnecessary conditional check while handling enums. Deleted the associated code and added tests to cover this scenario.


lazaroclapp (approved):
Test case makes sense. Approved.

PR #136, opened 2021-05-26
[PiranhaSwift] Expression simplification during equality comparison

mkr-plse:
```
if isTreated(flag) == true {
   // abc
} 
``` 
is refactored as 
``` 
if true == true {
  // abc
}
```

This update fixes this issue and simplifies the code further.


lazaroclapp (approved):
LGTM.

With the note that:

```
if isTreated(flag) == true {
   // abc
} 
```

is already redundant code to begin with... is that being produced as the output of some other refactoring?

PR #151, opened 2021-09-24
[PiranhaJava] Remove configured test methods

### LINKED ISSUE #47: PiranhaJava // Remove meaningless "assert" and "expect" from JUnits
Piranha simplification will need manual review and fixing of testcases, but I think it's useful to delete lines like:
`expect(true).andReturn(true);`
`assertFalse(true);`
which Piranha leaves behind in test classes after simplifying toggle expressions.

tnterdan:
Fixes #47 

This development allows the user of Piranha to specify a list of methods
where if a dynamic flag evaluation is in a certain argument position, we
remove that method.

This is useful for cleaning up `mock()` wrapper methods or `assert()`
methods in test code that are useless after a feature flag is removed.

Limitations:
  - This does not remove unused imports for the test methods because we
    have to account for cases of the test methods that will not be
    removed.

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -115,7 +115,7 @@ when `IsTreated` is `true`, and will be deleted completely when `IsTreated` is `
 
 An optional top-level field is `testMethodProperties`.
 
-Within that, there is an array of JSON objects, having the required fields `methodName` and `argumentIndex`. The both behave the same as the fields with the same name in `methodProperties`.
+Within that, there is an array of JSON objects, having the required fields `methodName` and `argumentIndex`. They both behave the same as the fields with the same name in `methodProperties`.
 
 What this field does, is that if we find one of the `methodProperties` fields inside a method that matches one of the methods in `testMethodProperties`, we remove that method. This is useful for removing `mock()` wrappers or `assert()` calls that are no longer useful after a flag is cleaned up.
 
--- a/java/README.md
+++ b/java/README.md
@@ -65,6 +65,13 @@ The properties file has the following template:
       },
       ...
     ],
+  "testMethodProperties": [
+    {
+      "methodName": "when",
+      "argumentIndex": 0
+    },
+    ...
+  ],
   "enumProperties":
     [
       {
@@ -106,7 +113,13 @@ public void some_unit_test() { ... }
 
 when `IsTreated` is `true`, and will be deleted completely when `IsTreated` is `false`.
 
-An optional top-level field is `enumProperties`.
+An optional top-level field is `testMethodProperties`.
+
+Within that, there is an array of JSON objects, having the required fields `methodName` and `argumentIndex`. The both behave the same as the fields with the same name in `methodProperties`.
+
+What this field does, is that if we find one of the `methodProperties` fields inside a method that matches one of the methods in `testMethodProperties`, we remove that method. This is useful for removing `mock()` wrappers or `assert()` calls that are no longer useful after a flag is cleaned up.
+
+Another optional top-level field is `enumProperties`.
 Within that, there is an array of JSON objects, having the required fields `enumName` and `argumentIndex`.
 
 What this field does, is if you specify an enum class name, Piranha will remove enum constants that have a constructor with a string argument that matches your `FlagName` value, along with their usages.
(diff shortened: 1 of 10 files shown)


java/README.md
```diff
@@ -106,7 +113,13 @@ public void some_unit_test() { ... }
 
 when `IsTreated` is `true`, and will be deleted completely when `IsTreated` is `false`.
 
-An optional top-level field is `enumProperties`.
+An optional top-level field is `testMethodProperties`.
+
+Within that, there is an array of JSON objects, having the required fields `methodName` and `argumentIndex`. The both behave the same as the fields with the same name in `methodProperties`.
```
mkr-plse, line 118: The -> They

java/README.md
```diff
@@ -106,7 +113,13 @@ public void some_unit_test() { ... }
...
+
+What this field does, is that if we find one of the `methodProperties` fields inside a method that matches one of the methods in `testMethodProperties`, we remove that method. This is useful for removing `mock()` wrappers or `assert()` calls that are no longer useful after a flag is cleaned up.
```
mkr-plse, line 120: Should we refer this to as `testMethodWrappers`?

java/piranha/src/test/java/com/uber/piranha/TestCaseCleanUpTest.java
```diff
@@ -833,4 +833,196 @@ public void testCleanBySettersHeuristicIgnoreSettersForOtherFlags() throws IOExc
             "}")
         .doTest();
   }
+
```
mkr-plse, line 836: nice set of tests.

java/piranha/src/main/java/com/uber/piranha/config/PiranhaFlagMethodRecord.java
```diff
@@ -0,0 +1,141 @@
...
+   * @param methodPropertyEntry The decoded json entry (as a Map of property names to values)
+   * @param isArgumentIndexOptional Whether argumentIdx should be treated as optional
+   * @return A PiranhaFlagMethodRecord corresponding to the given map/json record.
+   * @throws PiranhaConfigurationException if there was any issue reading or parsing the
+   *     configuration file.
+   */
+  static PiranhaFlagMethodRecord parseFromJSONPropertyEntryMap(
+      Map<String, Object> methodPropertyEntry, boolean isArgumentIndexOptional)
+      throws PiranhaConfigurationException {
+    String methodName = getValueStringFromMap(methodPropertyEntry, METHOD_NAME_KEY);
```
mkr-plse, line 111: Can code from line 111 to 132 be hoisted into a utility method and called. The same method can be called from PiranhaTestMethodRecord too. The messaging in the `throw` and the check on `flagType` seem to be the only differences between the two methods.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -880,19 +881,61 @@ public Description matchBinary(BinaryTree tree, VisitorState state) {
...
+      if (argumentIdx.isPresent()) {
+        if (methodTree.getArguments().size() > argumentIdx.get()) {
+          ExpressionTree argTree = methodTree.getArguments().get(argumentIdx.get());
+          API api = getXPAPI(argTree, state);
+          if (api != API.UNKNOWN) {
+            return true;
+          }
+        }
+      } else {
+        for (ExpressionTree argTree : methodTree.getArguments()) {
```
mkr-plse, line 898: does this mean the flagTest API can occur at any index?

mkr-plse (commented):
Overall, looks good. A minor refactoring comment will be helpful.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
tnterdan: Yes, but only if the argument index is not defined.

java/README.md
tnterdan: Should we keep the pattern so they all still end in properties? `testMethodWrapperProperties`? Since it behaves so similarly to `methodProperties`.

lazaroclapp:
> Limitations:
> 
> * This does not remove unused imports for the test methods because we
>   have to account for cases of the test methods that will not be
>   removed.

Arguably better in a follow-up, but this sounds like the use case for [UsageCounter](https://github.com/uber/piranha/blob/master/java/piranha/src/main/java/com/uber/piranha/UsageCounter.java), which I believe we use in similar cases. Not 100% sure because I haven't yet reviewed the code (starting on it right now :) )

danieltrt:
Closing since we don't have legacy versions anymore

PR #159, opened 2022-01-06
clean up enum and comma in annotation

valulucchesi:
when there is more than one enum in the annotation and it's being separated by comma piranha is keeping an extra comma after the cleanup.

code to do the cleanup by STALE_FLAG as treated:
`@ToggleTesting(treated = {TestExperimentName.STALE_FLAG, TestExperimentName.OTHER_FLAG})`

current output:
`@ToggleTesting(treated = {, TestExperimentName.OTHER_FLAG})`

expected output
`@ToggleTesting(treated = {TestExperimentName.OTHER_FLAG})`

The enum to be removed could be at the beginning, in the middle or at the end. For the first two cases we should remove the comma that is present after the enum and for the third case we should remove the comma that is present before the enum.
For the scenario when there is only one enum we shouldn't do anything else that what the code is currently doing.

### DIFF
--- a/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
+++ b/java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
@@ -57,6 +57,7 @@ import com.sun.source.tree.UnaryTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
 import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.tree.JCTree;
 import com.uber.piranha.config.Config;
 import com.uber.piranha.config.PiranhaConfigurationException;
 import com.uber.piranha.config.PiranhaEnumRecord;
@@ -1375,7 +1376,7 @@ public class XPFlagCleaner extends BugChecker
       // Note that the AST elements referenced by deletableAnnotations and deletableIdentifiers
       // should not overlap, given the logic above, so deleting each independently is safe.
       for (ExpressionTree expr : deletableIdentifiers) {
-        fixBuilder = deleteExprWithComma(state, tree, expr, resolvedTestAnnotations, fixBuilder);
+        fixBuilder = deleteExprWithComma(state, expr, resolvedTestAnnotations, fixBuilder);
         decrementAllSymbolUsages(expr, state, fixBuilder);
       }
       for (AnnotationTree at : deletableAnnotations) {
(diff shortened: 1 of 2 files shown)


java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -1564,6 +1565,39 @@ private boolean endsWithReturn(StatementTree stmt) {
...
+        int index =
+            IntStream.range(0, resolvedTestAnnotation.getFlags().size())
+                .filter(i -> resolvedTestAnnotation.getFlags().get(i).getValue().equals(xpFlagName))
+                .findFirst()
+                .orElse(-1);
+        if (index != -1) {
+          String replacementString = state.getSourceForNode(tree);
+          if (index == resolvedTestAnnotation.getFlags().size() - 1) {
+            replacementString =
+                replacementString.replace(", " + state.getSourceForNode(expressionTree), "");
```
ketkarameya, line 1587: While I believe this will work in most cases, it still ties us down to a formatting style where arguments are separated by `, `. I believe we should make it more general, (e.g. arguments could also be separated by a new line character or arguments could be separated by just a `,`).
One options would be to do it the way error-prone folks delete type arguments in their [`UnnecessaryTypeArguments` bug check](https://sourcegraph.com/github.com/google/error-prone@69f45c3059b7b858dbe4c41afef6c20ae6ccaf15/-/blob/core/src/main/java/com/google/errorprone/bugpatterns/UnnecessaryTypeArgument.java?L83).

java/piranha/src/test/java/com/uber/piranha/EnumConstantTest.java
```diff
@@ -848,4 +848,236 @@ public void testRemoveEnumWithNonEnumConstantConstructor() {
...
+            " enum TestExperimentName {",
+            "  OTHER_FLAG,",
+            "  STALE_FLAG;",
+            " }",
+            " @Retention(RetentionPolicy.RUNTIME)",
+            " @Target({ElementType.METHOD})",
+            " @interface ToggleTesting {",
+            "  TestExperimentName[] treated();",
+            " }",
+            " @ToggleTesting(treated = {TestExperimentName.STALE_FLAG, TestExperimentName.OTHER_FLAG})",
```
ketkarameya, line 939: Another test could be added `@ToggleTesting(treated = {TestExperimentName.STALE_FLAG,TestExperimentName.OTHER_FLAG})` and 
```
@ToggleTesting(treated = {TestExperimentName.STALE_FLAG,
                                                TestExperimentName.OTHER_FLAG})
```

ketkarameya (commented):
Thanks a lot for your contribution. Please find my comments on the pull request.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -1568,36 +1569,63 @@ private boolean endsWithReturn(StatementTree stmt) {
...
+
+        fixBuilder.replace(startAbsolute + lower, startAbsolute + upper, "");
       }
     }
     return fixBuilder;
   }
 
+  private int getLower(CharSequence source, int lower, int index, int flagSize) {
+    while (lower >= 0
+        && source.charAt(lower) != '{'
```
ketkarameya, line 1560: In what scenarios would we encounter a `{` or a `=` in an Enum List?  Can you please add a test scenario for this?

ketkarameya (commented):
These changes look good to me. I have a small comment below.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
valulucchesi: for `=` i was thinking in `@ToggleTesting(treated = TestExperimentName.STALE_FLAG)` but you are right that it is not needed at this stage so i'll remove it
for `{` the scenario is tested in `testRemoveFirstEnumFromAnnotationRemovingCommaAndKeepingOtherEnum` where the enum is at the beginning of the list: `@ToggleTesting(treated = {TestExperimentName.STALE_FLAG, TestExperimentName.OTHER_FLAG})`

ketkarameya:
Thanks a lot for your contribution :) 
I am merging the PR.

ketkarameya (approved):
LGTM.

ketkarameya:
@valulucchesi I was wondering if this new implementation could also handle something like [this](https://github.com/uber/piranha/pull/128) ?

PR #160, opened 2022-01-06
Preparing the release and configurations for the next version.

ketkarameya:


lazaroclapp (approved):
This should have landed as two commits. Doing a rebase merge as an exception.

PR #162, opened 2022-01-11
Refactoring to use the Matcher API 

ketkarameya:
The goal is to improve the readability and maintainability of the code by adapting Error-prone's Matcher API.


java/piranha/src/main/java/com/uber/piranha/PiranhaUtils.java
```diff
@@ -44,4 +48,45 @@ public static boolean isPrefixPath(TreePath prefix, TreePath path) {
...
+    private final int position;
+    private final Matcher<ExpressionTree> argumentMatcher;
+
+    public NewClassArgument(int position, Matcher<ExpressionTree> argumentMatcher) {
+      this.position = position;
+      this.argumentMatcher = argumentMatcher;
+    }
+
+    @Override
+    public boolean matches(NewClassTree newClassTree, VisitorState state) {
```
mkr-plse, line 62: Can we add some comments to this code?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -1352,8 +1296,7 @@ public Description matchMethod(MethodTree tree, VisitorState state) {
         }
         matchedFlagsWorkingSet.clear();
       }
-    } else if (config.shouldCleanTestMethodsByContent()
-        && PiranhaUtils.isUnitTestMethod(tree, state)) {
+    } else if (config.shouldCleanTestMethodsByContent() && isUnitTestMethod(tree, state)) {
```
mkr-plse, line 1299: the `PiranhaUtils.` qualifier helps with readability?

ketkarameya: I think so too. These changes creep in because IntelliJ suggests them and I just tend to select it.

mkr-plse (approved):
Nice cleanup. Thanks!

java/piranha/src/main/java/com/uber/piranha/PiranhaUtils.java
```diff
@@ -44,4 +48,78 @@ public static boolean isPrefixPath(TreePath prefix, TreePath path) {
...
+      int index, Matcher<ExpressionTree> argMatcher) {
+    return new NewClassArgument(index, argMatcher);
+  }
+
+  /**
+   * Returns a matcher for the arguments passed to a new class creation.
+   *
+   * @param matchType - all, any, none
+   * @param argMatcher The matcher for the argument
+   * @return Returns true if any (or all or none) arguments match, otherwise false.
```
lazaroclapp, line 69: Super small nit: "if all (or any or none)", just to keep the same order as in the documentation of `matchType`

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -356,57 +368,70 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
     return API.UNKNOWN;
   }
 
+  private Matcher<MethodInvocationTree> methodRecordMatcher(
+      MethodRecord methodRecord, VisitorState state) {
+
+    return allOf(
```
lazaroclapp, line 374: Maybe worth a few documenting comments for each of these clauses?
```
return allOf(
       // Method name must match methodRecord.getMethodName()
       (methodRecord.isStatic() ? staticMethod().anyClass() : instanceMethod().anyClass())
            .named(methodRecord.getMethodName()),
        // Method must have flag at argument index specified by record (if any)
        methodRecord
            .getArgumentIdx()
            .map(idx -> argument(idx, isArgumentMatchesFlagName(state)))
            .orElseGet(Matchers::anything),
        // Method's receiver must match record's receiver type (if any)
        methodRecord
            .getReceiverType()
            .map(
                type ->
                    receiverOfInvocation((receiver, st) -> isSameType(type).matches(receiver, st)))
            .orElseGet(Matchers::anything),
        // Method's return must match record's return type (if any)
        methodRecord.getReturnType().map(Matchers::isSameType).orElseGet(Matchers::anything));
```

I'd also consider flipping the check for receiver type and that of the argument, unless there is a performance reason for the order? (seems more natural to check method name and receiver type before the arguments)

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -356,57 +368,70 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
...
-      // when returnType is specified, check if mst's return type matches it
-      // if it's not a match, skip to next method property map
-      Optional<String> optionalReturnType = methodRecord.getReturnType();
-      if (optionalReturnType.isPresent()) {
-        String mReturn = ASTHelpers.getReturnType(mst).toString();
-        if (!optionalReturnType.get().equals(mReturn)) {
-          continue;
-        }
+      if (methodRecordMatcher(methodRecord, state).matches(mit, state)) {
+        return methodRecord.getApiType();
```
lazaroclapp, line 421: ❤️

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -356,57 +368,70 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
...
+                type ->
+                    receiverOfInvocation((receiver, st) -> isSameType(type).matches(receiver, st)))
+            .orElseGet(Matchers::anything),
+        methodRecord.getReturnType().map(Matchers::isSameType).orElseGet(Matchers::anything));
+  }
+
+  private Matcher<ExpressionTree> enumConstructorArgsContainsFlagNameMatcher(
+      ImmutableCollection<PiranhaEnumRecord> enumRecords) {
+    Matcher<ExpressionTree> matchFlag =
+        (arg, stt) ->
```
lazaroclapp, line 393: Wondering if we should just call the second argument on these two cases `state` if it's a `VisitorState`, for clarity. Or is `st` idiomatic in example Error Prone matchers? (Either way, `stt` looks like a typo to me)

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -356,57 +368,70 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
...
+            arg.getKind() == Kind.STRING_LITERAL
+                && ((LiteralTree) arg).getValue().equals(xpFlagName);
+    return toType(
+        NewClassTree.class,
+        (nct, st) ->
+            enumRecords
+                .stream()
+                .anyMatch(
+                    enumRecord ->
```
lazaroclapp, line 402: This is a bit over-nested, could we extract this lambda into a method we can use a method reference of here?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -416,56 +441,14 @@ private boolean isArgumentMatchesFlagName(
...
-    Symbol.ClassSymbol enclosingClass = ASTHelpers.enclosingClass(sym);
-
-    return isEnumConstantMatchingFlagName(sym.getSimpleName().toString(), enclosingClass, state);
+  private Matcher<ExpressionTree> isMatchingEnumFieldValue(VisitorState state) {
+    Matcher<ExpressionTree> enumFieldValueMatcher =
+        symbolMatcher(
+            (sym, st) ->
+                isEnumConstantMatchingFlagName(
+                    sym.getSimpleName().toString(), ASTHelpers.enclosingClass(sym), state));
+    return anyOf(enumFieldValueMatcher, contains(ExpressionTree.class, enumFieldValueMatcher));
```
lazaroclapp, line 451: Wouldn't this over-match any expression that includes the flag? That might be too aggressive, depending on where this is called (compared with the four cases of the original code).

For example, this matcher would match on:

```
TestExperimentName.STALE_FLAG || TestExperimentName.OTHER_FLAG
```

Would it not?

How about:

```
TestExperimentName.OTHER_FLAG.hasFoo(TestExperimentName.STALE_FLAG.getFoo())
```

?

Not 100% where this is being called from and if it's safe because of that, but in general this is matching a lot more aggressively than: enum constant alone or method call on enum constant. At the very least the documentation comments need to be moves around and adjusted (a more general javadoc for this method, the 4 cases above where they are validated to hold)

lazaroclapp (commented):
Generally makes sense to me, particularly since all tests match. Mostly I wonder if the matching on enum constants is a bit too broad compared to the existing code and why that would be safe.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: If u see this code in your editor they are very small expressions. I don't kow why github makes them look so scary.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: I have addressed this. I have made the matcher as "narrow" as it was before the refactoring.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: done!

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: done!

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: As per our discussion online, you can observe that I have made the matchers "not lazy" too.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -416,56 +447,19 @@ private boolean isArgumentMatchesFlagName(
...
+    Matcher<ExpressionTree> enumFieldValueMatcher =
+        symbolMatcher(
+            (sym, st) ->
+                isEnumConstantMatchingFlagName(
+                    sym.getSimpleName().toString(), ASTHelpers.enclosingClass(sym), state));
+    // Handles - the other three scenarios
+    return anyOf(
+            enumFieldValueMatcher,
+            Matchers.methodInvocation(
+                anyOf(enumFieldValueMatcher, memberSelectExpression(enumFieldValueMatcher))))
```
ketkarameya, line 461: @lazaroclapp  I addressed your comments regarding the new check being too aggressive. Now it is equivalent to the switch case check from before.

java/piranha/src/main/java/com/uber/piranha/PiranhaUtils.java
lazaroclapp: Again, this one doesn't matter a whole lot, but I don't think it was "resolved" per se. My suggestion was to list the match types in the same order for `@param matchType` and for `@return`.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -368,31 +367,43 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
...
 
-    return allOf(
-        (methodRecord.isStatic() ? staticMethod().anyClass() : instanceMethod().anyClass())
-            .named(methodRecord.getMethodName()),
-        methodRecord
-            .getArgumentIdx()
-            .map(idx -> argument(idx, isArgumentMatchesFlagName(state)))
-            .orElseGet(Matchers::anything),
-        methodRecord
+    return // Method name must match methodRecord.getMethodName()
```
lazaroclapp, line 373: If we are returning a boolean anyways, can't we make this a fair bit clearer by using conditionals vs a 20+ lines boolean expression? E.g.

```
// Method name must match methodRecord.getMethodName()
Matcher<MethodInvocationTree> nameMatcher = 
           (methodRecord.isStatic() ? staticMethod().anyClass() : instanceMethod().anyClass())
            .named(methodRecord.getMethodName());
if (!nameMatcher.matches(mit, state)) {
   return false;
}
```

and so on?

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -441,14 +447,19 @@ private API getXPAPI(
...
-  private Matcher<ExpressionTree> isMatchingEnumFieldValue(VisitorState state) {
+  private boolean isMatchingEnumFieldValue(ExpressionTree tree, VisitorState state) {
+    // Handles - Unqualified import enum constant
     Matcher<ExpressionTree> enumFieldValueMatcher =
         symbolMatcher(
             (sym, st) ->
                 isEnumConstantMatchingFlagName(
                     sym.getSimpleName().toString(), ASTHelpers.enclosingClass(sym), state));
-    return anyOf(enumFieldValueMatcher, contains(ExpressionTree.class, enumFieldValueMatcher));
+    // Handles - the other three scenarios
```
lazaroclapp, line 457: I think the first matcher handles both unqualified and qualified enum constant. No? At any rate, the method invocation matcher will only handle 3 & 4.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: Somehow GitHub swallowed my follow up comment here, but one point I want to make is about debugability: it's way easier to step through the code or set a breakpoint if this entire method is not a single statement.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
lazaroclapp: You call here, I think :)

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: I agree. This will ease debugging through code too. I ll actually split it.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
ketkarameya: Yes, this is a mistake in my comment. Thanks for pointing it out.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -370,37 +370,39 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
...
+    if (!nameMatches) return false;
+
+    // Method's receiver must match record's receiver type (if any)
+    boolean receiverTypeMatches =
+        !methodRecord.getReceiverType().isPresent()
+            || receiverOfInvocation(
+                    (receiver, st) ->
+                        isSameType(methodRecord.getReceiverType().get()).matches(receiver, st))
+                .matches(mit, state);
+    if (!receiverTypeMatches) return false;
```
lazaroclapp, line 388: I believe our coding guidelines recommend all conditionals to have a block for their branches (see [this](http://errorprone.info/bugpattern/MissingBraces) check, which we use internally and on NullAway, and which might make sense for this project too). So, if we are following that same style for Piranha, this probably should be:

```
if (!receiverTypeMatches) {
   return false;
}
```

lazaroclapp (approved):
LGTM. (Modulo the super minor nit noted below)

PR #163, opened 2022-01-22
Initial code and test scenarios for different api style

ketkarameya:
This PR adds functionality for matching experiment APIs that require invoking a chain of method invocations. 
This functionality can be toggled by setting the `allow_method_chain` in the clean_up options to `true`.
In addition, it also tests the possibility of using Piranha for experiment APIs (i) with non pre-determined names and (ii) non pre-determined membership to the treatment/control group. 
The PR contains many tests for a mock API that requires invoking a chain of method invocations.  The tests in this PR transform the input properties by replacing the tokens $notIsTreated$ and $flagName$ with values based on arguments passed to Piranha.


java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,198 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+        "{\n"
+            + "      \"methodName\": \"$flagName$.getValue\",\n"
+            + "      \"flagType\": \"treated\",\n"
+            + "      \"returnType\": \"boolean\"\n"
+            + "    },";
+    String annotationPropertiesToAdd =
+        "{\n"
+            + "      \"name\" : \"BoolParam\",\n"
+            + "      \"flag\" : \"$key\",\n"
+            + "      \"treated\" : \"$notIsTreated$\"\n"
```
ketkarameya, line 1054: This will ensure that the corresponding method is always deleted.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,198 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
   }
+
+  public void transformAndCreateNewPropertyFile(
+      String srcProp, String trgtProp, String stale_flag, boolean isTreated) throws IOException {
+
+    String[] temp = stale_flag.split("_");
+    String methodPropertiesToAdd =
+        "{\n"
+            + "      \"methodName\": \"$flagName$.getValue\",\n"
+            + "      \"flagType\": \"treated\",\n"
```
ketkarameya, line 1047: Always treat this series of invocation as treated. So that it will evaluate to `true` when `isTreated = true` else `false`

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -370,16 +368,6 @@ private API getXPAPI(ExpressionTree et, VisitorState state) {
 
   private boolean methodRecordMatcher(
       MethodRecord methodRecord, VisitorState state, MethodInvocationTree mit) {
-
-    // Method name must match methodRecord.getMethodName()
-    boolean nameMatches =
```
mkr-plse, line 375: Can we document this method on the semantics of when it returns true or false.

java/piranha/src/main/java/com/uber/piranha/config/Config.java
```diff
@@ -126,14 +128,26 @@ private Config(
...
+      MethodInvocationTree mit) {
+    MemberSelectTree mst = (MemberSelectTree) mit.getMethodSelect();
+    String methodName = mst.getIdentifier().toString();
+    if (configMethodProperties.containsKey(methodName))
+      return configMethodProperties.get(methodName);
+    if (mst.getExpression() instanceof MethodInvocationTree) {
+      MethodInvocationTree expression = (MethodInvocationTree) mst.getExpression();
+      if (expression.getMethodSelect() instanceof MemberSelectTree) {
+        MemberSelectTree chained_mst = (MemberSelectTree) expression.getMethodSelect();
+        String chained_methodName = chained_mst.getIdentifier() + "." + methodName;
```
mkr-plse, line 145: `chainedMethodName`, `chainedMST`, ...

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,330 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+        .addInputLines(
+            "MethodChainTest.java",
+            "package com.uber.piranha;",
+            "public class MethodChainTest{",
+            " public static void foobar(Parameter cp){",
+            " SomeParam sp = SomeParam.create(cp);",
+            "  if(sp.staleFlag().getValue()){",
+            "    System.out.println();",
+            "   }",
+            "  if(!sp.staleFlag().getValue()){",
```
mkr-plse, line 1148: can we also test with:
1. increased chain length
2. reversing the method invocations with what is given in properties.json
3. having a method that only matches partially the method invocations in the properties.json

a bit more emphasis on testing this out. also, it may be easier to test this with this [style](https://github.com/uber/piranha/blob/master/java/piranha/src/test/resources/com/uber/piranha/XPFlagCleanerPositiveCases.java) of testing where input and output files are compared. The number of cases for the input line and output lines is one too many and can be hard to follow.

java/piranha/src/main/java/com/uber/piranha/config/Config.java
```diff
@@ -124,16 +126,33 @@ private Config(
   }
 
   /**
-   * Return all configuration method records matching a given method name.
+   * Return all configuration method records matching a given method name. A method record name
+   * field can be used to express a single method invocation or a chain of two method invocations. A
+   * chain of method invocations can be expressed by joining the name of the chained methods (from
+   * left to right) with a `.`. For instance, the name `isToggle` will match a invocation like -
+   * :[1].isToggle(:[2])), and the name `stale_flag.getValue` will match a invocation like -
```
mkr-plse, line 133: can we make this discussion more generic? This may be reachable to whoever follows the `hole` terminology.

java/piranha/src/main/java/com/uber/piranha/config/Config.java
```diff
@@ -124,16 +126,33 @@ private Config(
   }
 
   /**
-   * Return all configuration method records matching a given method name.
+   * Return all configuration method records matching a given method name. A method record name
+   * field can be used to express a single method invocation or a chain of two method invocations. A
```
mkr-plse, line 130: is the chain restricted to two?

java/piranha/src/main/java/com/uber/piranha/config/Config.java
```diff
@@ -124,16 +126,33 @@ private Config(
...
-    return configMethodProperties.containsKey(methodName)
-        ? configMethodProperties.get(methodName)
-        : ImmutableSet.of();
+  public ImmutableCollection<PiranhaMethodRecord> getMethodRecordsForName(
+      MethodInvocationTree mit) {
+    MemberSelectTree mst = (MemberSelectTree) mit.getMethodSelect();
+    String methodName = mst.getIdentifier().toString();
+    if (configMethodProperties.containsKey(methodName))
+      return configMethodProperties.get(methodName);
+    if (mst.getExpression() instanceof MethodInvocationTree) {
```
mkr-plse, line 146: since this scenario will be infrequent for most users of Piranha, can we add a quick filter (say an option that is false by default) to handle "chaining of methods"?

java/piranha/src/main/java/com/uber/piranha/config/Config.java
```diff
@@ -124,16 +126,33 @@ private Config(
...
+      MethodInvocationTree expression = (MethodInvocationTree) mst.getExpression();
```
mkr-plse, line 147: the three lines here can probably go into a utility function as there is some repetition with the first three lines in the method here.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,112 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
             "}")
         .doTest();
   }
+
+  public void transformAndCreateNewPropertyFile(
+      String srcProp, String trgtProp, String stale_flag, boolean isTreated) throws IOException {
+
+    String[] temp = stale_flag.split("_");
+    String methodPropertiesToAdd =
```
mkr-plse, line 1044: Can all the strings here go into a custom properties file in the test resources directory?

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,112 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+    Path trgt = Paths.get(trgtProp);
+    Files.deleteIfExists(trgt);
+    Files.write(trgt, newContent.getBytes(UTF_8), StandardOpenOption.CREATE);
+  }
+
+  @Test
+  public void testMethodChainTreated() throws IOException {
+    String stale_flag = "stale_flag";
+    String isTreated = "true";
+    String srcProp = "config/properties.json";
```
mkr-plse, line 1100: this can read from a different properties.json with the $....$ parameters in there.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,112 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+    b.putFlag("Piranha:ArgumentIndexOptional", "true");
+    b.putFlag("Piranha:Config", trgtProp);
+
+    BugCheckerRefactoringTestHelper bcr =
+        BugCheckerRefactoringTestHelper.newInstance(new XPFlagCleaner(b.build()), getClass());
+
+    bcr = bcr.setArgs("-d", temporaryFolder.getRoot().getAbsolutePath());
+    bcr = bcr.addInput("Parameter.java").expectUnchanged();
+    bcr = bcr.addInput("BoolParameter.java").expectUnchanged();
+    bcr.addInput("XPMethodChainCases.java").addOutput("XPMethodChainCasesTreatment.java").doTest();
```
mkr-plse, line 1115: thanks a lot! makes it easier to review and understand the test.

java/piranha/src/test/resources/com/uber/piranha/XPMethodChainCases.java
```diff
@@ -0,0 +1,94 @@
...
+
+    static SomeParam create(Parameter cp) {
+      return null;
+    }
+  }
+
+  public static void foobar(Parameter cp) {
+    SomeParam sp = SomeParam.create(cp);
+    // Matches API
+    if (sp.staleFlag().getValue()) {
```
mkr-plse, line 21: nice. let us add one more condition on `otherFlag`. 

```
if (sp.otherFlag().getValue()) { .... }
```

Another case is to add a boolean expression 

```
if (sp.otherFlag().getValue() && sp.staleFlag().getValue()) { .... } 
```
Just to make sure that partial evaluation continues to hold.

java/piranha/src/test/resources/com/uber/piranha/XPMethodChainCases.java
```diff
@@ -0,0 +1,94 @@
...
+      System.out.println();
+    }
+
+    // Does not match annotation
+    @PVal(ns = "", key = "other_flag", val = "false")
+    public void testSomethingOther() {
+      System.out.println();
+    }
+  }
+  // Mock APIs to create corner case scenarios
```
mkr-plse, line 70: Can these mocks be moved into one of the other files you have created or this needs to be here? 
Since the related code doesn't change, moving it out can reduce the verbosity here.

java/piranha/src/test/resources/com/uber/piranha/XPMethodChainCases.java
```diff
@@ -0,0 +1,94 @@
+package com.uber.piranha;
+
+class XPMethodChainCases {
+
+  interface SomeParam {
+    @BoolParam(key = "other_flag")
+    BoolParameter otherFlag();
+
+    // Annotation and the abstract method should be removed
+    @BoolParam(key = "stale_flag")
```
mkr-plse, line 10: Curious on whether the interface implementation will be removed if `stale_flag` was the only parameter here. It can be a followup and need not be handled as part of this change.

mkr-plse (commented):
overall, looks good. left a few more comments.

java/piranha/src/main/java/com/uber/piranha/config/Config.java
ketkarameya: Yea. We can go `n` too, but it will have performance implications. Maybe we can accept method chain length as an input (default = 1) instead of adding a flag for "chaining of methods" (As u mentioned in your below comment) ?

java/piranha/src/test/resources/com/uber/piranha/XPMethodChainCases.java
ketkarameya: Will make a separate PR for this.

java/piranha/src/test/resources/com/uber/piranha/XPMethodChainCases.java
ketkarameya: ✅

java/piranha/src/main/java/com/uber/piranha/config/Config.java
ketkarameya: I decided to keep it to `2` for now, until we need more.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,81 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
             "}")
         .doTest();
   }
+
+  public void transformAndCreateNewPropertyFile(
+      String srcProp, String trgtProp, String stale_flag, boolean isTreated) throws IOException {
```
mkr-plse, line 1041: `staleFlag` instead of `stale_flag`.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,81 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+    String flagNameCamelCase =
+        temp[0]
+            + Arrays.stream(temp, 1, temp.length)
+                .map(str -> str.substring(0, 1).toUpperCase() + str.substring(1))
+                .collect(joining());
+    String newContent =
+        Files.readAllLines(Paths.get(srcProp))
+            .stream()
+            .map(x -> x.replace("$flagName$", flagNameCamelCase))
+            .map(x -> x.replace("$notIsTreated$", String.valueOf(!isTreated)))
```
mkr-plse, line 1053: any reason for representing this as a negation instead of `isTreated`?

java/piranha/src/main/java/com/uber/piranha/config/Config.java
```diff
@@ -124,17 +130,49 @@ private Config(
...
+   *
+   * @param mit Method invocation tree
+   * @return a string representing the name of the method invocation
+   */
+  public String getMethodName(MethodInvocationTree mit) {
+    if (mit.getMethodSelect() instanceof MemberSelectTree) {
+      MemberSelectTree mst = (MemberSelectTree) mit.getMethodSelect();
+      return mst.getIdentifier().toString();
+    }
+    // scenario when there is no receiver, like foo("bar")
```
mkr-plse, line 172: what will happen in the invocation of this at line 152? Please consider adding a related test for this.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,81 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+            .collect(joining("\n"));
+    Path trgt = Paths.get(trgtProp);
+    Files.deleteIfExists(trgt);
+    Files.write(trgt, newContent.getBytes(UTF_8), StandardOpenOption.CREATE);
+  }
+
+  @Test
+  public void testMethodChainTreated() throws IOException {
```
mkr-plse, line 1061: Can we add one more test with the same properties.json without referring `allow_method_chain` but having `.` in the methodName references in the json. Then, ensure the tests will result in no refactoring. This can be in a followup too.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
ketkarameya: Yes. This is how we can ensure that the method associated to the @BoolParm annotation is always deleted.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
ketkarameya: This is something which is a bit difficult to address because we have an independent visitor in XPFlagCleaner that deletes the abstract method from the interface based on the annotation match, irrespective of whether the allow chain option is set to true or false. The main concern here is that, the resulting code will contain a compilation error (abstract method is deleted, while invocations are not) which is hard to produce in the way we test through EP's APIs.

java/piranha/src/main/java/com/uber/piranha/config/Config.java
ketkarameya: Good catch. :)

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
mkr-plse: Ok. Maybe it is not completely unchanged and the definitions get deleted. The expected output can just have those deletions without the chain based deletions. There is an option for `allowBreakingChanges` with the refactoring that should handle any compilation errors that gets introduced due to this.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,228 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+    ErrorProneFlags.Builder b = ErrorProneFlags.builder();
+    b.putFlag("Piranha:FlagName", staleFlag);
+    b.putFlag("Piranha:IsTreated", isTreated);
+    b.putFlag("Piranha:ArgumentIndexOptional", "true");
+    b.putFlag("Piranha:Config", trgtProp);
+    BugCheckerRefactoringTestHelper bcr =
+        BugCheckerRefactoringTestHelper.newInstance(new XPFlagCleaner(b.build()), getClass());
+    bcr = bcr.setArgs("-d", temporaryFolder.getRoot().getAbsolutePath());
+    bcr = addMockAPIToNameSpace(bcr);
+    bcr.addInputLines(
```
mkr-plse, line 1131: can we not just use the same input file as before and use a different output file for testing?

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -1028,4 +1036,228 @@ public void testRemoveSpecificAPIpatterns() throws IOException {
...
+            "  System.out.println(\"done!\");",
+            " }",
+            "}")
+        .allowBreakingChanges()
+        .doTest();
+  }
+
+  // Checks basic method chain scenario - delete abstract method and chain of invocation
+  @Test
+  public void testMethodTestAllowsChainFlag() throws IOException {
```
mkr-plse, line 1191: isn't this already covered by the earlier tests?

mkr-plse (approved):
lgtm. if the comment related to the tests can be handled (even as a followup), that will be helpful.

java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
ketkarameya: Yup.

mkr-plse (approved):
Thanks for handling this.

PR #156, opened 2021-12-02
[PiranhaJava] Delete statements using EasyMock and JUnit API

ketkarameya:
Added new matchers for the EasyMock and Junit API

PR #167, opened 2022-02-03
Adding feature for removing unnecessary lambda expression block .

### LINKED ISSUE #166: Remove unnecessary block in lambda expressions after refactoring
Piranha transforms: 
```
someConsumer.accept((x) -> { 
                                        if(exp.isTreated(STALE_FLAG)){
                                           doSomething(x);
                                        }
                                        else{
                                           domeSomethingElse(x);
                                        }
                                       );
```

to 

```
someConsumer.accept((x) -> { 
                                           doSomething(x);
                                        }
                                       );
```
Rather it should:

```
someConsumer.accept((x) -> doSomething(x););
```

ketkarameya:
This PR adds a deep clean feature that removes unnecessary block statements in lambda expression bodies after the removal of code related to XP flags. Issue #166.


java/piranha/src/test/java/com/uber/piranha/CorePiranhaTest.java
```diff
@@ -906,137 +992,6 @@ public void testRemoveSpecificAPIpatternsJUnit() throws IOException {
         .doTest();
   }
 
-  @Test
-  public void testRemoveSpecificAPIpatterns() throws IOException {
```
ketkarameya, line 910: In my previous [PR](https://github.com/uber/piranha/pull/156/commits/1bd399497881e45ad9329c99f856cff948bef02e), I had split this test case into smaller test case. But I had not deleted this test case.

java/piranha/src/main/java/com/uber/piranha/XPFlagCleaner.java
```diff
@@ -1038,6 +1040,36 @@ public Description matchVariable(VariableTree tree, VisitorState state) {
...
+          ReturnTree stmt = (ReturnTree) singleStmt;
+          expr = stmt.getExpression();
+        }
+        if (singleStmt instanceof ExpressionStatementTree) {
+          ExpressionStatementTree expStmt = (ExpressionStatementTree) singleStmt;
+          expr = expStmt.getExpression();
+        }
+        if (expr != null) {
+          return buildDescription(tree)
+              .addFix(SuggestedFix.replace(bt, visitorState.getSourceForNode(expr)))
```
lazaroclapp, line 1065: Might be safer to build a new lambda expression with `expr` as the body and replace `tree` (but again, this is moot, since this won't do the rewrite you want).

lazaroclapp (changes requested):
Unfortunately, this is not easy to do, and this PR is incorrect at the moment (see the refactoring you are testing for in the test case, it's not actually doing anything new). Comments below as to why this approach won't directly work.

p.s. separately, this one is your call now rather than mine, but prefixing Piranha Java PRs with `[PiranhaJava]` will make them much easier to track in the changelog, particularly when - unlike now - other versions of the tool (for other languages) are seeing more active development.

PR #169, opened 2022-02-07
Adding tests for the paths not covered (According to coverall report)

ketkarameya:
The goal of this PR is to improve the code coverage of Piranha. Specifically, it contains new tests for : 
(1) unnecessary method invocation feature for instance method 
(2) Incorrect config file exception 
(3) Disable Unless configured option 
(4) Enum properties when argument index optional is true


mkr-plse (approved):
Great. Thanks for adding the tests to help improve coverage.

PR #170, opened 2022-02-07
Removing the meta variable hack from the tests

ketkarameya:
Based on the offline discussion, this PR removes the meta variable hack from the test cases.


mkr-plse (approved):
LGTM.

Issue #183, opened 2022-06-02
Unable to Refactor when flag is passed as argument to function in another file


Srihari123456:
The refactoring works if the flag is used inside a function. But when I call a function in another file and pass the value of a stale flag as an argument, the refactoring doesn't take place inside the called function.

file1.js
```
function f1(a){
   if(a){
     console.log("in if")
  } else {
     console.log("in else")
 }
}

module.exports = {f1}
```

file2.js
```
const file1 = require('./file1')
let a = isFlagTreated('featureFlag')
file1.f1(a)
```

**Expected Output:**

file1.js
```
function f1(true){
     console.log("in if")
}

module.exports = {f1}
```

file2.js
```
const file1 = require('./file1')
file1.f1(true)
```

**Obtained Output**

file1.js
```
function f1(a){
   if(a){
     console.log("in if")
  } else {
     console.log("in else")
 }
}

module.exports = {f1}
```

file2.js
```
const file1 = require('./file1')
file1.f1(true)
```

It would be great if a workaround is suggested.

Srihari123456:
Hey @ketkarameya 
Within the same file the behaviour is as follows:

**Case 1:**
Before Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

var a = isFlagTreated('featureFlag')
f1(a)
```

After Refactoring:
```
function f1(true) {
    console.log("in if")
}

f1(true)
```

**Case 2:**
Before Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

var b = isFlagTreated('featureFlag')
f1(b)
```

After Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

f1(true)
```

**Case 3:**
Before Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

f1(isFlagTreated('featureFlag'))
```

After Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

f1(true)
```

Based on these observations, I'm unsure if the ```inline parameter when possible``` functionality is present.
Its the variable which undergoes the substitution.

ketkarameya:
[...]

PR #189, opened 2022-06-07
Fixed bug - passing parameter values

Srihari123456:
Addressed Issue #183

**Case 1:**
Before Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

var a = isFlagTreated('featureFlag')
f1(a)
```

After Refactoring (Previously):
```
function f1(true) {
    console.log("in if")
}

f1(true)
```
After Refactoring (Now):
```
function f1(true) {
    console.log("in if")
}

f1(true)
```

**Case 2:**
Before Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

var b = isFlagTreated('featureFlag')
f1(b)
```

After Refactoring (Previously):
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

f1(true)
```
After Refactoring (Now):
```
function f1(true) {
    console.log("in if")
}

f1(true)
```

**Case 3:**
Before Refactoring:
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

f1(isFlagTreated('featureFlag'))
```

After Refactoring (Previously):
```
function f1(a) {
    if (a) {
        console.log("in if")
    } else {
        console.log("in else")
    }
}

f1(true)
```
After Refactoring (Now):
```
function f1(true) {
    console.log("in if")
}

f1(true)
```

### DIFF
--- a/javascript/src/refactor.js
+++ b/javascript/src/refactor.js
@@ -136,7 +136,6 @@ class RefactorEngine {
         if (func.body === null || func.body.body == null) {
             return false;
         }
-
         const returnIndex = func.body.body.length - 1;
 
         if (returnIndex < 0)
@@ -144,7 +143,6 @@ class RefactorEngine {
             return false;
         else {
             var returnNode = func.body.body[returnIndex];
-
             if (
                 returnNode.type === 'ReturnStatement' &&
                 returnNode.argument !== null &&
@@ -322,24 +320,23 @@ class RefactorEngine {
             enter: function (node) {
                 if (node.type === 'CallExpression') {
                     if (methodHashMap.has(node.callee.name)) {
-                        const argumentIndex = methodHashMap.get(node.callee.name).argumentIndex;
-                        const nodeArgument = node.arguments[argumentIndex];
-
+                        const argumentIndex = methodHashMap.get(node.callee.name).argumentIndex; // 0
+                        const nodeArgument = node.arguments[argumentIndex]; // node.arguments[0] = 'featureFlag' || featureFlag
                         let nodeArgumentIsFlag = false;
                         switch (nodeArgument.type) {
-                            case 'Identifier':
-                                nodeArgumentIsFlag = nodeArgument.name === engine.flagname;
+                            case 'Identifier': // node.arguments[0] =  featureFlag
+                                nodeArgumentIsFlag = nodeArgument.name === engine.flagname; // checks if the flag under consideration is the required flag
                                 break;
-                            case 'Literal':
+                            case 'Literal': // node.arguments[0] = 'featureFlag'
                                 nodeArgumentIsFlag = nodeArgument.value === engine.flagname;
                                 break;
                         }
                         if (nodeArgumentIsFlag) {
-                            const flagType = methodHashMap.get(node.callee.name).flagType;
+                            const flagType = methodHashMap.get(node.callee.name).flagType; // control || treated
                             engine.changed = true;
 
                             if (
-                                (flagType === 'treated' && engine.behaviour) ||
+                                (flagType === 'treated' && engine.behaviour) || // engine.behaviour = true => treated
                                 (flagType === 'control' && !engine.behaviour)
                             ) {
                                 return engine.trueLiteral();
@@ -352,6 +349,8 @@ class RefactorEngine {
             },
 
             fallback: 'iteration', // Ignore nodes in the AST that estraverse does not recognize
+            // By passing visitor.fallback option, we can control the behavior when encountering unknown nodes.
+            // Iterating the child **nodes** of unknown nodes.
         });
     }
 
@@ -361,7 +360,6 @@ class RefactorEngine {
     //  NOT true -> false, NOT false -> true
     evalBoolExpressions() {
         var engine = this;
-
         estraverse.replace(this.ast, {
             leave: function (node) {
                 if (node.type === 'LogicalExpression') {
@@ -409,10 +407,10 @@ class RefactorEngine {
                     engine.isPiranhaLiteral(node.test)
                 ) {
                     if (node.test.value === true) {
+                        // If statement is true
                         // node.consequent is always non-null so no check required
                         engine.changed = true;
                         engine.moveCommentsToConsequent(node);
-
                         return node.consequent;
                     } else if (node.test.value === false) {
                         if (node.alternate == null) {
@@ -448,9 +446,9 @@ class RefactorEngine {
     // var foo = cond -> _
     // where cond evaluates to a Boolean literal in a previous step
     getRedundantVarnames() {
+        // Makes the assignment between literal and value
         var assignments = {};
         var engine = this;
-
         // Get a list of variable names which are assigned to a boolean literal
         estraverse.traverse(this.ast, {
             enter: function (node) {
@@ -462,7 +460,7 @@ class RefactorEngine {
                             engine.isPiranhaLiteral(declaration.init) &&
                             typeof declaration.init.value === 'boolean'
                         ) {
-                            assignments[declaration.id.name] = declaration.init.value;
+                            assignments[declaration.id.name] = declaration.init.value; // Assigning a -> true
                         }
                     }
                 } else if (node.type === 'AssignmentExpression') {
@@ -478,11 +476,39 @@ class RefactorEngine {
         return assignments;
     }
 
+    adjustFunctionParams(parameterList) {
+        var engine = this;
+        var assignments = {};
+        estraverse.replace(this.ast, {
+            enter: function (node, parent) {
+                if (node.type === 'Identifier' && parent.type === 'FunctionDeclaration') {
+                    for (let i = 0; i < parent.params.length; i++) {
+                        if (node.name === parent.params[i].name) {
+                            for (let j = 0; j < parameterList.length; j++) {
+                                if (
+                                    parameterList[j].functionName === parent.id.name &&
+                                    parameterList[j].parameterIdx === i
+                                ) {
+                                    assignments[parent.params[i].name] = parameterList[j].value;
+                                    engine.pruneVarReferencesinFunc(parent, assignments);
+                                    engine.changed = true;
+                                    if (parameterList[j].value === true) return engine.trueLiteral();
+                                    return engine.falseLiteral();
+                                }
+                            }
+                        }
+                    }
+                }
+            },
+
+            fallback: 'iteration',
+        });
+    }
+
     // Remove all variable declarations corresponding variables in `assignments`
     // Replace all references of variables in `assignments` with the corresponding literal
-    pruneVarReferences(assignments) {
+    pruneVarReferences(assignments, parameterList) {
         var engine = this;
-
         // Remove redundant variables by deleting declarations and replacing variable references
         estraverse.replace(this.ast, {
             enter: function (node, parent) {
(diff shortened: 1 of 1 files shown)


ketkarameya:
Hey Thanks for the PR. I am looking into it. I am very unfamiliar with this code, so I may take some time. :|

Srihari123456:
Sure @ketkarameya !

ketkarameya:
Hey @Srihari123456 , 

I was looking at the goal of the PR. I have one concern here - isn't the literal `true`  redundant? Couldn't that be eliminated too?

For **Case 1** - should'nt the refactoring be 
```
function f1() {
    console.log("in if")
}

f1()
```
Note I have removed the literal `true` from argument. Unrelated,  I am a Javascript noob, but I was surprised to find that JS allows one to pass `true` as a parameter for `f1` . 
Similarly for other cases.

javascript/src/refactor.js
```diff
@@ -322,24 +320,23 @@ class RefactorEngine {
...
+                                nodeArgumentIsFlag = nodeArgument.name === engine.flagname; // checks if the flag under consideration is the required flag
                                 break;
-                            case 'Literal':
+                            case 'Literal': // node.arguments[0] = 'featureFlag'
                                 nodeArgumentIsFlag = nodeArgument.value === engine.flagname;
                                 break;
                         }
                         if (nodeArgumentIsFlag) {
-                            const flagType = methodHashMap.get(node.callee.name).flagType;
+                            const flagType = methodHashMap.get(node.callee.name).flagType; // control || treated
```
ketkarameya, line 335: Dangling comments?

javascript/src/refactor.js
```diff
@@ -322,24 +320,23 @@ class RefactorEngine {
...
                             engine.changed = true;
 
                             if (
-                                (flagType === 'treated' && engine.behaviour) ||
+                                (flagType === 'treated' && engine.behaviour) || // engine.behaviour = true => treated
```
ketkarameya, line 339: Dangling comments?

javascript/src/refactor.js
```diff
@@ -462,7 +460,7 @@ class RefactorEngine {
                             engine.isPiranhaLiteral(declaration.init) &&
                             typeof declaration.init.value === 'boolean'
                         ) {
-                            assignments[declaration.id.name] = declaration.init.value;
+                            assignments[declaration.id.name] = declaration.init.value; // Assigning a -> true
```
ketkarameya, line 463: Dangling comments?

javascript/src/refactor.js
```diff
@@ -498,6 +524,64 @@ class RefactorEngine {
                         return this.remove();
                     }
                 } else if (node.type === 'Identifier') {
+                    // this is the part where the function arguments are substituted values if they are a part of the assignments
+                    if (node.name in assignments) {
+                        if (assignments[node.name] === true) {
+                            engine.changed = true;
+                            return engine.trueLiteral();
+                        } else if (assignments[node.name] === false) {
+                            engine.changed = true;
```
ketkarameya, line 533: should `engine.changed` be false here? if not, can we pull out the two assignments outside the `if statement`

javascript/src/refactor.js
```diff
@@ -498,6 +524,64 @@ class RefactorEngine {
                         return this.remove();
                     }
                 } else if (node.type === 'Identifier') {
+                    // this is the part where the function arguments are substituted values if they are a part of the assignments
+                    if (node.name in assignments) {
+                        if (assignments[node.name] === true) {
```
ketkarameya, line 529: Simplify `if-else-if` to `if-else` or `ternary`.
`return assignments[node.name] === true ?  engine.trueLiteral() : engine.falseLiteral()` ?

javascript/src/refactor.js
```diff
@@ -478,11 +476,39 @@ class RefactorEngine {
         return assignments;
     }
 
+    adjustFunctionParams(parameterList) {
```
ketkarameya, line 479: Can we have some comments here - what is `adjustFunctionParams` doing?

javascript/src/refactor.js
```diff
@@ -498,6 +524,64 @@ class RefactorEngine {
...
+                    }
+                }
+            },
+
+            fallback: 'iteration',
+        });
+        return parameterList;
+    }
+
+    pruneVarReferencesinFunc(nn, assignments) {
```
ketkarameya, line 568: Can we have some comments on what this function is doing?

javascript/src/refactor.js
```diff
@@ -478,11 +476,39 @@ class RefactorEngine {
...
+        var engine = this;
+        var assignments = {};
+        estraverse.replace(this.ast, {
+            enter: function (node, parent) {
+                if (node.type === 'Identifier' && parent.type === 'FunctionDeclaration') {
+                    for (let i = 0; i < parent.params.length; i++) {
+                        if (node.name === parent.params[i].name) {
+                            for (let j = 0; j < parameterList.length; j++) {
+                                if (
+                                    parameterList[j].functionName === parent.id.name &&
```
ketkarameya, line 489: Can we move this check outside of this for loop ? From what i understand, you are checking if the `parent` function declaration name is same as the function name of `parameter_list` right?  Maybe this could be like the first line inside `enter` ?

ketkarameya: Also do we also need to check that `parameterList.length == parent.params.length` ?

javascript/src/refactor.js
```diff
@@ -536,7 +620,7 @@ class RefactorEngine {
         estraverse.traverse(this.ast, {
             enter: function (node, parent) {
                 if (node.type === 'FunctionDeclaration') {
-                    current = node.id.name;
+                    current = node.id.name; // name of the function
```
ketkarameya, line 623: dangling comment?

javascript/src/refactor.js
```diff
@@ -679,15 +762,16 @@ class RefactorEngine {
...
+            this.pruneVarReferences(redundantVarnames, parameterList);
+            this.adjustFunctionParams(parameterList);
             var functionsWithSingleReturn = this.getFunctionsWithSingleReturn();
             var redundantFunctions = this.getRedundantFunctions(functionsWithSingleReturn);
             this.pruneFuncReferences(redundantFunctions);
 
             iterations++;
         }
-
+        console.log(this.changed);
```
ketkarameya, line 774: stale code? could be removed!

ketkarameya (commented):
I have done my first pass over the PR.

javascript/src/refactor.js
Srihari123456: @ketkarameya the value of assignments[node.name] can be true or false or undefined. Thats why I didnt code it in if-else. Anyways I'll incorporate this suggestion and refine the code

javascript/src/refactor.js
Srihari123456: ```engine.changed``` must be true here. I'll refine the code such that it isnt redundant

javascript/src/refactor.js
Srihari123456: I'll remove this in the updated commit

javascript/src/refactor.js
Srihari123456: I'll remove this in the updated commit

javascript/src/refactor.js
Srihari123456: I'll remove this in the updated commit

javascript/src/refactor.js
Srihari123456: @ketkarameya ```parameterList``` contains the details of the parameters to be removed across all the functions in the file. On the other hand, ```parent.params``` denotes only the arguments of a specific function declaration. Hence ```parameterList.length``` need not be same as ```parent.params.length```

ketkarameya: Then how do we know that `parent` is the correct method declaration? Will this handle overriden methods?

Srihari123456: > Can we move this check outside of this for loop ? From what i understand, you are checking if the `parent` function declaration name is same as the function name of `parameter_list` right? Maybe this could be like the first line inside `enter` ?

```parameterList``` is a list of objects, each of which has ```functionName, parameterIndex, value```. So even if its brought under ```enter```, we need to iterate of the list of objects available in ```parameterList```

ketkarameya:
> Hey @Srihari123456 ,
> 
> I was looking at the goal of the PR. I have one concern here - isn't the literal `true` redundant? Couldn't that be eliminated too?
> 
> For **Case 1** - should'nt the refactoring be
> 
> ```
> function f1() {
>     console.log("in if")
> }
> 
> f1()
> ```
> 
> Note I have removed the literal `true` from argument. Unrelated, I am a Javascript noob, but I was surprised to find that JS allows one to pass `true` as a parameter for `f1` . Similarly for other cases.

@Srihari123456 I just wanted to make sure that you didn't miss this comment.

javascript/src/refactor.js
Srihari123456: I'll remove this in the updated commit

javascript/src/refactor.js
Srihari123456: I'll remove this in the updated commit

javascript/src/refactor.js
ketkarameya: > 

got it.

Srihari123456:
> > Hey @Srihari123456 ,
> > I was looking at the goal of the PR. I have one concern here - isn't the literal `true` redundant? Couldn't that be eliminated too?
> > For **Case 1** - should'nt the refactoring be
> > ```
> > function f1() {
> >     console.log("in if")
> > }
> > 
> > f1()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note I have removed the literal `true` from argument. Unrelated, I am a Javascript noob, but I was surprised to find that JS allows one to pass `true` as a parameter for `f1` . Similarly for other cases.
> 
> @Srihari123456 I just wanted to make sure that you didn't miss this comment.

@ketkarameya Yes I have figured the way to eliminate redundant literals and actually wanted to implement it this way. So I thought of raising a separate ticket for the same. Anyways I'll include this in my updated commit as well

javascript/src/refactor.js
Srihari123456: > Then how do we know that `parent` is the correct method declaration? Will this handle overriden methods?

In order to handle overriden functions, I've used your clue : checking ```parameterList[j].length == parent.params.length```
I'll update this in the new commit and let you know

javascript/src/refactor.js
Srihari123456: Yes added necessary comments

javascript/src/refactor.js
Srihari123456: Added necessary comments

Srihari123456:
> I have done my first pass over the PR.

@ketkarameya I've commited the necessary changes. Please check that when you can

ketkarameya:
Hey @Srihari123456 i am out on vacation until end of the week. I ll take a look at it as soon as I am back.

ketkarameya:
@Srihari123456 can you please look into why these test cases are failing? Also I was wondering if you could add an end to end test case for this new feature you have implemented (The scenarios similar to the ones you have added in the PR description)

Srihari123456:
Will check this

PR #190, opened 2022-06-07
Adding enum declaration cleanup for Java and Kotlin.

### LINKED ISSUE #187: [Piranha Generic] Incomplete Enum cleanup compared to the behavior with PiranhaJava
PiranhaJava basically removes not only the treated part, but it also removes the enum values which are now going to be obsolete and have no usages.

But I don't see the same behaviour when working with `Piranha Generic`.

Here's an example

**Original Java class**

```java
class SampleJava {

  public void sampleMethod(ExperimentInterface exp) {
    if (exp.isToggleEnabled(MyEnum.SAMPLE_STALE_FLAG)) {
      System.out.println("SAMPLE_STALE_FLAG is enabled");
    } else {
      System.out.println("SAMPLE_STALE_FLAG is disabled");
    }
  }

  public void sampleMethod1(ExperimentInterface exp) {
    if (exp.isToggleDisabled(ExpEnum.SAMPLE_STALE_FLAG)) {
      System.out.println("SAMPLE_STALE_FLAG is disabled");
    } else {
      System.out.println("SAMPLE_STALE_FLAG is enabled");
    }
  }

  public enum MyEnum{
    SAMPLE_STALE_FLAG,
    DRAGON_WARRIOR
  }
}
```

Now with the below arguments and configuration file, when I run the piranha cli, I notice the output class looks like below

**Altered Java class by Piranha**

```java
class SampleJava {

  public void sampleMethod(ExperimentInterface exp) {
    System.out.println("SAMPLE_STALE_FLAG is enabled");
  }

  public void sampleMethod1(ExperimentInterface exp) {
    if (exp.isToggleDisabled(ExpEnum.SAMPLE_STALE_FLAG)) {
      System.out.println("SAMPLE_STALE_FLAG is disabled");
    } else {
      System.out.println("SAMPLE_STALE_FLAG is enabled");
    }
  }

  public enum MyEnum{
    SAMPLE_STALE_FLAG,
    DRAGON_WARRIOR
  }
}
```

The java based piranha was removing the stale entry in my enum `MyEnum`, but piranha generic doesn't do that. It only removes the enum reference from the method, but not from the actual enum as well.

`piranha_arguments.toml`

```
language = ["java"]
substitutions = [
    ["stale_flag_name", "MyEnum.SAMPLE_STALE_FLAG"],
    ["treated", "true"],
    ["treated_complement", "false"],
]
```

`rules.toml`

```
[[rules]]
name = "replace_isToggleEnabled_with_boolean_literal"
query = """((
    (method_invocation 
        name : (_) @name
        arguments: ((argument_list 
                        ([
                          (field_access field: (_)@argument)
                          (_) @argument
                         ])) )
            
    ) @method_invocation
)
(#eq? @name "isToggleEnabled")
(#eq? @argument "@stale_flag_name")
)"""
replace_node = "method_invocation"
replace = "@treated"
groups = [ "replace_expression_with_boolean_literal"]
holes = ["treated", "stale_flag_name"]

#
# For @stale_flag_name = STALE_FLAG and @treated = true
# Before :
#  exp.isToggleDisabled(STALE_FLAG)
# After :
#  false
#
[[rules]]
name = "replace_isToggleDisabled_with_boolean_literal"
query = """((
    (method_invocation 
        name : (_) @name
        arguments: ((argument_list 
                        ([
                          (field_access field: (_)@argument)
                          (_) @argument
                         ])) )
            
    ) @method_invocation
)
(#eq? @name "isToggleDisabled")
(#eq? @argument "@stale_flag_name")
)"""
replace_node = "method_invocation"
replace = "@treated_complement"
groups = [ "replace_expression_with_boolean_literal"]
holes = ["treated_complement", "stale_flag_name"]
```

The other difference I noticed was that I now had to give the qualified name of the enum 

```
    ["stale_flag_name", "MyEnum.SAMPLE_STALE_FLAG"],
```

in my configuration, which wasn't the case when dealing with `PiranhaJava`

ketkarameya:
This PR fixes #187.
This PR adds new cleanup rules and E2E test scenarios for deleting all the content of the file when a source file has no declarations in it. Declarations considered for Java- `Class`, `Enum` and `Interface`; and for Kotlin - `Class`, `Object`, `Property`, `Function`.
The Piranha-CLI will delete the file itself instead of erasing the content. 

**Before the PR:** 
Before:
```
enum Abc {
 STALE_FLAG
}
```
 After 
```
enum Abc {
}
```

**After the PR:** 
Before:
```
enum Abc {
 STALE_FLAG
}
```
 After (delete the enum)
```
<>
```

### DIFF
--- a/generic/piranha/src/tests/test_piranha_java.rs
+++ b/generic/piranha/src/tests/test_piranha_java.rs
@@ -18,7 +18,7 @@ static LANGUAGE: &str = "java";
 #[test]
 fn test_java_scenarios_treated_ff1() {
   initialize();
-  run_test(LANGUAGE, "feature_flag_system_1", "treated", 1);
+  run_test(LANGUAGE, "feature_flag_system_1", "treated", 2);
 }
 
 #[test]
@@ -30,7 +30,7 @@ fn test_java_scenarios_treated_ff2() {
 #[test]
 fn test_java_scenarios_control_ff1() {
   initialize();
-  run_test(LANGUAGE, "feature_flag_system_1", "control", 1);
+  run_test(LANGUAGE, "feature_flag_system_1", "control", 2);
 }
 
 #[test]
--- a/generic/piranha/test-resources/java/feature_flag_system_1/control/input/XPFlagCleanerPositiveCases.java
+++ b/generic/piranha/test-resources/java/feature_flag_system_1/control/input/XPFlagCleanerPositiveCases.java
@@ -33,6 +33,10 @@ class XPFlagCleanerPositiveCases {
     SOME_OTHER_FLAG
   }
 
+  enum TestExperimentName {
+    STALE_FLAG,
+  }
+
   @Retention(RetentionPolicy.RUNTIME)
   public @interface Autorollout {
     boolean staged() default false;
--- a/generic/piranha/test-resources/java/feature_flag_system_1/control/input/XPFlagCleanerPositiveCases.java
+++ b/generic/piranha/test-resources/java/feature_flag_system_1/control/input/XPFlagCleanerPositiveCases.java
@@ -33,6 +33,9 @@ class XPFlagCleanerPositiveCases {
     SOME_OTHER_FLAG
   }
 
+  enum TestEmptyEnum {
+  }
+
   enum TestExperimentName {
     STALE_FLAG,
   }
--- a/generic/piranha/demo/java/TestEnum.java
+++ b/generic/piranha/demo/java/TestEnum.java
@@ -1,0 +1,18 @@
+/**
+ * Copyright (c) 2022 Uber Technologies, Inc.
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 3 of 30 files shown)


ketkarameya:
But what if the enum is the only type declaration in the file? Should this feature also support deleting files ?

generic/piranha/test-resources/java/feature_flag_system_1/control/input/XPFlagCleanerPositiveCases.java
```diff
@@ -33,6 +33,10 @@ enum AnotherTestExperimentName {
     SOME_OTHER_FLAG
   }
```
mkr-plse, line 35: can we add a test case where the input is 
```
enum TestEmptyEnum {
}
```
Essentially, to avoid deleting enums that are empty to begin with.

mkr-plse (commented):
changes are fine. can we add a test case as described in the comments?

generic/piranha/test-resources/java/feature_flag_system_1/control/input/XPFlagCleanerPositiveCases.java
ketkarameya: Done

generic/piranha/src/tests/test_piranha_java.rs
```diff
@@ -18,7 +18,7 @@ static LANGUAGE: &str = "java";
 #[test]
 fn test_java_scenarios_treated_ff1() {
   initialize();
-  run_test(LANGUAGE, "feature_flag_system_1", "treated", 1);
+  run_test(LANGUAGE, "feature_flag_system_1", "treated", 2);
```
mkr-plse, line 21: what is 2?

mkr-plse (approved):
lgtm

generic/piranha/src/tests/test_piranha_java.rs
ketkarameya: Number of files updated in the E2E case.

PR #192, opened 2022-06-13
Adding Swift parser

ketkarameya:
Added Swift parser and a basic demo for running user provided swift rules. 
Note this PR does not add piranha refactoring support for swift. 

Since the PR itself does not add any feature that is used by core Piranha I have added no tests.


generic/piranha/src/config.rs
```diff
@@ -52,20 +52,20 @@ pub(crate) fn read_config_files(
   // Read the language specific cleanup rules and edges
   let language_rules: Rules = read_toml(
     &path_to_language_specific_cleanup_config.join("rules.toml"),
-    false,
+    true,
```
ketkarameya, line 55: This allows cleanup rules for a language to be empty.  Similar for the below three similar updates.

mkr-plse (approved):

PR #204, opened 2022-07-11
User option for - deleting an empty file and replacing 3+ consecutive blank lines with 2.

ketkarameya:
This PR adds a user option for 
(i) deleting (or not deleting) an file that ends up being empty after Piranha cleanups 
(ii) Allow users to perform regex-replace over the file after Piranha has cleaned it up. 

## Context: 
* For file deletion: 
E.g., When cleaning up `strings` files, the users had a requirement that they didn't want to delete the file if it was empty after piranha cleanup. 

* User specified Regex replace after piranha cleanup : 
We do not want to guarantee the any good formatting after piranha cleanup, because we assume that there are handy formatters like google java format, swift lint, ... However, for some files like `strings` these formatters are not available. So in cases like these, users can replace 3+ consecutive blank lines with two (based on their specific scenario) after piranha cleanup to fix stuff.  The updated demo shows the applicability of this feature.

### DIFF
--- a/polyglot/piranha/src/models/source_code_unit.rs
+++ b/polyglot/piranha/src/models/source_code_unit.rs
@@ -4,12 +4,13 @@ use std::{
   path::{Path, PathBuf},
 };
 
+use regex::Regex;
 use tree_sitter::{InputEdit, Node, Parser, Range, Tree};
 use tree_sitter_traversal::{traverse, Order};
 
 use crate::utilities::{eq_without_whitespace, tree_sitter_utilities::get_tree_sitter_edit};
 
-use super::edit::Edit;
+use super::{edit::Edit, piranha_arguments::PiranhaArguments};
 
 // Maintains the updated source code content and AST of the file
 #[derive(Clone)]
@@ -43,16 +44,24 @@ impl SourceCodeUnit {
   }
 
   /// Writes the current contents of `code` to the file system.
-  pub fn persist(&self) {
-    if self.code.as_str().is_empty(){
-      _ = fs::remove_file(&self.path).expect("Unable to Delete file");
-    }else{
-      fs::write(&self.path, self.code.as_str()).expect("Unable to Write file");
+  /// Based on the user's specifications, this function will delete a file if empty
+  /// and replace three consecutive newline characters with two.
+  pub(crate) fn persist(&self, piranha_arguments: &PiranhaArguments) {
+    if self.code.as_str().is_empty() {
+      if piranha_arguments.delete_file_if_empty() {
+        _ = fs::remove_file(&self.path).expect("Unable to Delete file");
+      }
+    } else {
+      let content = if piranha_arguments.delete_consecutive_new_lines() {
+        let regex = Regex::new(r"\n\s*\n(\s*\n)").unwrap();
+        regex.replace_all(&self.code(), "\n${1}").to_string()
+      } else {
+        self.code()
+      };
+      fs::write(&self.path, content).expect("Unable to Write file");
     }
-    
   }
 
- 
   pub(crate) fn apply_edit(&mut self, edit: &Edit, parser: &mut Parser) -> InputEdit {
     // Get the tree_sitter's input edit representation
     self._apply_edit(edit.replacement_range(), edit.replacement_string(), parser)
@@ -85,14 +94,17 @@ impl SourceCodeUnit {
     // Handle errors, like removing extra comma.
     self.remove_additional_comma_from_sequence_list(parser);
     if self.ast.root_node().has_error() {
-      panic!("Produced syntactically incorrect source code {}", self.code());
-  }
+      panic!(
+        "Produced syntactically incorrect source code {}",
+        self.code()
+      );
+    }
     ts_edit
   }
 
   /// Sometimes our rewrite rules may produce errors (recoverable errors), like failing to remove an extra comma.
   /// This function, applies the recovery strategies.
-  /// Currently, we only support recovering from extra comma. 
+  /// Currently, we only support recovering from extra comma.
   fn remove_additional_comma_from_sequence_list(&mut self, parser: &mut Parser) {
     if self.ast.root_node().has_error() {
       let c_ast = self.ast.clone();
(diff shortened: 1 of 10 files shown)


ketkarameya:
Flaky coveralls. :|

polyglot/piranha/src/models/source_code_unit.rs
```diff
@@ -43,16 +44,24 @@ impl SourceCodeUnit {
...
+  /// and replace three consecutive newline characters with two.
+  pub(crate) fn persist(&self, piranha_arguments: &PiranhaArguments) {
+    if self.code.as_str().is_empty() {
+      if piranha_arguments.delete_file_if_empty() {
+        _ = fs::remove_file(&self.path).expect("Unable to Delete file");
+      }
+    } else {
+      let content = if piranha_arguments.delete_consecutive_new_lines() {
+        let regex = Regex::new(r"\n\s*\n(\s*\n)").unwrap();
+        regex.replace_all(&self.code(), "\n${1}").to_string()
```
lazaroclapp, line 57: a) Will this work to replace 4 empty new lines in a row with one?
b) Will it do so efficiently? Without first substituting 4 empty lines by 3, then by 2, then by 1.

Or is there a constraint in the rest of our transformation that we will only ever insert a single new empty line that might need to be deleted?

polyglot/piranha/src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -1,3 +1,13 @@
+use std::{
```
lazaroclapp, line 1: Going to assume that this placement of unit tests (rather than under `polyglot/piranha/src/test` is idiomatic for Rust, but still noting that it's a bit weird to me coming from either our Java or Go codebases.

polyglot/piranha/src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -96,3 +106,98 @@ fn test_apply_edit_negative() {
...
+  let source_code = "class Test {
+    public void foobar() {
+
+      System.out.println(\"Hello World!\");
+
+
+      System.out.println();
+    }
+  }";
+  fn check(temp_dir: &TempDir) -> Result<bool, io::Error> {
```
lazaroclapp, line 165: Wonder if this `check` function can't be shared between this test case and the next, it just needs to be able to take `expect_str` as a parameter, right?

polyglot/piranha/src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -96,3 +106,98 @@ fn test_apply_edit_negative() {
...
+    let paths = fs::read_dir(temp_dir)?;
+    let path = paths.find_or_first(|_| true).unwrap()?;
+    let actual_content = fs::read_to_string(path.path().as_path())?;
+    return Ok(actual_content.eq(&actual_content));
+  }
+  assert!(execute_persist_in_temp_folder(source_code, &args, &check)?);
+  Ok(())
+}
```
lazaroclapp, line 203: Would it make sense to test an example with more than 2 empty lines in a row? (see my other comment)

lazaroclapp (approved):
Not a lot of Rust expertise, so just general (and very much optional) comments about regexes and testing strategy 🙂 

Maybe the description of the PR would benefit from a bit more context about when do we expect to see these empty lines being introduced and why we want to remove them?

(I _can_ think of cases like that from the old implementation of Piranha, such as removing an annotation that happens to be on its own line, but, still, worth calling them out? Plus, depending on how the implementation handles removing statements vs expressions, I could have imagined it no longer being a problem, so it's interesting that it is. On the other hand, wanting to remove empty files after a rewrite is a lot more intuitive to me).

polyglot/piranha/src/models/unit_tests/source_code_unit_test.rs
ketkarameya: Rust's module system is very flexible and weird.  I too am figuring out the best organization strategy. This is something that  @mkr-plse  and I had agreed upon for the time being. 

 So this is how I have designed my tests: 
(i) Integration tests  (End to End tests)  They are as one would expect in `polyglot/piranha/src/tests`. 
([ref](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html))
They basically test functionality of end to end refactorings, 
(ii) Unit tests: Function level tests.
 In`Rust` one to writes unit test within the same module where the logic is implemented ([Ref](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html)). In this way we can write tests for private methods too . Now adding tests, within the same file was too much . So for example [here](https://sourcegraph.com/github.com/uber/piranha/-/blob/polyglot/piranha/src/models/source_code_unit.rs?L129) I have annotated and added the "unit-test" sub module `souce_code_unit_test.rs` in `source_code_unit.rs`. 
Since I can basically give a `path` to the submodule (which is checked at compile time) I had liberty of putting the tests where-ever.  So i collected all the unit-tests for `models` and put them in one directory. 

Other options we had thought of : 


(i) Convert each `.rs.` file for which I want a unit test to a folder like : 
`source_code_unit`
       | - `mod.rs` (implements the logic)
       | - `test.rs` (implements the test)

(ii)  have a directory `src/unit-tests`.

polyglot/piranha/src/models/source_code_unit.rs
ketkarameya: So, I want to give no guarantee for formatting of code after the refactoring is performed. Tools like`google-java-format` or `swift-lint` or ... could be run as a followup. 
However, formats like `strings` dont have such handy formatters. So this `regex replace` particularly fixes the formatting for the `strings` file format.  Note, by default the value of this is `false` - [ref](https://github.com/uber/piranha/pull/204/files#diff-17b1537c28ddaf25ae2355afe2161ce8a3e813eaf27d06523109a7a66bfbbbf6R123)

polyglot/piranha/src/models/unit_tests/source_code_unit_test.rs
ketkarameya: good catch. I will update.

polyglot/piranha/src/models/unit_tests/source_code_unit_test.rs
ketkarameya: Added.

polyglot/piranha/src/models/source_code_unit.rs
ketkarameya: updated the regex.

PR #208, opened 2022-08-09
Try and match all global rules to each file when no rule contains a hole

ketkarameya:
Until now we assumed that the user will provide the rules and some extra arguments that parameterizes the rules. For instance, we provide the  stale flag name and its treated value along with the rules that express how to transform the feature flag API.  So earlier our strategy was: 
Scan all files in the target, and check if they contain the concrete values of the `holes`, if so parse it. 
This is very similar to how Piranha-Java workflow, where we first `grep` and run Piranha on specific files.

But now we have a usecase (simple one step structural replacement) where the user may not want to provide the "extra arguments" . Rather they would want to try to hardcoded rules.  For instance, when I want to perform a simple structural replacement that replaces API A with API A', by renaming some call site. 
To support his usecase, I have added the predicate `no_global_rules_with_holes` in `piranha.rs`. 

I have also added a test case for this. 

I have also have resolved a bug related to global rules, where we would unnecessarily add a rule even if it already existed in the "global rule list".


polyglot/piranha/src/models/rule.rs
```diff
@@ -28,13 +28,13 @@ use super::{
 
 static FEATURE_FLAG_API_GROUP: &str = "Feature-flag API cleanup";
 
-#[derive(Deserialize, Debug, Clone, Hash, PartialEq, Eq, Default)]
```
lazaroclapp, line 31: We are removing the option to compare rules for equality here, correct? Is this related to the new type of rules or just an unrelated refactoring/removal? Was this something that worked before but we don't need or something that was broken?

(Should be fine either way, just want to make sure I understand a bit what's happening)

Edit: Ok, looks to me like below we are using a custom definition of equality for `Rule` at least. Wonder if it makes sense to implement `Rule::Eq(...)` somehow.

polyglot/piranha/src/piranha.rs
```diff
@@ -265,7 +267,8 @@ impl FlagCleaner {
       // Read the file
       .map(|f| (f.path(), read_file(&f.path()).unwrap()))
       // Filter the files containing the desired regex pattern
-      .filter(|x| pattern.is_match(x.1.as_str()))
+      .filter(|x| no_global_rules_with_holes ||
```
lazaroclapp, line 270: 👍

lazaroclapp (approved):
LGTM, one very small question below.

polyglot/piranha/src/models/rule.rs
ketkarameya: Yes. I think the most correct way would be to defile my own Eq and Hash. I ll do that as a followup .

ketkarameya: I had defined my own eq, then I thought IT was too much ceremony for such a simple check.  - https://github.com/uber/piranha/pull/208/commits/1100973c2ecae5a07a5b67d85f499aa738820a4f

PR #209, opened 2022-08-11
This is a general cleanup that makes fields optional rather than having the user pass empty string

ketkarameya:
To provide dummy rules, that is rules that have no query/replace (makes it easy to write recursive rules, as discussed in readme file). Instead of requiring the user to pass empty string we now make the field itself optional.


lazaroclapp (approved):
LGTM!

PR #211, opened 2022-08-11
Type Migration from `(Range, HashMap<String, String>)` to `Match` struct. 

ketkarameya:
The tuple basically captures the range of the query match and the mapping between query tags and string representation of the captured AST nodes.

Since this is a general maintenance based PR - I made two more refactoring: 
(i) Refactored "god method" from tree-sitter utilities 
(ii) ran clippy fix.

Edited:
Ok it seems I had a git accident. The earlier PR that lazaro #209 approved got merged without any commit in it (because I did some `rebase onto` hackery). 
This PR contains that change (#209 - basically commit [868d165](https://github.com/uber/piranha/pull/211/commits/868d1651801d3679d902ae0490c93bf9faa1c6bd)) and the changes described above. 
My apologies for this.


polyglot/piranha/src/models/matches.rs
```diff
@@ -0,0 +1,46 @@
...
+  /// Get the edit's replacement range.
+  pub(crate) fn range(&self) -> Range {
+    self.range
+  }
+
+  pub(crate) fn matches(&self) -> &HashMap<String, String> {
+    &self.matches
+  }
+
+
```
rajbarik, line 45: Remove the extra space

rajbarik, line 45:
```suggestion
```

polyglot/piranha/src/models/rule.rs
```diff
@@ -146,15 +142,25 @@ impl Rule {
...
+    panic!("No query pattern!")
   }
 
   pub(crate) fn replace(&self) -> String {
-    String::from(&self.replace)
+    if let Some(rp) = &self.replace{
+      return rp.to_string();
+    }
+    panic!("No replace pattern!")
+
```
rajbarik, line 163:
```suggestion
```

rajbarik (approved):

Issue #26, opened 2020-04-07
Output summary at the end of execution


mkr-plse:
When Piranha is run, output the summary of the run that includes the configuration under which it ran, the number of APIs considered for refactoring, the number of APIs refactored, etc. 
Related - issue #207

manishamishra2715:
hii..I am trying to integrate piranha using maven.Also I have configured all the required jar files inside pom.xml file.Also the maven test result is getting successful.But where can I get the expected output??
I m not able to analyse where the output is generated.Please help me with this

mkruber:
When you build it with the given options in the instruction, the refactoring should happen in place and whatever you provided as source must have been updated.

manishamishra2715:
Thanks for quick response...actually the source is not getting updated. I am running by project->Run As->maven test.Also tried with pom.xml->Run As->maven test.Is this the correct approach? 
can you please guide me , which one to run?

manishamishra2715:
Almost reached the solution referring the previous solutions to the issue raised

msridhar:
Glad it’s almost working! Two tips for the future:

1. If you’re having a problem with Piranha, please either open a new issue or comment on an issue reflecting the same problem. This issue is about something different.

2. If you’re having a build system integration problem, please provide output from running on the command line. Unfortunately we don’t have time to support IDE integrations at the moment.

manishamishra2715:
yes sure...thanks for info :)

pranavsb:
It would also be helpful to output "near misses" during execution?
Makes it easier to identify is the returnType or receiverType is slightly off, for instance. Or if the properties file path was specified incorrectly.


Does it make sense to log it in an optionally specified file or would that be too much overhead?

mkr-plse:
Great idea. We can do two things here: 
a) By default, we will print the summary of what happened as part of the refactoring. 
b) Optionally, if the user specifies an option,  we can log the activity of the refactoring process at the appropriate granularity.

The latter will be more expensive but will only be used for debugging the process of refactoring.

j1nma:
Could this also work independently from the in-place removals?

That is, run Piranha and only get this summary, and don't modify any files.

ketkarameya:
Yes. Polyglot Piranha allows one to express match-only rules. So the task you suggest could also be done.

j1nma:
❤️. Can we expect this sometime this year?

ketkarameya:
[...]

PR #212, opened 2022-08-12
Add support for `match-only` piranha rules

ketkarameya:
**Background**
Semantics of a Piranha Rewrite rule: It basically matches a source file to the `query`, and replaces this AST node captured by this query with the replace pattern. Moreover, it also adds the mappings of query tags to string representation of the matched ASTs to an internal table that piranha maintains during its execution. The mappings in this table is used by piranha when propagating new rewrite rules based on a recently applied one. 

**Goal:** Adding a `match-only` rules feature in piranha. 
We need this feature because: 
(i) We want to be able to look up AST patterns (not transform them) and propagate new rules based on the concrete match. For instance, if I want to say rename all the callsites `Foo.someMethod()` to `Foo.anotherMethod`. I would also want an ability to lookup type aliases of `Foo`  in the code base say `T` and propagate rules to  `T.someMethod()` to `T.anotherMethod()`. Note here we don't want to affect the type alias declaration `T`, we just want to look up and propagate new rules.  (look at test scenario added in this PR under `test-resources/java/find_and_propagate`
(ii) We want to summarize output #26 . There is a user requirement that, they want to be able to use piranha as a structural find rule that reports all occurrences of a pattern. (something like grep ). 

**Solution:** `match-only` is basically an identity rewrite rules i.e. replaces the matched AST with itself while updating the internal cache table.  For that purpose, I bypass the `apply_edit` action in this case and propagate it as identity edit in the rule graph.  We can also say that `match-only` produces side effects only. 

**Changes Performed**: 
* extracted out the `propagate` method so that I can use the logic for both `match ` and for `match-replace`  
* Added fields in `SourceCodeUnit` that collect each match or rewrite performed by Piranha
* Added a test around the structural find functionality. 
  * `test-resources/java/structural_find` tests the `match` functionality for the structural find scenario (use case 2)
  * `test-resources/java/find_and_propagate` tests the `match` functionality  for the find and propagate scenario (use case 1)

The `src` files changed are only 4. Its the test scenarios that make it 18 files :| 

As a follow up PR I will add command line options to write the `matches` and `rewrites` recorded into a JSON. 

While there were comments inline for the implementation, I copy pasted and adapted it to make method level.

### Update 

Addressed comments in [commit_1](https://github.com/uber/piranha/pull/212/commits/ac96d51ff2f15041d730b34bc22851d647a3a85c) , [commit_2](https://github.com/uber/piranha/pull/212/commits/56bd6215946cff13514fcb963bc2c9782fce4275)


polyglot/piranha/src/models/rule.rs
```diff
@@ -260,15 +268,29 @@ impl Rule {
...
       ) {
+        output.push(p_match);
+      }
+    }
+    output
+  }
+
+
+  /// Gets the first match for the rule in `self`
+  pub(crate) fn get_edit(
```
ketkarameya, line 279: Basically here, I extract `get_matches` out of `get_edit`

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +60,136 @@ impl SourceCodeUnit {
...
-      // Add all the (code_snippet, tag) mapping to the substitution table.
-      self.add_to_substitutions(edit_1.matches());
-
-      let mut current_edit = applied_ts_edit;
-      let mut current_rule = rule.clone();
-      let mut next_rules_stack: VecDeque<(String, Rule)> = VecDeque::new();
-
-      // Perform the parent edits, while queueing the Method and Class level edits.
-      // let file_level_scope_names = [METHOD, CLASS];
-      loop {
```
ketkarameya, line 87: extracted a method from line 81 until 130 into the `propagate` method.

polyglot/piranha/src/piranha.rs
```diff
@@ -40,10 +41,20 @@ use crate::{
 use std::collections::VecDeque;
 use tree_sitter::{InputEdit, Node};
```
ketkarameya, line 43: Add Documentation.

polyglot/piranha/src/piranha.rs
```diff
@@ -197,6 +282,24 @@ impl FlagCleaner {
     self.relevant_files.values().cloned().collect_vec()
   }
```
ketkarameya, line 284: Add Documentation here.

polyglot/piranha/src/models/rule.rs
```diff
@@ -233,17 +238,20 @@ impl Rule {
     None
   }
 
+
   /// Gets the first match for the rule in `self`
-  pub(crate) fn get_edit(
+  pub(crate) fn get_matches(
```
ketkarameya, line 243: Add unit test

polyglot/piranha/src/tests/mod.rs
```diff
@@ -34,14 +36,37 @@ fn initialize() {
...
+  let args = PiranhaArguments::new(CommandLineArguments {
+    path_to_codebase: format!("{path_to_test_ff}/input/"),
+    path_to_configurations: format!("{path_to_test_ff}/configurations/"),
+  });
+  let (updated_files, _matches, rewrites) = execute_piranha(&args);
+  // Checks if there are any rewrites performed for the file
+  assert!(rewrites.values().flat_map(|x|x.iter()).count() > 0);
+
+  for r in rewrites.values().flat_map(|x|x.iter()).collect_vec() {
+    print!("{:?}", r);
```
lazaroclapp, line 68: Same question as above about (debug?) prints.

polyglot/piranha/test-resources/java/find_and_propagate/configurations/piranha_arguments.toml
```diff
@@ -0,0 +1,13 @@
...
+# except in compliance with the License. You may obtain a copy of the License at
+# <p>http://www.apache.org/licenses/LICENSE-2.0
+# 
+# <p>Unless required by applicable law or agreed to in writing, software distributed under the
+# License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+language = ["java"]
+substitutions = []
```
lazaroclapp, line 13: Newline?

polyglot/piranha/test-resources/java/find_and_propagate/configurations/rules.toml
```diff
@@ -0,0 +1,40 @@
...
+# Before 
+#  exp.isToggleEnabled(Experiment.STALE_FLAG)
+# After 
+#  true
+#
+[[rules]]
+name = "find_interface_extension"
+query = """(
+(class_declaration name: (_)@class_name 
+ 				interfaces: (super_interfaces (_) @t)
```
lazaroclapp, line 25: Is this amount of indention normal/expect for toml rules?

polyglot/piranha/test-resources/java/find_and_propagate/input/TestClassUsage.java
```diff
@@ -0,0 +1,22 @@
...
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.piranha;
+
+class TestClassUsage extends TestClass {
```
lazaroclapp, line 16: Wouldn't `TestSubClass` be a better name here? This isn't so much using `TestClass` as subtyping it.

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
+  /// This is implements the main algorithm of piranha.
+  /// Parameters:
+  /// * `rule` : the rule to be applied
+  /// * `rule_store`: contains the input rule graph.
+  ///
+  /// Algo:
+  /// * check if the rule is match only
+  /// ** IF not (i.e. it is a rewrite):
+  /// *** Get the first match of the rule for the file
+  ///  (We only get the first match because the idea is that we will apply this change, and keep calling this method `_apply_rule` until all
```
lazaroclapp, line 85: Until all what?

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
 
-  // Applies the rule to the first match in the source code
-  // This is implements the main algorithm of piranha.
+  /// Applies the rule to the first match in the source code
+  /// This is implements the main algorithm of piranha.
+  /// Parameters:
+  /// * `rule` : the rule to be applied
+  /// * `rule_store`: contains the input rule graph.
+  ///
+  /// Algo:
```
lazaroclapp, line 81: Unless this is a Rust convention I am unaware of, let's just use "Algorithm" in full.

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
+  /// * check if the rule is match only
+  /// ** IF not (i.e. it is a rewrite):
+  /// *** Get the first match of the rule for the file
+  ///  (We only get the first match because the idea is that we will apply this change, and keep calling this method `_apply_rule` until all
+  /// *** Apply the rewrite
+  /// *** Update the substitution table
+  /// *** Propagate the change
+  /// ** Else (i.e. it is a match only rule):
+  /// *** Get all the matches , and for each match
```
lazaroclapp, line 90: No space before the comma.

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
+    // Propagate each match. Note that,  we pass a identity edit (where old range == new range) in to the propagate logic. 
+    // The next edit will be applied relative to the identity edit. 
+    else {
+      for m in rule.get_matches(&self.clone(), rule_store, scope_node, true) {
+        self.matches_mut().push((rule.clone(), m.clone()));
+
+        self.add_to_substitutions(m.matches());
+        
+        // Create identity edit, that replaces a range with itself. 
+        let identity_edit = InputEdit {
```
lazaroclapp, line 131: Fine to do in a follow up PR, but remember offline discussion about naming this something other than edit, specially if not actually applied to the source.

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
+    query_again
+  }
+
+  /// This is the propagation logic of the Piranha's main algorithm.
+  /// Parameters:
+  ///  * Rule `rule` that was just applied
+  ///  * `applied_ts_edit` -  it's(`rule`'s) application site (in terms of replacement range)
+  ///  * `rule_store` - contains the input "rule graph"
+  ///
+  /// Algo:
```
lazaroclapp, line 152: Is this by convention? I don't see anything in [RFC1574](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.htm). Probably just use "Algorithm"

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
+      }
+    }
+    query_again
+  }
+
+  /// This is the propagation logic of the Piranha's main algorithm.
+  /// Parameters:
+  ///  * Rule `rule` that was just applied
+  ///  * `applied_ts_edit` -  it's(`rule`'s) application site (in terms of replacement range)
+  ///  * `rule_store` - contains the input "rule graph"
```
lazaroclapp, line 150: This parameter list is incomplete and out of order with the actual function.

polyglot/piranha/src/piranha.rs
```diff
@@ -59,76 +72,146 @@ impl SourceCodeUnit {
...
+      // Find the rules to be applied in the "Parent" scope that match any parent (context) of the changed node in the previous edit
+      if let Some(edit) = Rule::get_edit_for_context(
+        &self.clone(),
+        current_edit.start_byte,
+        current_edit.new_end_byte,
+        rules_store,
+        &next_rules_by_scope[PARENT],
+      ) {
+        self.rewrites_mut().push(edit.clone());
+        #[rustfmt::skip]
```
lazaroclapp, line 198: Why do we need to skip formatting here?

polyglot/piranha/src/piranha.rs
lazaroclapp: Also, I don't fully see where the check is that distinguishes between a real edit and this before applying. Can I get a quick rundown of the logic here?

polyglot/piranha/src/piranha.rs
ketkarameya: Lol! its not. Thanks for pointing out. I ll make a change.

polyglot/piranha/test-resources/java/find_and_propagate/configurations/piranha_arguments.toml
ketkarameya: Addressed.

polyglot/piranha/test-resources/java/find_and_propagate/configurations/rules.toml
ketkarameya: Actually its the formatting for the tree-sitter query s-expression. I had over-indented there.

polyglot/piranha/test-resources/java/find_and_propagate/input/TestClassUsage.java
ketkarameya: Yes.

polyglot/piranha/test-resources/java/structural_find/input/XPFlagCleanerPositiveCases.java
```diff
@@ -0,0 +1,288 @@
...
+package com.uber.piranha;
+
+import dagger.Provides;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.inject.Inject;
+
+class XPFlagCleanerPositiveCases {
```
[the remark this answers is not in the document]
ketkarameya: Makes sense. I have updated two instances to SOME_OTHER_FLAG, and now I am checking for 20 occurences.

polyglot/piranha/src/piranha.rs
ketkarameya: Fixed.

polyglot/piranha/src/tests/mod.rs
ketkarameya: Yes. Thank you for pointing out.

polyglot/piranha/src/tests/mod.rs
```diff
@@ -34,14 +36,37 @@ fn initialize() {
...
+fn run_match_test(relative_path_to_tests: &str, number_of_matches: usize) {
   let path_to_test_ff = format!("test-resources/{relative_path_to_tests}");
 
   let args = PiranhaArguments::new(CommandLineArguments {
     path_to_codebase: format!("{path_to_test_ff}/input/"),
     path_to_configurations: format!("{path_to_test_ff}/configurations/"),
   });
-  let updated_files = execute_piranha(&args);
+  let (_updated_files, matches,_rewrites) = execute_piranha(&args);
+  print!("{:?}", matches);
```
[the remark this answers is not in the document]
ketkarameya: This is a debugginf print. Removed the print.

polyglot/piranha/src/piranha.rs
ketkarameya: One line [105](https://github.com/uber/piranha/pull/212/files/9cf8a1f970527f37d341e9f4c5fae0d1ac260266#diff-31027e2b569bb16bf28a5b188a0ca6edda8c55cda1ff87475f47e0366a84d778R105) we check if the rule is "rewrite rule"  - `!rule.is_match_only_rule()`. So if it is a rewrite rule : 
(i) It will get the `Edit`
(ii) Update the substitution table 
(iii)  apply it to the source code unit ([l. 114]
(iv) Propagate the edit 
(https://github.com/uber/piranha/pull/212/files/9cf8a1f970527f37d341e9f4c5fae0d1ac260266#diff-31027e2b569bb16bf28a5b188a0ca6edda8c55cda1ff87475f47e0366a84d778R114) ) - `apply_edit`.

IF its a match-only rule, we get all the matches and for each match : 
(i) Update substitution table 
(ii) Create this identity edit 
(iii) Propagate the identity edit

polyglot/piranha/src/piranha.rs
ketkarameya: Its legacy stuff. I ll remove it.

polyglot/piranha/src/piranha.rs
ketkarameya: I have done the refactoring to make the intention more clear. The change was simple and straightforward.,

polyglot/piranha/src/piranha.rs
ketkarameya: Addressed

lazaroclapp (approved):
LGTM. Still not super confident in my ability to review Rust code, but as far as I can tell it all makes sense :)

ketkarameya:
> LGTM. Still not super confident in my ability to review Rust code, but as far as I can tell it all makes sense :)

Awesome! Thanks a lot. I think this was a very good review. Rust isn't that bad eh!

PR #216, opened 2022-08-17
Reorganize the code a bit so that Piranha can be used as a binary and as library

ketkarameya:
This is a refactoring.  
Trying to make the code structure conform the recommended guideline - https://doc.rust-lang.org/cargo/guide/project-layout.html 
Our `piranha.rs` was actually `lib.rs` we just were not exposing it as a library. This PR does this refactoring (and manages all the access modifiers appropriately).

 While doing this, I also realized that I was unnecessarily exposing `SourceCodeUnit`, while all I needed was its content (Which is part of `PiranhaOutputSummary`). So I made that `pub(crate)` and updated the tests to use `PiranhaOutputSummary`. 

------------
ran `cargo fmt`

### DIFF
--- a/polyglot/piranha/Cargo.toml
+++ b/polyglot/piranha/Cargo.toml
@@ -8,6 +8,14 @@ include = [
     "src/"
 ]
 
+[[bin]]
+name = "polyglot_piranha"
+path = "src/main.rs"
+
+[lib]
+name = "polyglot_piranha"
+path = "src/lib.rs"
+
 [build-dependencies]
 cc = "1.0.73"
 
--- a/polyglot/piranha/src/lib.rs
+++ b/polyglot/piranha/src/lib.rs
@@ -11,7 +11,10 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
-use models::{piranha_arguments::PiranhaArguments, source_code_unit::SourceCodeUnit, piranha_output::PiranhaOutputSummary};
+use models::{
+  piranha_arguments::PiranhaArguments, piranha_output::PiranhaOutputSummary,
+  source_code_unit::SourceCodeUnit,
+};
 
 mod config;
 pub mod models;
@@ -28,7 +31,8 @@ use log::info;
 use regex::Regex;
 use tree_sitter::{Parser, Range};
 
-use crate::{models::{rule_store::RuleStore,},
+use crate::{
+  models::rule_store::RuleStore,
   utilities::{read_file, tree_sitter_utilities::get_match_and_replace_range},
 };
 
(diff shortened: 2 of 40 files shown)


polyglot/piranha/Cargo.toml
```diff
@@ -8,6 +8,14 @@ include = [
     "src/"
 ]
 
+[[bin]]
```
ketkarameya, line 11: Now piranha can be used as a  binary (like before ) and as a library.

ketkarameya:
Flaky coveralls. Note I have not changed any Java files.

polyglot/piranha/Cargo.toml
ketkarameya: currently library provides only one API `execute_piranha`

polyglot/piranha/src/lib.rs
```diff
@@ -11,6 +11,14 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
+use models::{piranha_arguments::PiranhaArguments, source_code_unit::SourceCodeUnit, piranha_output::PiranhaOutputSummary};
+
+mod config;
+pub mod models;
+#[cfg(test)]
```
lazaroclapp, line 18: Just to be clear: this applies only to the next element, right? So, `mod tests;` are only included in the test configuration, but `utilities` are included either way?

polyglot/piranha/src/lib.rs
```diff
@@ -290,7 +304,7 @@ impl FlagCleaner {
   }
 
   /// Performs cleanup related to stale flags
-  fn perform_cleanup(&mut self) {
+    fn perform_cleanup(&mut self) {
```
lazaroclapp, line 307: Is this indentation correct? I would expect the function definition to be indented with the comment above and one level out from the body.

polyglot/piranha/src/lib.rs
```diff
@@ -371,7 +385,7 @@ impl FlagCleaner {
   }
 
   /// Instantiate Flag-cleaner
-  fn new(args: &PiranhaArguments) -> Self {
+fn new(args: &PiranhaArguments) -> Self {
```
lazaroclapp, line 388: This indentation also looks weird to me.

polyglot/piranha/src/models/piranha_output.rs
```diff
@@ -36,4 +38,12 @@ impl PiranhaOutputSummary {
   pub(crate) fn rewrites(&self) -> &[Edit] {
     self.rewrites.as_ref()
   }
+
+    pub fn path(&self) -> PathBuf {
```
lazaroclapp, line 42: Indentation? This seems more deeply indented than the function before it. Are you mixing tabs and spaces?

lazaroclapp (commented):
I think code indentation is a bit weird here. Otherwise, the refactor makes sense to me.

polyglot/piranha/src/lib.rs
ketkarameya: This annotation applies to the next statement only. 
`mod tests` is used as "test-code" only.
`Utilities` is used as "main-code", and it is also used in "test-code". But because it is used in "main-code" we didnt annotate it.

polyglot/piranha/src/models/piranha_output.rs
ketkarameya: Yes I think so. I will run `cargo format` and `clippy` next.

lazaroclapp (approved):
LGTM assuming the last commit is just auto-formatting :)

PR #217, opened 2022-08-17
Feature - Add support for global tags

ketkarameya:
**Background** When we apply Piranha rules and update the substitution table, this table is "local" to each file i.e. Matches `m = {m1: v1, m2: v2, ..}` found for a rule `r`  in a file `f` are only accessible to all the rules being applied within this file and not outside of `f`. This is limiting, because there might be instances when we would want to have access to some tag match we found in another file. 
**When is this useful?** 
Plugin cleanups. 
Theoretically, it gives us the ability to to generate a new rewrite rule based on some value matched previously in **some other file**. So you can make to "global" matches in two files, and make rewrite rule that updates a third file using these two matches.  
 **Solution**:
I simply track the "global tags" and add them to the set of default substitutions available for each file (aka source code unit)
I identify a tag as global iff its name is prefixed with `GLOBAL_TAG.`. Note that, I provide a user configuration to change this prefix. 

**Tests**:
I have added two tests in Swift, because  thats the user case we have : 
(i) Find a switch case based on the returned string literal (in our case `Premium-Icon`) and delete it > Delete the corresponding enum (`premiumIcon`) > Find type aliases using this enum declaration (`SomePlugin`) > Delete a factory class that uses enum (`premiumIcon`) and type alias (`SomePluginFactory`) . Note this requires a global tag support, because enum (`premiumIcon`) was found in `Experiment.swift` while type alias (`SomePluginFactory`) was found in `SomeProtocol.swift`
(ii) Its the same test case, but it uses a custom global tag - `universal_tag`

----

I had to add `scope_config.toml` under Swift to support these test scenarios


polyglot/piranha/test-resources/swift/cascade_file_delete/expected/Experiment.swift
```diff
@@ -0,0 +1,32 @@
...
+ */
+import Experiment
+
+public enum SomePluginSwitch: PluginSwitch {
+    case defaultIcon
+
+    case defaultTitle
+    case premiumTitle
+
+
```
rajbarik, line 22: cleanup the spaces.

rajbarik:
We should track code coverage of polyglot piranha as part of PR merge.

rajbarik (approved):

PR #218, opened 2022-08-18
refactoring/ Cargo clippy and some cleanup 

ketkarameya:
Some auto-refactoring done by `clippy`. 
Followed by a small cleanup that removes the annotation hack I was using for serializing the type `tree_sitter::Range`. 
Now I simply, convert `tree_sitter::Range` to `Matches::Range` internally.  I do this so that I can "derive" whatever I Want using `Matches::Range`

No new test required.


lazaroclapp (approved):
Refactoring LGTM

PR #219, opened 2022-08-20
Introducing simple python interface for Polyglot Piranha library

ketkarameya:
At a higher level this PR introduces python interface for our polyglot piranha library. 
Some `Rust` context: We use [Pyo3](https://pyo3.rs/v0.17.1/) (and [maturin](https://github.com/PyO3/maturin)) for introducing "rust binaries for python interpreter".  Basically, as a user you put in those annotations `pyfunction`, `pymodule`, `pyo3(get)` ... appropriately, organize the code "correctly" and things work. As its [webpage](https://maturin.rs/#maturin) says : 

"
_This (maturin) project is meant as a zero configuration replacement for [setuptools-rust](https://github.com/PyO3/setuptools-rust) and [milksnake](https://github.com/getsentry/milksnake). It supports building wheels for python 3.5+ on windows, linux, mac and freebsd, can upload them to [pypi](https://pypi.org/) and has basic pypy support._
"

*  Added a new function in `lib.rs` - `run_piranha_cli` , which basically executes piranha for a given configurations upon a target. 
*  Added annotations for exposing the library `pyfunction`, `pymodule`, `pyo3(get)` ... 
*  Added a `pytest` that tests this interface via python. These 2 tests are basically the same integration test we run in our `rust` tests . The rewrite test is equivalent to [test_java_scenarios_treated_ff1](https://github.com/uber/piranha/blob/master/polyglot/piranha/src/tests/test_piranha_java.rs#L19) and match test is equivalent to [test_java_match_only](https://github.com/uber/piranha/blob/master/polyglot/piranha/src/tests/test_piranha_java.rs#L87)
*  Added a Github Workflow to execute the pytest . This also shows that, we can actually successfully run the python api generated by pyo3. 

For purely technical reasons (basically trying to replicate best practice of using `maturin` - [reference 1](https://maturin.rs/#mixed-rustpython-projects), [ reference 2](https://github.com/PyO3/maturin/tree/main/test-crates/pyo3-pure))
* added `pyproject.toml`  
*  Moved `README.md` and added (copy pasted from root) `LICENSE` at `cargo.toml` and `pyproject.toml` level
* Added all recommended fields . Trying to follow [PEP 621](https://peps.python.org/pep-0621/)
* Since I was doing this, I thought I might as well make `cargo.toml` consistent . - [Guideline](https://doc.rust-lang.org/cargo/reference/manifest.html)

Oh, and finally ! fixed `Multilingual` -> `Polyglot` in README.md

Now I know it sounds like a lot of changes, but i thought it was best we land them together so that CI integration works properly.


polyglot/piranha/Cargo.toml
```diff
@@ -1,12 +1,15 @@
 [package]
-authors = ["Ameya Ketkar <ketkara@uber.com>", "Murali Krishna Ramanathan <murali@uber.com>"]
+authors = ["Ameya Ketkar <ketkara@uber.com>", "Lazaro Clapp <lazaro@uber.com>", "Murali Krishna Ramanathan"]
```
lazaroclapp, line 2: Bit of an exaggeration, since I haven't contributed any code to PolyglotPiranha, but I guess code reviews sort of count (plus the original PiranhaJava)...

lazaroclapp: Also, unless there is an explicit organization field, maybe "Uber Technologies Inc." should be here at the end?

polyglot/piranha/src/lib.rs
```diff
@@ -11,9 +11,10 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
+use config::CommandLineArguments;
 use models::{
-  piranha_arguments::PiranhaArguments, piranha_output::PiranhaOutputSummary,
-  source_code_unit::SourceCodeUnit,
+  piranha_arguments::PiranhaArguments,
+  piranha_output::PiranhaOutputSummary, source_code_unit::SourceCodeUnit,
```
lazaroclapp, line 17: Shouldn't these be each in one line? (since they don't all fit in a single line)

polyglot/piranha/src/lib.rs
```diff
@@ -46,11 +47,36 @@ use crate::{
...
 use std::collections::VecDeque;
 use tree_sitter::Node;
 
-/// Executes piranha for the given configuration
-/// Returns (List of updated piranha files, Map of matches found for each file, map of rewrites performed in each file)
+/// Executes piranha for the provided configuration at {path_to_configurations} upon the given {path_to_codebase}. 
+/// 
+/// # Arguments:
+/// * path_to_codebase: Path to the root of the code base that Piranha will update
+/// * path_to_configuration: PAth to the directory that contains - `piranha_arguments.toml`, `rules.toml` and optionally `edges.toml`
```
lazaroclapp, line 58: Why the capitalized `A`?

polyglot/piranha/tests/tests.py
```diff
@@ -0,0 +1,23 @@
+from polyglot_piranha import run_piranha_cli
+from os.path import join, basename
+from os import listdir
+
+
+def test_piranha_rewrite():
```
lazaroclapp, line 6: Do these run without `import unittest` and being part of a Python `unittest.TestCase`? I also don't see anything in `pyproject.toml` specifying to run this file... is it by convention from the name? Have we verified that this fails the build/CI if we include a failing test?

polyglot/piranha/tests/tests.py
```diff
@@ -0,0 +1,23 @@
...
+
+def is_as_expected(path_to_scenario, output_summary):
+    expected_output = join(path_to_scenario, 'expected')
+    is_as_expected = True
+    for file_name in listdir(expected_output):
+        with open(join(expected_output, file_name), 'r') as f:
+            file_content = f.read()
+            expected_content = ''.join(file_content.split())
+            updated_content = [''.join(o.content.split()) for o in output_summary if basename(o.path) == file_name][0]
+            is_as_expected = is_as_expected and (is_as_expected and expected_content == updated_content)
```
lazaroclapp, line 22: Redundant boolean conditional, I think, this should be:

```
is_as_expected = (is_as_expected and expected_content == updated_content)
```

But I think in this case, it might be clearer and simpler to just explicitly do:

```
if expected_content != updated_content:
   return False
```

(since, also, there is no point to continuing checking, and we can get rid of `is_as_expected` entirely, returning `True` if this finishes the loop)

polyglot/piranha/tests/tests.py
ketkarameya: Oh boy! I wondered the same for a lot of time staring at the example repos.

So `pytest` uses name based heuristics to decide which is the test method. [Documentation](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery) 

I have not tried explicitly the failing scenario, i will try that and report here.

polyglot/piranha/tests/tests.py
ketkarameya: Lol yea.

polyglot/piranha/src/lib.rs
ketkarameya: I ll run cargo format as a follow up

polyglot/piranha/tests/tests.py
ketkarameya: yayy! the test fails https://github.com/uber/piranha/runs/8163737912?check_suite_focus=true 
I added a test that asserts `19 == 20`.

ketkarameya: Oh also, i invoke `pytest` in the Polyglot Piranha's CI file.

polyglot/piranha/src/lib.rs
ketkarameya: typo

polyglot/piranha/Cargo.toml
ketkarameya: Will add Uber to author list - [Ref](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field)

lazaroclapp (approved):
LGTM!

PR #220, opened 2022-08-23
Allow user defined cleanups 

ketkarameya:
**Background:** 
There are two types of rules - seed rules and cleanup rules. 
Seed rules are the ones given by the user and we exhaustively apply this rule in the target. 
Cleanup rule is only applied relative to a seed rule. 

**Goal:**
Until now our assumption was that users will only provide "seed rules" . But there are scenarios when a user would want to express a cleanup rule. 

This PR introduces that functionality and adds a new test case for it. 
Test case workflow : 
Convert `ArrayList<>()` to `Collections.SingletonList()` and add `Collections` to import if absent. 
Adding import for `Collections` is the user defined "Cleanup rule".


polyglot/piranha/src/models/rule.rs
```diff
@@ -140,12 +141,13 @@ impl Rule {
...
 
-  /// Adds the rule to a new group - "Feature-flag API cleanup"
-  pub(crate) fn add_to_feature_flag_api_group(&mut self) {
-    let group_name: String = FEATURE_FLAG_API_GROUP.to_string();
-    match self.groups.as_mut() {
-      None => self.groups = Some(vec![group_name]),
-      Some(_groups) => _groups.push(group_name),
+  /// Adds the rule to a new group - "SEED" if applicable.
+  pub(crate) fn add_to_seed_rules_group(&mut self) {
+    if !self.groups().contains(&CLEAN_UP.to_string()) {
```
lazaroclapp, line 146: Should this be checking for `CLEAN_UP` or `SEED`? The way I interpret this is as enforcing exclusion (either the rule is a clean up or a seed rule), but if that's so, then doesn't the logic below cause multiple copies of `SEED.to_string()` to be appended to `_groups` when this method is called multiple times?

Also, is there a notion of an enum in rust we could use here rather than two string constants?

lazaroclapp: p.s. Besides, if groups are exclusive, then why the vector and not just `group: Option<GroupEnum>`? Or is it that we plan to have more non-exclusive groups in the future?

polyglot/piranha/src/config.rs
```diff
@@ -73,7 +73,7 @@ pub(crate) fn read_config_files(
 
   // Label the input-rules as `Feature-flag API cleanup`
```
lazaroclapp, line 74: Comment outdated?

ketkarameya: yes.

polyglot/piranha/src/models/rule.rs
ketkarameya: By default we assume that if a user defines a rule it is a `SEED RULE` . However, if the users marks their rule as adds `Cleanup Rules` to the `groups` of the respective rule, we consider it as `CLEAN_UP`.  We check if a user defined rule's group contains `Cleanup Rules`, if not we add current group to `SEED`. 

Our rules are not mutually exclusive (hence, list of group names). The purpose of groups is to make it easy to define edges between sets of rules. 
For instance, if I want to trigger boolean cleanup after doing some feature flag API cleanup, I would add a edge from "feature flag cleanup" rule to group [`boolean_expression_simplify`](https://github.com/uber/piranha/blob/master/polyglot/piranha/src/cleanup_rules/java/rules.toml#L265) rather than adding edges one by one from "feature flag cleanup" rule to each rule in this group . Depending on the rule, we could have that a rule belongs to two groups where one group is sub-group of other.  

Here I "misused" the groups to implement this functionality, instead of adding another field in `Rule` which tracks wether its cleanup / seed.

lazaroclapp: The logic makes sense to me now. That said, doesn't it make sense for `group` to be a set rather than a vector here?

ketkarameya: IT makes sense for it to be a `set`. But `HashSet` doesn't derive `Default`, `Deserialize` and `Serialize`  :| . I will do some refactoring to ensure that everything is always deduplicated.

ketkarameya: I think the problem is that I have tight coupling between the `toml` serialization/deserialization and my structs.  . I should have had an independent layer for `serde` stuff  #lessonLearned

ketkarameya: Ok I am wrong. HashSet is supported by `serde`.

lazaroclapp (approved):
LGTM!

PR #222, opened 2022-08-25
Remove magic number refactoring

ketkarameya:
It is kind of a "Remove magic number refactoring".  Here I basically replace the hard coded value `4` with a user configurable value that defaults to 4.


lazaroclapp (approved):
Makes sense

PR #224, opened 2022-08-27
Bugfix/Cleanup: Piranha was escaping and unescaping new line characters when not required 

ketkarameya:
Sometimes we need to escape new line characters when it is part of the tree-sitter query. i.e. convert `\n` to `\\n`. 
Problem is that, I was using the method `substitute_tags` for both substituting tags in a replacement pattern and in a tree-sitter query. There was a bug because of this. 
To fix this bug, I basically added a boolean flag indicating wether the input is a tree-sitter query . This flag determines if a  new line character is escaped in the substituted value.

### DIFF
--- a/polyglot/piranha/src/utilities/tree_sitter_utilities.rs
+++ b/polyglot/piranha/src/utilities/tree_sitter_utilities.rs
@@ -314,12 +314,19 @@ fn _get_tree_sitter_edit(
   }
 }
 
-pub(crate) fn substitute_tags(s: String, substitutions: &HashMap<String, String>) -> String {
-  let mut output = s;
+pub(crate) fn substitute_tags(
+  input_string: String, substitutions: &HashMap<String, String>, is_tree_sitter_query: bool,
+) -> String {
+  let mut output = input_string;
   for (tag, substitute) in substitutions {
     // Before replacing the key, it is transformed to a tree-sitter tag by adding `@` as prefix
     let key = format!("@{}", tag);
-    output = output.replace(&key, &substitute.replace('\n', "\\n"))
+    let substitution_value = if is_tree_sitter_query {
+      substitute.replace('\n', "\\n").to_string()
+    } else {
+      substitute.to_string()
+    };
+    output = output.replace(&key, &substitution_value);
   }
   output
 }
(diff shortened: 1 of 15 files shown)


ketkarameya:
I dont see a test for this.  This was just some confusing and convoluted logic that I straightened.

polyglot/piranha/test-resources/java/new_line_character_used_in_string_literal/expected/SomeClass.java
```diff
@@ -0,0 +1,21 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.piranha;
+
+class SomeClass {
+  
+  void someMethod(String s){      
+      assert("Hello \n World".equals(s));
```
lazaroclapp, line 19: Presumably, without the change, this becomes:

```
  void someMethod(String s){      
      assert("Hello \\n World".equals(s));
```

correct?

Or is it:

```
  void someMethod(String s){      
      assert("Hello 
 World".equals(s));
```

?

ketkarameya: yes. It's the later.

polyglot/piranha/src/utilities/tree_sitter_utilities.rs
```diff
@@ -314,6 +314,11 @@ fn _get_tree_sitter_edit(
   }
 }
 
+/// As input, `substitute_tags` requires the (i) `input_string`, (ii) substitutions for the tags (i.e. @<var>) 
+/// and whether the input string is a tree-sitter query (else it is considered as a replacement pattern).
+/// `substitute_tags` replaces the all the occurrences of a specific tag  with the corresponding string values 
+/// specified in `substitutions`, 
+/// Not it escapes newline characters for tree-sitter-queries.
```
lazaroclapp, line 321: `Not` -> `Note`, I assume. Also, isn't there a standard way to document arguments in Rust doc comments? https://doc.rust-lang.org/rust-by-example/meta/doc.html (the example there seems to suggest there is)

ketkarameya: Updated the comment to follow the documentation guidelines.

lazaroclapp (approved):
LGTM!

PR #225, opened 2022-08-29
Bug Fix/ Feature ? Allow performing consecutive scope level rules. 

ketkarameya:
Currently, when we say apply `Rule 1` and then apply `Rule 2` in the class scope of `Rule 1`, we assume that `Rule 1` is applied to a node which is a child node of the `Class` (i.e. `Class` node should be ancestor of the node changed by `Rule 1`). But it could happen that, `Rule 1` changes the class node itself.  
For such case, we now also match the scope query against the changed node itself and its ancestors. 
Earlier we matched scope query only against ancestors.

### DIFF
--- a/polyglot/piranha/test-resources/java/consecutive_scope_level_rules/configurations/rules.toml
+++ b/polyglot/piranha/test-resources/java/consecutive_scope_level_rules/configurations/rules.toml
@@ -1,0 +1,53 @@
+# Copyright (c) 2022 Uber Technologies, Inc.
+# 
+# <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+# except in compliance with the License. You may obtain a copy of the License at
+# <p>http://www.apache.org/licenses/LICENSE-2.0
+# 
+# <p>Unless required by applicable law or agreed to in writing, software distributed under the
+# License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file contains rules to the specific feature flag API.
+
+#
+
+
+[[rules]]
+name = "add_inner_class"
+query = """(
+(class_declaration name: (_)@class_name 
+    body : (class_body ((_)*) @class_members)  @class_body
+ ) @class_declaration
+(#eq? @class_name "FooBar")
+)"""
+replace_node = "class_body"
+replace = """{
+ @class_members
+ public class InnerFooBar {  
+    private String name;
+ }  
+}"""
+[[rules.constraints]]
+matcher = "(class_declaration ) @c_cd"
+queries = ["""(
+(class_declaration name:(_) @name ) @cd
+(#eq? @name "InnerFooBar")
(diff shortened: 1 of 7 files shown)


polyglot/piranha/test-resources/java/consecutive_scope_level_rules/configurations/rules.toml
```diff
@@ -0,0 +1,55 @@
...
+name = "add_field_declaration"
+query = """(
+(class_declaration name: (_)@class_name 
+    body : (class_body ((_)*) @class_members)  @class_body
+ ) @class_declaration
+)"""
+replace_node = "class_body"
+replace = "{\n private String address;\n @class_members \n}"
+groups = ["Cleanup Rule"]
+[[rules.constraints]]
```
lazaroclapp, line 49: Question: What does `rules.constraints` mean? Are they conditions that must hold after applying the transformation specified by the rule? Because, presumably, this second rule `add_field_declaration` doesn't trigger further rules, or does it?

ketkarameya: Yes, you are correct. For some more context - A rules.query allows you to write tree-sitter to match an AST. However, we can only match downwards. 
We often want to match a rule and a specific context (i.e. a specific pattern say we want to check if the enclosing method contains a particularly named local variable declaration or not)

lazaroclapp: Ok, so just to be clear, the way to read this whole block: `[[rules]] + [[rules.constraints]]` is:

> Select AST nodes matching `rules.query`, excluding those that match (any of? / all of?) `rules.constraints.queries` and replace the node identified as `rules.replace_node` with the string content of `rules.replace`.

So `[[rules.constraints]]` is actually an exclude filter. Correct?

ketkarameya: on a side note, this is the weird part of `toml` maybe.  We have the `[[rules.constraints]]` syntax not `constraints = ...` because it is a list of custom struct (i.e .`matcher`/`queries`)

ketkarameya: Oh sorry. Array of tables https://learnxinyminutes.com/docs/toml/ :|

lazaroclapp: If we came up with the name rather than it being something tree-sitter already uses, it might be worth giving it a more semantically meaningful name than `constraints`. If that element is only ever used to exclude filter matches maybe `rules.exclude`? 

That said, definitely not part of this PR! :)

lazaroclapp (approved):
Makes sense to me :)

polyglot/piranha/test-resources/java/consecutive_scope_level_rules/configurations/rules.toml
ketkarameya: yes, I will do that. #229

PR #230, opened 2022-08-31
Refactoring? Improve`delete_file_with_no_type_declarations` rule.

ketkarameya:
Improve the rule.
(i)  Add `annotation_type_declaration` just to make it complete.
(ii) improved names,
(iii) `@package_name` to be caught as an optional tag **(which is the main motivation behind the PR)**
(iv)  Add comments for this rule

### DIFF
--- a/polyglot/piranha/src/cleanup_rules/java/rules.toml
+++ b/polyglot/piranha/src/cleanup_rules/java/rules.toml
@@ -343,15 +343,23 @@ matcher = "(enum_declaration) @ed"
 queries = ["(enum_constant) @ec"]
 
 
+# This rule deltes all the content of the file if it contains 
+# no type declaration, enum declaration and annotation type declaration. 
+#
+# It optionally captures the @package_name for the deleted file. 
 [[rules]]
 name = "delete_file_with_no_type_declarations"
-query = """
-(program) @program"""
+query = """(
+(program 
+  (package_declaration (_) @package_name)?) @program)"""
 replace = ""
 replace_node = "program"
 [[rules.constraints]]
-matcher = "(program) @prgrm"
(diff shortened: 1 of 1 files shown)


polyglot/piranha/src/cleanup_rules/java/rules.toml
```diff
@@ -343,15 +343,23 @@ matcher = "(enum_declaration) @ed"
...
 replace = ""
 replace_node = "program"
 [[rules.constraints]]
-matcher = "(program) @prgrm"
-queries = ["(enum_declaration) @ec", "(class_declaration) @cd", "(interface_declaration) @id"]
+matcher = "(program) @c_program"
+queries = ["(enum_declaration) @c_enum_declaration", 
+            "(class_declaration) @c_class_declaration",
+            "(interface_declaration) @c_interface_declaration",
+            "(annotation_type_declaration) @c_annotation_type_declaration"]
```
lazaroclapp, line 362: Do we know these are all the possible cases?

What about e.g. module declarations: see [JLS 7.3](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.3)

Also, JDK 17 records in the 'near' future.

This change is strictly an improvement, of course, since it adds annotations which we were previously missing, but still... is there a grammar here for all the things that can be found at this AST level in the tree-sitter spec for Java?

ketkarameya: Yes all the possible cases are here - https://github.com/tree-sitter/tree-sitter-java/blob/master/grammar.js#L636. This is the Java grammar `tree-sitter-java` uses.  I think if I am adding annotation type declaration, I might as well add others.

lazaroclapp (approved):
LGTM 🚀

PR #231, opened 2022-09-02
Resolve logging problem

ketkarameya:
So since a long time, our logs were not getting printed properly :| I tried a lot hacks and work arounds but for some reason I am unable to have the logs written to a particular log file with color and has some encoding issue :-|   Now there are some work arounds mentioned in the issues : [1](https://github.com/env-logger-rs/env_logger/issues/208) and [2](https://github.com/env-logger-rs/env_logger/issues/178), but it doesn't work for me :-| . Now we either use some other library (apparently the one we are using is most popular) or we just `cat` when we want logs to be written to a particular file. 

I also cleaned up the logs - 
(i) removed unnecessary logs 
(ii) `println` -> `info`

### DIFF
--- a/polyglot/piranha/src/models/rule_store.rs
+++ b/polyglot/piranha/src/models/rule_store.rs
@@ -14,7 +14,7 @@ Copyright (c) 2022 Uber Technologies, Inc.
 use std::collections::HashMap;
 
 use colored::Colorize;
-use log::{info};
+use log::info;
 use tree_sitter::{Language, Query};
 
 use crate::{
@@ -199,7 +199,7 @@ impl RuleStore {
       global_rules: vec![],
       piranha_args: PiranhaArguments::default(),
       scopes,
-      global_tags : HashMap::new()
+      global_tags: HashMap::new(),
     }
   }
 }
--- a/polyglot/piranha/src/models/rule_store.rs
+++ b/polyglot/piranha/src/models/rule_store.rs
@@ -14,7 +14,7 @@ Copyright (c) 2022 Uber Technologies, Inc.
 use std::collections::HashMap;
 
 use colored::Colorize;
-use log::info;
+use log::{info};
 use tree_sitter::{Language, Query};
 
 use crate::{
--- a/polyglot/piranha/src/tests/mod.rs
+++ b/polyglot/piranha/src/tests/mod.rs
@@ -13,11 +13,13 @@ Copyright (c) 2022 Uber Technologies, Inc.
 
 use std::path::{Path, PathBuf};
 
+use log::{error};
+
 use crate::config::CommandLineArguments;
 use crate::execute_piranha;
 use crate::models::piranha_arguments::PiranhaArguments;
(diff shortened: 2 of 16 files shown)


polyglot/piranha/src/models/rule_store.rs
```diff
@@ -14,7 +14,7 @@ Copyright (c) 2022 Uber Technologies, Inc.
 use std::collections::HashMap;
 
 use colored::Colorize;
-use log::info;
+use log::{info};
```
lazaroclapp, line 17: Needed? I think `log::info` is the right syntax if we are just importing that, no?

polyglot/piranha/src/tests/mod.rs
```diff
@@ -13,11 +13,13 @@ Copyright (c) 2022 Uber Technologies, Inc.
 
 use std::path::{Path, PathBuf};
 
+use log::{error};
```
lazaroclapp, line 16: `use log::error`

polyglot/piranha/src/utilities/tree_sitter_utilities.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
 };
 use colored::Colorize;
 use itertools::Itertools;
-use log::info;
+use log::{info};
```
lazaroclapp, line 22: Am I missing something about `log::{info}` vs `log::info`

lazaroclapp (approved):
Minor nits about imports, but overall I got not problem with this change.

Question: when you say you can't output logs in color to a file, do you mean that if you add calls like `.green()` the logs aren't written at all when sent to a file? Because if you just mean the color is lost, then I'd expect that from the fact that it's a plain test file and not a rich TTY.

polyglot/piranha/src/utilities/tree_sitter_utilities.rs
ketkarameya: No its equivalent. 
I think during my changes I had added a warn macro in this file and made this import -> 
`log::info` ->  log::{info, warn}`. But then I removed `warn`, and `{}` remained. I will clean i up.

ketkarameya:
> Minor nits about imports, but overall I am fine with this change.
> 
> Question: when you say you can't output logs in color to a file, do you mean that if you add calls like `.green()` the logs aren't written at all when sent to a file? Because if you just mean the color is lost, then I'd expect that from the fact that it's a plain test file and not a rich TTY.

Hmmm Its actually that the characters don't get encoded and are shown as blobs :| . It is open issue (https://github.com/env-logger-rs/env_logger/issues/208 and https://github.com/env-logger-rs/env_logger/issues/178) it has something to do with the `pipe` operation.

polyglot/piranha/src/models/rule_store.rs
ketkarameya: YEs. Addressed by cargo format. 

I think I ll add a new Github action that runs cargo format on each PR, as a commit. Idk

PR #246, opened 2022-09-16
Feature/add_pyo3_logging

ketkarameya:
Use the feature that enables streaming the rust logging into python logging. I also took this opportunity to improve our logging.

### DIFF
--- a/polyglot/piranha/src/lib.rs
+++ b/polyglot/piranha/src/lib.rs
@@ -109,9 +109,7 @@ pub fn execute_piranha(
     let number_of_matches = &summary.matches().len();
     info!("File : {:?}", &summary.path());
     info!("{}", format!("# Rewrites : {number_of_rewrites}"));
-    info!("{}", format!("# Rewrites : {number_of_matches}"));
-    info!("# Rewrites : {}", &summary.rewrites().len() );
-    info!("# Matches : {}", &summary.matches().len() );
+    info!("{}", format!("# Matches : {number_of_matches}"));
     total_number_of_rewrites += number_of_rewrites;
     total_number_of_matches += number_of_matches;
   }
--- a/polyglot/piranha/src/lib.rs
+++ b/polyglot/piranha/src/lib.rs
@@ -101,22 +101,25 @@ pub fn execute_piranha(
     .iter()
     .map(PiranhaOutputSummary::new)
     .collect_vec();
-  
-  let mut total_number_of_matches : usize = summaries.iter().map(|x|x.matches().len()).sum();
-  let mut total_number_of_rewrites : usize = summaries.iter().map(|x|x.rewrites().len()).sum();
-  for summary in &summaries {
-    let number_of_rewrites =&summary.rewrites().len();
-    let number_of_matches = &summary.matches().len();
-    info!("File : {:?}", &summary.path());
-    info!("{}", format!("# Rewrites : {number_of_rewrites}"));
-    info!("{}", format!("# Matches : {number_of_matches}"));
-    total_number_of_rewrites += number_of_rewrites;
-    total_number_of_matches += number_of_matches;
-  }
-  info!("{}", format!("Total files affected/matched {}", &summaries.len()));
-  info!("{}", format!("Total number of matches {total_number_of_matches}"));
-  info!("{}", format!("Total number of rewrites {total_number_of_rewrites}"));
+  log_the_summary(&summaries);
   return summaries;
+}
+
+fn log_the_summary(summaries: &Vec<PiranhaOutputSummary>) {
+    let mut total_number_of_matches : usize = summaries.iter().map(|x|x.matches().len()).sum();
+    let mut total_number_of_rewrites : usize = summaries.iter().map(|x|x.rewrites().len()).sum();
+    for summary in summaries {
+      let number_of_rewrites =&summary.rewrites().len();
+      let number_of_matches = &summary.matches().len();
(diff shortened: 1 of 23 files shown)


polyglot/piranha/src/lib.rs
```diff
@@ -93,11 +96,30 @@ pub fn execute_piranha(
...
     .get_updated_files()
     .iter()
     .map(PiranhaOutputSummary::new)
-    .collect_vec()
+    .collect_vec();
+  log_the_summary(&summaries);
+  return summaries;
+}
+
+fn log_the_summary(summaries: &Vec<PiranhaOutputSummary>) {
```
lazaroclapp, line 108: Why `log_the_summary` rather than `log_summaries`? Specially if it's receiving multiple summaries (whatever that means)

polyglot/piranha/src/lib.rs
```diff
@@ -93,11 +96,30 @@ pub fn execute_piranha(
...
+    let mut total_number_of_matches : usize = summaries.iter().map(|x|x.matches().len()).sum();
+    let mut total_number_of_rewrites : usize = summaries.iter().map(|x|x.rewrites().len()).sum();
+    for summary in summaries {
+      let number_of_rewrites =&summary.rewrites().len();
+      let number_of_matches = &summary.matches().len();
+      info!("File : {:?}", &summary.path());
+      info!("{}", format!("# Rewrites : {number_of_rewrites}"));
```
lazaroclapp, line 115: Isn't info already taking a format string? Why the double templating? (Also, shouldn't we be indenting per-file results?)

polyglot/piranha/src/lib.rs
```diff
@@ -93,11 +96,30 @@ pub fn execute_piranha(
...
+      info!("{}", format!("# Matches : {number_of_matches}"));
+      total_number_of_rewrites += number_of_rewrites;
```
lazaroclapp, line 117: Aren't this and the next line double-counting if we already computed `total_number_of_rewrites` as `summaries.iter().map(|x|x.rewrites().len()).sum()` above? I think this ends up printing 2x the total number of rewrites...

polyglot/piranha/src/lib.rs
```diff
@@ -346,7 +379,7 @@ impl FlagCleaner {
     loop {
       let current_rules = self.rule_store.global_rules();
 
-      info!("{}", format!("# Global rules {}", current_rules.len()));
+      debug!("\n{}", format!("# Global rules {}", current_rules.len()));
```
lazaroclapp, line 382: Again, not sure how the API works here... but is it necessary to use `format` rather than just `debug!(...)` itself expanding a format string? Similar question on a few cases below.

polyglot/piranha/src/models/rule_graph.rs
```diff
@@ -15,8 +15,9 @@ use crate::{
   models::{outgoing_edges::OutgoingEdges, rule::Rule},
   utilities::MapOfVec,
 };
-use std::collections::HashMap;
+use std::collections::{HashMap};
```
lazaroclapp, line 18: Remove extra `{` and `}`

polyglot/piranha/src/lib.rs
ketkarameya: `# Rewrites : {number_of_rewrites}"` being able to lookup a variable `number_of_rewrites` in the context to format the string is only possible using the `format!` macro.

polyglot/piranha/src/lib.rs
ketkarameya: Good catch.

polyglot/piranha/src/lib.rs
ketkarameya: Yes, here using the `format!` macro is unnecessary.

polyglot/piranha/src/models/rule_graph.rs
ketkarameya: I will run cargo fmt before landing.

polyglot/piranha/src/lib.rs
ketkarameya: At some places I am doing double formatting because I add color to the logs using the `format!` macro.

polyglot/piranha/src/lib.rs
```diff
@@ -93,11 +96,30 @@ pub fn execute_piranha(
...
+}
+
+fn log_piranha_output_summaries(summaries: &Vec<PiranhaOutputSummary>) {
+    let mut total_number_of_matches : usize = 0;
+    let mut total_number_of_rewrites : usize = 0;
+    for summary in summaries {
+      let number_of_rewrites =&summary.rewrites().len();
+      let number_of_matches = &summary.matches().len();
+      info!("File : {:?}", &summary.path());
+      info!("  {}", format!("# Rewrites : {number_of_rewrites}"));
```
lazaroclapp, line 115: This (and others below) still seem like unnecessary calls to `format!`, unless I am missing something. Also, wouldn't it be better to indent with tabs? (i.e. `info!("\t# Rewrites : {number_of_rewrites}");` )

ketkarameya: Yea. Thats the thing. `info!` does not support formatting strings using values of variables from the context. That is something that only the `format!` marcro supports. 
If we had to do it using info I would do something like: 
`info!(" # Rewrites : {}", number_of_rewrites))`
I will update it to this pattern (in cases where I am not using `color`)

lazaroclapp (approved):

PR #247, opened 2022-09-19
Use include_str! 

ketkarameya:
Since we know the path to the language specific rules/edges/scopes at static time, we can use the `include_str!` feature.

### DIFF
--- a/polyglot/piranha/src/config.rs
+++ b/polyglot/piranha/src/config.rs
@@ -22,10 +22,10 @@ use crate::{
     rule::{Rule, Rules},
     scopes::{ScopeConfig, ScopeGenerator},
   },
-  utilities::read_toml,
+  utilities::{read_toml, parse_toml},
 };
 
-use std::path::{Path, PathBuf};
+use std::path::{Path};
 
 use clap::Parser;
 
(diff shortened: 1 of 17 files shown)


polyglot/piranha/src/config.rs
```diff
@@ -22,10 +22,10 @@ use crate::{
     rule::{Rule, Rules},
     scopes::{ScopeConfig, ScopeGenerator},
   },
-  utilities::read_toml,
+  utilities::{read_toml, parse_toml},
 };
 
-use std::path::{Path, PathBuf};
+use std::path::{Path};
```
lazaroclapp, line 28: `use std::path::Path;

polyglot/piranha/src/utilities/mod.rs
```diff
@@ -92,6 +99,9 @@ pub(crate) fn find_file(input_dir: &PathBuf, name: &str) -> PathBuf {
     .path()
 }
 
+
+
+
```
lazaroclapp, line 104: What are these newlines for?

lazaroclapp (approved):
👍

polyglot/piranha/src/utilities/mod.rs
ketkarameya: I Am going to see how to run cargo fmt each time before the code gets to review.

lazaroclapp: Have you considered setting it up as a git pre-commit hook and adding a CI job to check for it on PRs?

PR #251, opened 2022-09-20
Simplify `delete trailing comma`  logic (bug fix/enhancement)

ketkarameya:
I don't know why I did not implement the _"delete trailing comma"_ feature in this manner before. 
Anyhoo. 

#### Previously 
Our current approach to _delete tailing comma_ is  - uses the tree-sitter's error diagnostics and regex heuristics . Unsurprisingly this is flaky.  For instance, Polyglot Piranha fails to delete the comma when deleting the enum constant `STALE_FLAG` in the below case  : 
```
public enum Something implements ExperimentName {

  /** Show menu */
  SHOW_MENU,

  /** Foobar */
  STALE_FLAG,

  /** barfoo. */
  BAR_FOO,
}
```
Tree-sitter diagnosis does not correctly report the extra comma after deleting `STALE_FLAG` (due to the interleaving `block_comment`s) and our regexes dont match (because our regex are not _comment aware_)
Long story short, we should not have relied on Tree-sitter error diagnostics to begin with. 

#### Current idea: 
Remove all the above stuff. 

Implement a very simple algorithm : 
* Identify when the edit is a _Delete operation_
* Look up the node **after** (Like the closest node that begins after the end of this _to be deleted node_ ) this _to be deleted node_.
* If this node is a `,` extend the deletion range to include this node. 

All the current test cases related to deleting commas pass, plus i added this new test scenario. 

P.S.
This significantly simplifies the implementation too.

### DIFF
--- a/polyglot/piranha/src/models/source_code_unit.rs
+++ b/polyglot/piranha/src/models/source_code_unit.rs
@@ -207,13 +207,14 @@ impl SourceCodeUnit {
     let mut new_deleted_range = deleted_range;
 
     // Get the node immediately after the to-be-deleted code
-    if let Some(node_after_to_be_deleted_node) = self
+    if let Some(parent_node) = self
       .ast
       .root_node()
-      .descendant_for_byte_range(deleted_range.end_byte, deleted_range.end_byte + 5)
+      .descendant_for_byte_range(deleted_range.end_byte, deleted_range.end_byte + 1)
+      .and_then(|n|n.parent())
     {
-      // Traverse this `node_after_to_be_deleted_node` to find the closest next node after the `replace_range`
-      if let Some(node) = traverse(node_after_to_be_deleted_node.walk(), Order::Post)
+      // Traverse this `parent_node` to find the closest next node after the `replace_range`
+      if let Some(node_after_to_be_deleted_node) = traverse(parent_node.walk(), Order::Post)
         .filter(|n| n.start_byte() >= deleted_range.end_byte)
         .min_by(|a, b| {
           (a.start_byte() - deleted_range.end_byte).cmp(&(b.start_byte() - deleted_range.end_byte))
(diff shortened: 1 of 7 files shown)


polyglot/piranha/test-resources/java/feature_flag_system_1/treated/input/XPFlagCleanerPositiveCases.java
```diff
@@ -34,6 +34,16 @@ enum AnotherTestExperimentName {
     SOME_OTHER_FLAG
   }
 
+  public enum Something implements ExperimentName {
+    /** Show menu */
+    SHOW_MENU,
+  
+    /** Foobar */
```
lazaroclapp, line 41: Shouldn't the treated **input** still have an XP flag enum here?

polyglot/piranha/src/models/source_code_unit.rs
```diff
@@ -174,21 +167,74 @@ impl SourceCodeUnit {
...
+  /// Traverse this node and get the node closest to the range {deleted_range}'s end byte
+  /// IF this closest node is a comma, extend the {new_delete_range} to include the comma.
+  fn delete_trailing_comma(&mut self, deleted_range: Range) -> Range {
+    let mut new_deleted_range = deleted_range;
+
+    // Get the node immediately after the to-be-deleted code
+    if let Some(node_after_to_be_deleted_node) = self
+      .ast
+      .root_node()
+      .descendant_for_byte_range(deleted_range.end_byte, deleted_range.end_byte + 5)
```
lazaroclapp, line 213: What does `+5` mean here? Is there a way to get the next token here, rather than trying to guess the byte indexes? What happens with:
```
STALE_FLAG              ,
```
? (GJF would get rid of that, but I think it's still valid Java)

lazaroclapp (commented):
Seems like a definite improvement over the previous approach, one question and one possible error on the tests to check before landing, though.

polyglot/piranha/test-resources/java/feature_flag_system_1/treated/input/XPFlagCleanerPositiveCases.java
ketkarameya: Done. Thanks good catch.

lazaroclapp (approved):

PR #253, opened 2022-09-22
Selectively triggering GitHub workflows based on directory changes

dvmarcilio:
## Context

Currently, every pull request is triggering all GitHub workflows, regardless of where the changes are made. For instance, all the recent Polyglot's PRs are executing unrelated CI jobs for Java and JavaScript.

This PR uses https://github.com/dorny/paths-filter to conditionally execute GitHub workflows based on changes made to specific directories.
paths-filter seems stable and is used by established projects/organizations (e.g., [GoogleChrome](https://github.com/GoogleChrome/web.dev/blob/e1f0c28964e99ce6a996c1e3fd3ee1985a7a04f6/.github/workflows/lint-and-test-workflow.yml#L33), [Roblox](https://github.com/Roblox-ActionsCache/dorny-paths-filter), [sentry.io](https://github.com/getsentry/sentry/blob/ca0e43dc5602a9ab2e06d3f6397cc48fb5a78541/.github/workflows/backend-test-py3.6.yml#L32)).

GitHub has a feature to skip workflow executions with path filtering, [but it leaves the checks as "Pending"](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs). AFAIK, Piranha requires all checks to pass for pull request approval.
With this PR, skipped jobs will still appear as passing.

## Details

The filters globs are defined in `.github/filters.yaml` and reused across the workflows `.yaml` files.

I have three PRS on my fork testing the conditional execution:

- https://github.com/dvmarcilio/piranha/pull/1 targeting only Java CI.
- https://github.com/dvmarcilio/piranha/pull/2 aiming at only JS CI.
- https://github.com/dvmarcilio/piranha/pull/3 triggering Java, JS, and Polyglot.

During my tests, I noticed that the filter has to be declared after `actions/checkout`. It cannot figure out the changes before it.

Finally, I did some minor workflows/jobs renaming. I added the `JS` prefix for both `Lint` and `Format` to improve clarity.

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -1,0 +1,6 @@
+js:
+  - javascript/**
+java:
+  - java/**
+polyglot:
+  - polyglot/**
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -23,11 +23,19 @@ jobs:
         node-version: [10.x, 12.x, 14.x]
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
-      with:
-        node-version: ${{ matrix.node-version }}
-    - run: npm ci
-    - run: npm test
-    - run: npm test
+      - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: '.github/filters.yaml'
+      - name: Use Node.js ${{ matrix.node-version }}
+        if: ${{ steps.filter.outputs.js == 'true' }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - if: ${{ steps.filter.outputs.js == 'true' }}
+        run: npm ci
+      - if: ${{ steps.filter.outputs.js == 'true' }}
+        run: npm test
+      - if: ${{ steps.filter.outputs.js == 'true' }}
+        run: npm test
--- a/.github/workflows/polyglot_build.yml
+++ b/.github/workflows/polyglot_build.yml
@@ -8,37 +8,48 @@ env:
   CARGO_TERM_COLOR: always
 jobs:
   build_and_test:
-    name: Piranha
+    name: Polyglot Piranha Build
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Piranha sources
         uses: actions/checkout@v2
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: '.github/filters.yaml'
       - name: Install stable toolchain
         uses: actions-rs/toolchain@v1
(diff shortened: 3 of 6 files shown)


.github/filters.yaml
```diff
@@ -0,0 +1,6 @@
+js:
```
ketkarameya, line 1: what about objc and swift?

dvmarcilio: I only added filters for workflows in https://github.com/uber/piranha/tree/master/.github/workflows
It doesn't seem that we have CI for objc.

.github/workflows/polyglot_build.yml
```diff
@@ -8,37 +8,48 @@ env:
   CARGO_TERM_COLOR: always
 jobs:
   build_and_test:
-    name: Piranha
+    name: Polyglot Piranha Build
```
ketkarameya, line 11: Can we run all the steps when `if: ${{ steps.filter.outputs.polyglot == 'true' }}` instead of adding this condition to each step?

.github/workflows/node.js.yml
```diff
@@ -23,11 +23,19 @@ jobs:
         node-version: [10.x, 12.x, 14.x]
```
ketkarameya, line 24: same question here!

ketkarameya: Can we have the selective running for PRs . 
And run all the jobs when we push into main ?

ketkarameya (commented):
Did you test this after the changes ? like u earlier tested :
https://github.com/dvmarcilio/piranha/pull/1 targeting only Java CI.
https://github.com/dvmarcilio/piranha/pull/2 aiming at only JS CI.
https://github.com/dvmarcilio/piranha/pull/3 triggering Java, JS, and Polyglot.

dvmarcilio:
I followed your suggestions which made the overall changes much cleaner IMO. Thanks!

- reduced the duplication across files by having a Reusable Workflow in `detect_changes.yml`.
- added a README file summarizing the filtering.
- slightly renamed a few jobs and workflows names

Tested in the following PRS: 

https://github.com/dvmarcilio/piranha/pull/8 all CI
https://github.com/dvmarcilio/piranha/pull/7 only Polyglot CI
https://github.com/dvmarcilio/piranha/pull/6 only JS CI
https://github.com/dvmarcilio/piranha/pull/5 onyl Java CI

ketkarameya (approved):
LGTM.

ketkarameya (approved):
LGTM. The updates look like what we discussed online.

dvmarcilio:
I added `skip_` workflows that are executed when a regular workflow is skipped; they allow required checks to pass.
This is suggested by [GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#example)).
I described them in the README.

PR #256, opened 2022-09-27
Onboarding Go

dvmarcilio:
This PR adds the Go grammar as a dependency and a match only demo.

## Demo

The demo has two simple match rules:

- rule`find_for`: matches  any `for_statement`
- rule `find_go_stmt_for_loop`: matches a `go_stmt` inside the block of a `for_statement`

For the code below (reduced version of the demo), the rules will produce the following matches:

- `find_for`: 3
- `find_go_stmt_for_loop`: 1

```go
func f(n int) {
	for i := 0; i < 10; i++ {
		fmt.Println(n, ":", i)
	}
}

func main() {
	go f(0) // go_statement does not get matched
	go f(1) // go_statement does not get matched
}

func go_stmt() {
	for i := 0; i < 10; i++ {
		go f(i) // go_statement gets matched
	}
	var input string
	fmt.Scanln(&input)
}

func f3() {
	var pow = []int{1, 2, 4, 8, 16, 32, 64, 128}
	for i, v := range pow {
		fmt.Printf("2**%d = %d\n", i, v)
	}
}
```

### DIFF
--- a/polyglot/piranha/demo/match_only/go/configurations/rules.toml
+++ b/polyglot/piranha/demo/match_only/go/configurations/rules.toml
@@ -13,15 +13,12 @@
 name = "find_go_stmt_for_loop"
 query = """
 (
-    (for_statement
-        (block
-            (_)*
-            (go_statement) @go_stmt
-            (_)*
-        ) @block
-    ) @for_stmt
+    (go_statement) @go_stmt
 )
 """
+[[rules.constraints]]
+matcher = "(for_statement) @fs"
+queries=[]
 
 [[rules]]
 name = "find_for"
--- a/polyglot/piranha/demo/match_only/go/configurations/rules.toml
+++ b/polyglot/piranha/demo/match_only/go/configurations/rules.toml
@@ -1,0 +1,32 @@
+# Copyright (c) 2022 Uber Technologies, Inc.
+#
+# <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+# except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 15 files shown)


polyglot/piranha/demo/match_only/go/configurations/rules.toml
```diff
@@ -0,0 +1,32 @@
...
+# <p>http://www.apache.org/licenses/LICENSE-2.0
+#
+# <p>Unless required by applicable law or agreed to in writing, software distributed under the
+# License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+[[rules]]
+name = "find_go_stmt_for_loop"
+query = """
```
ketkarameya, line 14: Can we use a constraint here? 

like query =  `(go_statement) @go_stmt`

[[rules.constraints]]
matcher = "(for_statement) @fs"
queries=[]

ketkarameya: it will make the rule easier and we dont have to use quantifiers (Which we know dont always work well)

ketkarameya (approved):
LGTM. Small nit about the rule.

polyglot/piranha/demo/match_only/go/configurations/rules.toml
dvmarcilio: Great suggestion. Thanks. I didn't think of an empty query.

PR #257, opened 2022-09-27
Polyglot Java demo: find barFoo in *non* static method

dvmarcilio:
This PR renames for clarity a rule in the match only Java demo.
The rule _excludes_ (through `rules.constraints`) occurrences of `barFoo` that occur inside `static`methods.


ketkarameya (approved):
Good Catch. LGTM.

PR #262, opened 2022-11-09
Py script to generate a .dot graph of rules/groups and their edges

dvmarcilio:
The script generates a .dot file used to generate a graph (.dot) for Polyglot's rules and edges through [graphviz](https://graphviz.org/).

I added a graph image to the README of the pre-built Java cleanup rules.
I also added a section with instructions on how to run the script and its requirements.


polyglot/piranha/rules_dot_graph.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # rules.toml should exist, deliberately fail otherwise
+    rules_toml_dict = toml.load(rules_file)
+
+    for rule_toml in rules_toml_dict['rules']:
+        rule_name: str = sanitize_name(rule_toml['name'])
+        if 'groups' in rule_toml:
+            for group_name in rule_toml['groups']:
+                group_name = sanitize_name(group_name)
+                if group_name != 'Cleanup_Rule':
```
ketkarameya, line 73: i think there is no `_`

dvmarcilio: You're right. In `rules.toml`, and probably in the rust code, it's `Cleanup Rule`.
For Graphviz, we are converting all blank spaces to `_`. The conversion makes our lives easier when defining nodes and edges. I'm not sure Graphviz plays well with spaces in names.

https://github.com/uber/piranha/pull/262/files/5b26d2d4c3dae196cf76376509182a11e7e9cd1f#diff-62d9897c4603bf7c92c8ea08ac5a69b1cb508aa8f9cf262bac0e3d83649fc425R48-R51

ketkarameya: I see. I see. Could be a good naming convention to add .

polyglot/piranha/visualize_rules_graph.py
```diff
@@ -0,0 +1,234 @@
...
+        for to_node in to_nodes:
+            to_node = sanitize_name(to_node)
+            edge = Edge(to=to_node, scope=scope)
+            if from_node in outgoing_edges_by_node:
+                outgoing_edges_by_node[from_node].append(edge)
+            else:
+                outgoing_edges_by_node[from_node] = [edge]
+
+
+def group_rule_names_label(rule_names: 'list[str]') -> 'list[str]':
```
ketkarameya, line 91: What are we grouping here? 
When you say `group_<something>`  I guess the function should return a `dict` or a `list of list`.

polyglot/piranha/visualize_rules_graph.py
```diff
@@ -0,0 +1,234 @@
...
+    return [inline_name(rule_name) for rule_name in rule_names]
+
+
+def inline_name(rule_name: str) -> str:
+    """Should be called for rules under a group."""
+    if rule_name in cleanup_rules:
+        return f'{rule_name} (Cleanup Rule)'
```
ketkarameya, line 98: Why are we appending `(Cleanup Rule)` ?

polyglot/piranha/visualize_rules_graph.py
ketkarameya: I guess this is the _"group"_ from our `Rule`. :|

polyglot/piranha/visualize_rules_graph.py
```diff
@@ -0,0 +1,234 @@
...
+        They get an additional label `(Cleanup_Rule)`.
+
+    Nodes without a group are added to a different set.
+    """
+    for rule_toml in rules_toml_dict['rules']:
+        rule_name: str = sanitize_name(rule_toml['name'])
+        if 'groups' in rule_toml:
+            rule_groups = rule_toml['groups']
+            collect_node_for_group_rule(rule_name, rule_groups)
+        else:  # rule without group
```
ketkarameya, line 50: why would node without a `group` be orphan node? Shouldn't it be when it is a node without an edge?

polyglot/piranha/visualize_rules_graph.py
```diff
@@ -0,0 +1,234 @@
...
+        if rule_name in rules_by_group_dict:
+            generate_node_for_group_rule(rule_name)
+        else:  # a rule not in a group
+            generate_node_for_non_group_rule(rule_name)
+
+        added_nodes.add(rule_name)
+
+
+def all_non_orphan_rule_names() -> 'list[str]':
+    node_names_with_only_outgoing_edges = [
```
ketkarameya, line 130: is this the [set difference](https://www.geeksforgeeks.org/python-set-difference/) between the keys of two dicts?

polyglot/piranha/visualize_rules_graph.py
```diff
@@ -0,0 +1,234 @@
...
+    if rule_name not in added_nodes:
+        if rule_name in cleanup_rules:
+            node_label = f'{rule_name}\\n(Cleanup Rule)'
+            graph.node(rule_name, node_label)
+        else:
+            graph.node(rule_name)
+
+
+def generate_orphan_nodes():
+    # orphan_nodes don't have any group, not even 'Cleanup_Rule'
```
ketkarameya, line 162: Same comment as above. Are we using the term `orphan node` in a non standard way?

ketkarameya (changes requested):
Left a few comments.

polyglot/piranha/visualize_rules_graph.py
dvmarcilio: Nice catch. I changed it to use the set difference operator

polyglot/piranha/visualize_rules_graph.py
dvmarcilio: I changed the name to make it more clear.

dvmarcilio:
Thanks for the careful review. It made me realize that the script was not that clear. 
I changed some names and added more comments. It should be better now.

polyglot/piranha/visualize_rules_graph.py
```diff
@@ -0,0 +1,257 @@
...
+
+def collect_node_for_rule_with_group(rule_name: str, rule_groups: 'list[str]'):
+    for group_name in rule_groups:
+        group_name = sanitize_name(group_name)
+        if group_name != 'Cleanup_Rule':
+            if group_name in rules_by_group_dict:
+                rules_by_group_dict[group_name].append(rule_name)
+            else:
+                rules_by_group_dict[group_name] = [rule_name]
+        else:  # we don't want to group 'Cleanup Rule's under the same graphviz shape
```
ketkarameya, line 62:
```suggestion
        else:  # we don't want to group `Cleanup Rule`s under the same graphviz shape
```

ketkarameya (approved):
small nit. please merge the PR after addressing it.

PR #269, opened 2022-11-16
Fix the disjunction rule

ketkarameya:
The disjunction rules for `<something> || true` earlier returned `true`. This is spurious specially when `<something>` has side-effects 😕 . This PR fixes that.


polyglot/piranha/test-resources/java/feature_flag_system_1/control/expected/XPFlagCleanerPositiveCases.java
```diff
@@ -68,10 +68,14 @@ public void conditional_with_else_contains_stale_flag() {
     System.out.println("Hi world");
   }
 
-  public void complex_conditional_contains_stale_flag(boolean tBool) {
+    public void complex_conditional_contains_stale_flag(boolean tBool) {
```
lazaroclapp, line 71: Formatting seems off for this example. There is an extra space before `public`, and the if is not indented. Not exactly a major issue, but is there any reason this can't be more standardly formatted? 🙂

p.s. Same in other places below. Yet I checked these files and sampling a bit at random, the rest of the Java code looks properly indented.

lazaroclapp (approved):
LGTM. Minor nit about code formatting in the test cases, which you might want to change before landing, but stamped either way.

polyglot/piranha/test-resources/java/non_seed_user_rule/expected/SomeClass.java
```diff
@@ -12,9 +12,9 @@
  * limitations under the License.
  */
 package com.uber.piranha;
-
-import java.util.ArrayList;
 import java.util.Collections;
+import java.util.ArrayList;
```
lazaroclapp, line 16: Any reason we need these imports in non-alphabetical order (not blocking at all, just curious)

lazaroclapp (commented):
Was perfectly fine to land, but one minor nit/question looking at it now 🙂

polyglot/piranha/test-resources/java/non_seed_user_rule/expected/SomeClass.java
ketkarameya: So the way I have the rules, a new import statement is added at the head of the list. 
GJF reorders it, which is fine but then our test cases wont pass. 
Since our testing logic is not immune to reordering imports. 
Like if output produced by piranha is : 
```
import java.util.Collections;
import java.util.ArrayList;
```

and expected output is 
```
import java.util.ArrayList;
import java.util.Collections;
```
our test case will fail because our testing logic to account for the fact that order of import does not matter.

lazaroclapp: Ok, makes sense. A bit unfortunate, since it means we can't just keep running GJF automatically on these files, but not worth the effort involved in addressing it right now, I think.

PR #275, opened 2022-11-22
Use objects instead of tuples for top-level matches

gregintuitaio:
Since tuples are hard for deserialization for some libraries (namely fp-ts), I changed the `(String, Match)` tuple into the `MatchDto` structure.

@ketkarameya I hope you like the change, but I don't know what your plans for deserialization are.

### DIFF
--- a/polyglot/piranha/demo/match_only_demos.py
+++ b/polyglot/piranha/demo/match_only_demos.py
@@ -10,7 +10,7 @@ def java_demo():
     info("Running the Match-only demo for Java")
     output_summary_java = run_piranha_cli(join(match_only_dir, "java"), join(match_only_dir, "java/configurations"), False)
 
-    rule_match_counter = Counter([m[0] for m in output_summary_java[0].matches])
+    rule_match_counter = Counter([m.name for m in output_summary_java[0].matches])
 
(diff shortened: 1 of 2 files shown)


ketkarameya:
Thank you for this contribution. I think what you suggest totally makes sense to me, rather this is how it should have been done in the first place :| I got lazy for some reason (or I didn't notice). 
Anyhoo, this PR will break a lot of our internal tooling (which is okay with me but I need to think about it). 

Just curious, are you not using the Python or Rust bindings?

ketkarameya (commented):
I wonder how complex it is to have "rule name" as part of the `Match` object.

gregintuitaio:
@ketkarameya I wanted to get back to you earlier, but I was quite busy. I don't use Rust/Python bindings ATM, I call the executable and inspect the output file using `fp-ts` / `io-ts` library duo in JS/TS

ketkarameya:
Interesting. 
What if you could call the Rust functions from javascript using _wasm bindgen_ ?
https://blog.frankel.ch/start-rust/5/#:~:text=To%20call%20Rust%20from%20JavaScript,you%20want%20to%20make%20available.

I am not an expert at this library, but i guess it is somewhat similar to how I am using `pyo3` to expose it as a Python library.

PR #283, opened 2022-12-01
Handle the configuration chaos

ketkarameya:
I know this is a slightly large PR :| But it solves two mangled problems related to the way we handle our configurations.

##### Problem 1
There are two ways to use Piranha - as a binary (i.e. command line) and as a library (rust/python). 
Currently our library has only one function (which is equivalent to our command line options).
To execute Piranha we provide it multiple command line options like - `path_to_codebase`, `path_to_configurations`, `path_to_output` and `dry_run`. Now `path_to_configurations` itself contains `piranha_arguments.toml`. This file contains more options to specify piranha's behavior like  cleanup comments, delete an empty file, and so on. 
All of this causes some chaos :| 

Internally, we had a struct [`CommandLineArguments`](https://github.com/uber/piranha/blob/master/polyglot/piranha/src/config.rs#L35) to capture the command line arguments. We also re-used this struct within the only function that our library provides. 
We had another struct [`PiranhaConfiguration` ](https://github.com/uber/piranha/blob/master/polyglot/piranha/src/models/piranha_config.rs#L18) to capture the `piranha_arguments.toml`. 
We then combined the contents of these two structs `CommandLineArguments` and `PiranhaConfiguration` into a larger struct [`PiranhaArguments`](https://github.com/uber/piranha/blob/master/polyglot/piranha/src/models/piranha_arguments.rs#L31) and used `PiranhaArguments` as our oracle of configurations. 
Unsurprisingly, it causes chaos especially when : 
1. Handling default values 
2. scenario where we want to expose the option via command line and `piranha_arguments.toml`
3. Testing required constructing the objects for these three structs 

##### Solution

We just removed `PiranhaConfigurations` and `CommandLineArguments` :)  We use the power of Rust macros to cleanly  repurpose `PiranhaArguments` for parsing command line and `piranha_arguments.toml`. 
We can choose what properties need to be part of `toml_serde`  and `clap` (command line parsing library).
This also allows us to easily expose flags/features via command line, api or `piranha_arguments.toml`

We first read the `PiranhaArguments` from the toml file and then "merge"  it with the `PiranhaArguments` we parsed from command line, or testing specs or Python API .  

I also introduced `PiranhaInput`  to encode command line vs library usage of piranha (as a enum type), this significantly streamlines our implementation.
We move all defaults to the `deault.rs` module, and use the same default values for commandline argument parsing, toml parsing and builder pattern (used in testing). 

##### Problem 2
[...]

PR #289, opened 2022-12-07
Simplify the rules

ketkarameya:
This PR does the following : 
1. convert Rules to an enum (Rewrite, MatchOnly, Dummy )
2. Convert fields of type `Option<HashSet<...>>` -> `HashSet<...>`
3. Add `serde default` to fields of type  HashSet/Vec so that folks **don't have to**  define empty lists. 

Real changes for this PR only in `rules.rs`.

#### Why am I doing all of this ? 

I feel in our rapid prototyping days we made choices :| Anyhoo...

We support three kinds of rules - Rewrite Rules, Match Only rules, Dummy Rules (helpers for recurssion)
We also eventually plan to split Rewrite rules into insert/update/delete. 

Currently we just populate the rules object (via deserialization) and check the "type" of rule by evaluating what fields are empty. Now in our core algorithm, we have control paths for each kind of rule (currently guarded under `if-else` conditions). 
This is getting cumbersome, because it gets confusion to track code path of each type of rule (and makes it easier to introduce bugs), and makes it intimidating to add new features. I eventually want to make all of this more typed at rust level. 

Currently : 
```
struct  Rule {
  name: String,
  query: String, 
   ..
}

impl Rule {
    fn isMatchOnly(){ .. }
    fn isRewrite(){ .. }
    ...
}
fn core_algo(rule: Rule){ 
   ..
   if rule.isMatcOnly() {
    ..
   }
   if rule.isRewrite() {
    ..
   }
 }

fn core_algo_helpers(rule: Rule){ .. }
```
Things to note here:
1. We have one single Rule struct
2. At some abstract level the core algorithm accepts a `Rule`, and all its helpers also accept rule.  In each of these functions we have used `if else` to check the kind of rule. Using `if-else`  is voluntary, so it is easy to have unhandled scenarios. 

Now : 

```
enum  Rule {
  Rewrite{
     name: String,
     query: String, 
     replace: String, 
     ..
   }
 MatchOnly {
     name: String,
     query: String, 
     ..
   }
   ..
}

impl Rule {
    ...
}

fn core_algo(rule: Rule){
   
    match rule {
       Rewrite (..) =>
       MatcOnly (..) =>
    }

 }

fn core_algo_helpers(rule: Rule){ .. }
```
Benefits of this: 
1. Everywhere within `core_algo` if we use Rule we (as in I am) are forced to reason about each category of `Rule`. 
-----------
**Eventually  (in follow up PRs)**  I am planning to do something like below:
```
struct Rewrite{
     name: String,
     query: String, 
     replace: String, 
     ..
   }
struct  MatchOnly {
     name: String,
     query: String, 
     ..
   }
...
enum  Rule {
  Rewrite(Rewrite),
 MatchOnly (MatchOnly)
   ..
}
// all rules have a name 
trait PiranhaRule {
      fn name();
}

trait PiranhaMatcher: PiranhaRule {
    fn query(); 
    fn constraints(); 
    fn holes(); 
    ..
}

// All rewrite rules have a matcher 
trait PiranhaRewrite : PiranhaMatcher {
     fn replace();
}  

impl Rule {
    ...
}

fn core_algo<T: PiranhaRule>(rule: T){ .. }

fn core_algo_helpers1<T: PiranhaMatcher>(rule: T){ .. }
fn core_algo_helpers2<T: PiranhaRewrite>(rule: T){ .. }
```

Benefits of doing this: 
1. Reduces awkwardness introduced by repeated pattern matching 
2. Makes code more reusable (because now we will have functions accepting parameters bound by our traits)
3. Will make it easy when we plan to introduce the new operators

### DIFF
--- a/polyglot/piranha/demo/match_only/go/configurations/piranha_arguments.toml
+++ b/polyglot/piranha/demo/match_only/go/configurations/piranha_arguments.toml
@@ -10,4 +10,4 @@
 # limitations under the License.
 
 language = ["go"]
-substitutions = []
+
--- a/polyglot/piranha/demo/match_only/tsx/configurations/rules.toml
+++ b/polyglot/piranha/demo/match_only/tsx/configurations/rules.toml
@@ -36,7 +36,7 @@ matcher = """
     ) @j
 )
 """
-queries = []
+
 
 [[rules]]
 name = "find_props_identifiers_within_variable_declarators_not_within_divs"
--- a/polyglot/piranha/src/models/language.rs
+++ b/polyglot/piranha/src/models/language.rs
@@ -5,7 +5,7 @@ use tree_sitter::{Parser, Query};
 use crate::utilities::parse_toml;
 
 use super::{
-  default_configs::{default_language, GO, JAVA, KOTLIN, PYTHON, STRINGS, SWIFT, TSX, TYPESCRIPT},
+  default_configs::{GO, JAVA, KOTLIN, PYTHON, STRINGS, SWIFT, TSX, TYPESCRIPT},
   outgoing_edges::Edges,
   rule::Rules,
   scopes::{ScopeConfig, ScopeGenerator},
@@ -75,7 +75,7 @@ impl PiranhaLanguage {
 
 impl Default for PiranhaLanguage {
   fn default() -> Self {
-    PiranhaLanguage::from(default_language().as_str())
+    PiranhaLanguage::from(JAVA)
   }
 }
 
--- a/polyglot/piranha/src/models/rule.rs
+++ b/polyglot/piranha/src/models/rule.rs
@@ -17,7 +17,6 @@ use std::{
 };
 
 use colored::Colorize;
-use itertools::Itertools;
 use serde_derive::Deserialize;
 
 use crate::utilities::{tree_sitter_utilities::substitute_tags, MapOfVec};
@@ -35,43 +34,29 @@ pub(crate) struct Rules {
 
 #[derive(Deserialize, Debug, Clone)]
 #[serde(untagged)]
+#[doc = ""]
 pub(crate) enum Rule {
   Rewrite {
-    /// Name of the rule. (It is unique)
-    name: String,
-    /// Tree-sitter query as string
-    query: String,
-    /// The tag corresponding to the node to be replaced
-    replace_node: String,
-    /// Replacement pattern
-    replace: String,
-    /// Group(s) to which the rule belongs
-    groups: Option<HashSet<String>>,
-    /// Holes that need to be filled, in order to instantiate a rule
-    holes: Option<HashSet<String>>,
-    /// Additional constraints for matching the rule
-    constraints: Option<HashSet<Constraint>>,
-    /// Heuristics for identifying potential files containing occurrence of the rule.
-    grep_heuristics: Option<HashSet<String>>,
+    name: String,                             // Name of the rule. (It is unique)
+    query: String,                            // Tree-sitter query as string
+    replace_node: String,                     // The tag corresponding to the node to be replaced
+    replace: String,                          // Replacement pattern
+    groups: Option<HashSet<String>>,          // Group(s) to which the rule belongs
+    holes: Option<HashSet<String>>, // Holes that need to be filled, in order to instantiate a rule
+    constraints: Option<HashSet<Constraint>>, // Additional constraints for matching the rule
+    grep_heuristics: Option<HashSet<String>>, // Heuristics for identifying potential files containing occurrence of the rule.
   },
 
   MatchOnly {
-    /// Name of the rule. (It is unique)
     name: String,
-    /// Tree-sitter query as string
     query: String,
-    /// Group(s) to which the rule belongs
     groups: Option<HashSet<String>>,
-    /// Holes that need to be filled, in order to instantiate a rule
     holes: Option<HashSet<String>>,
-    /// Additional constraints for matching the rule
     constraints: Option<HashSet<Constraint>>,
-    /// Heuristics for identifying potential files containing occurrence of the rule.
     grep_heuristics: Option<HashSet<String>>,
   },
 
   Dummy {
-    /// Name of the rule. (It is unique)
     name: String,
   },
 }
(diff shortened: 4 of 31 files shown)


polyglot/piranha/demo/match_only/go/configurations/piranha_arguments.toml
```diff
@@ -10,4 +10,4 @@
 # limitations under the License.
 
 language = ["go"]
-substitutions = []
```
ketkarameya, line 13: no need of passing empty list now. (since we use defaults during `serde`)

polyglot/piranha/demo/match_only/tsx/configurations/rules.toml
```diff
@@ -36,7 +36,7 @@ matcher = """
     ) @j
 )
 """
-queries = []
```
ketkarameya, line 39: no need of passing empty list now. (since we use defaults during `serde`)

polyglot/piranha/src/models/language.rs
```diff
@@ -75,26 +74,30 @@ impl PiranhaLanguage {
 
 impl Default for PiranhaLanguage {
   fn default() -> Self {
-    let rules: Rules = parse_toml(include_str!("../cleanup_rules/java/rules.toml"));
```
ketkarameya, line 78: Addressing comment from previous PR (from yesterday #283)

polyglot/piranha/src/models/rule_graph.rs
```diff
@@ -70,6 +70,22 @@ impl RuleGraph {
     }
     (self.0.len(), edges)
   }
+
+  /// Groups the rules based on the field `rule.groups`
+  /// Note: a rule can belong to more than one group.
+  pub(crate) fn group_rules(
```
ketkarameya, line 76: moved from `rule.rs`

polyglot/piranha/src/models/rule.rs
```diff
@@ -29,39 +32,103 @@ pub(crate) struct Rules {
   pub(crate) rules: Vec<Rule>,
 }
 
-#[derive(Deserialize, Debug, Clone, Default, PartialEq)]
-pub(crate) struct Rule {
```
ketkarameya, line 33: convert Rule Struct to enum

polyglot/piranha/src/models/rule.rs
```diff
@@ -73,147 +140,148 @@ impl Rule {
...
-    for hole in self.holes() {
-      if let Some(x) = substitutions.get(&hole) {
-        gh.insert(x.clone());
-      }
+  fn apply_substitutions(&self, substitutions: &HashMap<String, String>) -> Rule {
+    let update = |value: &String, is_tree_sitter_query: bool| {
+      substitute_tags(value.to_string(), substitutions, is_tree_sitter_query)
+    };
+
+    let gh = HashSet::from_iter(
```
ketkarameya, line 178: The logic of `add_grep_heuristics_for_global_rules` has been moved here.

polyglot/piranha/src/models/rule.rs
```diff
@@ -73,147 +140,148 @@ impl Rule {
...
+        .map(|x| x.to_string()),
+    );
+
+    match &self {
+      r @ Rule::Rewrite {
+        query,
+        replace,
+        replace_node,
+        ..
+      } => Rule::Rewrite {
```
ketkarameya, line 192: Lol I know this looks funny. Functional Record Update Syntax doesnt work for enums, it only works for struct , thus this redundancy.

polyglot/piranha/src/models/language.rs
```diff
@@ -75,26 +74,30 @@ impl PiranhaLanguage {
...
-      comment_nodes: vec!["line_comment".to_string(), "block_comment".to_string()],
-    }
+    PiranhaLanguage::from("java")
   }
 }
 
 impl From<&str> for PiranhaLanguage {
   fn from(language: &str) -> Self {
     match language {
-      "java" => PiranhaLanguage::default(),
```
ketkarameya, line 97: addressing your comment from the previous PR #283

polyglot/piranha/src/models/rule_store.rs
```diff
@@ -105,13 +105,10 @@ impl RuleStore {
...
     &mut self, rule: &Rule, tag_captures: &HashMap<String, String>,
   ) {
-    if let Ok(mut r) = rule.try_instantiate(tag_captures) {
-      if !self.global_rules.iter().any(|r| {
-        r.name().eq(&rule.name()) && r.replace().eq(&rule.replace()) && r.query().eq(&rule.query())
-      }) {
-        r.add_grep_heuristics_for_global_rules(tag_captures);
+    if let Ok(r) = rule.try_instantiate(tag_captures) {
+      if !self.global_rules.contains(&r) {
         #[rustfmt::skip]
```
lazaroclapp, line 110: Why do we need to skip formatting here?

polyglot/piranha/src/models/source_code_unit.rs
```diff
@@ -128,43 +128,45 @@ impl SourceCodeUnit {
...
-        //
-        self.add_to_substitutions(m.matches(), rule_store);
-
-        self.propagate(m.range(), rule.clone(), rule_store, parser);
+      // When rule is a "match-only" rule :
+      // Get all the matches
+      // Add mappings to the substitution
+      // Propagate each match. Note that,  we pass a identity edit (where old range == new range) in to the propagate logic.
+      // The next edit will be applied relative to the identity edit.
+      r @ _ => {
```
lazaroclapp, line 156: Isn't this exactly what you don't want with the new structural matches? Instead of handling this in the "everything else" case, you probably want to make sure you have a `MatchOnly` rule here, because in the future you might have more than two cases. Or am I getting the idea wrong?

lazaroclapp (commented):
Did a pass. Without understanding enough Rust to know whether this is optimal or not, the idea of having the rule "subtypes" checked by structural matching makes sense to me, as does using traits to group similar kinds of rules. It still feels in some ways more verbose than a traditional inheritance based hierarchy with a `Kind` enum, though... but I don't know the language well enough to know what's idiomatic for Rust.

PR #290, opened 2022-12-12
Handle `default` properly, and extract string literals to constants 

ketkarameya:
Addresses the comments from #283 . 
Extract string literal as constants

### DIFF
--- a/polyglot/piranha/src/models/language.rs
+++ b/polyglot/piranha/src/models/language.rs
@@ -5,7 +5,7 @@ use tree_sitter::{Parser, Query};
 use crate::utilities::parse_toml;
 
 use super::{
-  default_configs::default_language,
+  default_configs::{default_language, GO, JAVA, KOTLIN, PYTHON, STRINGS, SWIFT, TSX, TYPESCRIPT},
   outgoing_edges::Edges,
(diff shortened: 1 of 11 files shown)


polyglot/piranha/src/models/language.rs
```diff
@@ -75,27 +75,31 @@ impl PiranhaLanguage {
...
   fn from(language: &str) -> Self {
     match language {
-      "java" => PiranhaLanguage::default(),
-      "go" => PiranhaLanguage {
+      JAVA => {
+        let rules: Rules = parse_toml(include_str!("../cleanup_rules/java/rules.toml"));
+        let edges: Edges = parse_toml(include_str!("../cleanup_rules/java/edges.toml"));
+        Self {
+          name: language.to_string(),
+          supported_language: SupportedLanguage::default(),
```
lazaroclapp, line 90: Shouldn't this be `SupportedLanguage::Java` here? Since, again, we want to concentrate the "Java is the default" logic in a single place that can be changed later?

lazaroclapp (commented):
Mostly looks good to me, one minor comment.

lazaroclapp (approved):
LGTM!

PR #291, opened 2022-12-13
Add support to execute Piranha only on a specific file

ketkarameya:
This PR does two things : 
1. Adds support to allow developers to apply piranha only on file 
2. Renames `PiranhaArgument` fields (so that they do not break the command line interface we provided earlier) 
3. Update readme


lazaroclapp (approved):
Makes sense to me :)

PR #295, opened 2022-12-26
Different shape for dummy nodes/rules

### LINKED ISSUE #288: Display dummy nodes differently in the graph visualization
Dummy nodes have no "query"

dvmarcilio:
Resolves #288.

I experimented with shapes, and `doubleoctagon` was the best one.
Its automatic size seems just right while being distinguishable from default nodes.
Circle ones would get too large and would require manually set fixed sizes.


ketkarameya (commented):
LGTM.

ketkarameya (approved):

PR #297, opened 2022-12-27
Introduce the precommit hooks for formatting and linting

ketkarameya:
Please review this PR as two commits: 

[Add rust and toml formatting (and linting) as pre-commit (And CI step)](https://github.com/uber/piranha/pull/297/commits/8a10305f1980cb7144840fc12a9e9adc40bf6562)

* Uses [pre-commit](https://github.com/pre-commit/pre-commit) which is sort of a hook manager.
* Uses [taplo](https://taplo.tamasfe.dev/cli/introduction.html) for toml formatting 
* Runs rust format, cargo check and toml formatting as precommit hook 
* I add the same step to the CI too. 

[
Running rust and toml format](https://github.com/uber/piranha/pull/297/commits/2adb25e601803a97a569bed79cc7fab7bc2d235a)
Now since I add this step to CI (with `--all-files`) I have to format the files to satisfy this check. This commit does so  that.


polyglot/piranha/README.md
```diff
@@ -469,6 +469,10 @@ The purpose of Piranha Arguments is determining the behavior of Piranha.
 
 ## Contributing
 
+Prerequisites: 
+* Install [pre-commit](https://pre-commit.com/)
```
lazaroclapp, line 473: So, this requires manual installation steps? Isn't there a way to add the git pre-commit hook automatically on any local build like we do using Gradle in Java projects? I think I'd rather have a checked in pre-commit-hook.template and copy it on build, even if the configuration is less clean, just to avoid more manual installation steps.

lazaroclapp (approved):
Sounds good to me. All pip package installs are on a virtual environment, right?

lazaroclapp:
Actually, it seems like it's running into issues with setting the pre-commit hook in CI:

```
  thread 'main' panicked at 'Install pre-commit hook: Os { code: 2, kind: NotFound, message: "No such file or directory" }', build.rs:23:6
```

Also, given how many CI jobs we have, I wonder... do we still run any of the non-Polyglot versions of Piranha? Do others depend on them eternally? Should we start thinking about removing them and their corresponding jobs?

ketkarameya:
> Actually, it seems like it's running into issues with setting the pre-commit hook in CI:
> 
> ```
>   thread 'main' panicked at 'Install pre-commit hook: Os { code: 2, kind: NotFound, message: "No such file or directory" }', build.rs:23:6
> ```
> 
> Also, given how many CI jobs we have, I wonder... do we still run any of the non-Polyglot versions of Piranha? Do others depend on them eternally? Should we start thinking about removing them and their corresponding jobs?

Yes, I agree. We should do a decommissioning strategy for these "language specific" implementations. 
We not only have too many CI jobs, it also makes it more difficult for a new users to get started and understand what this repository is all about.

lazaroclapp (approved):
Ok, LGTM! Big formatting change, so skipping the second commit that applies most of the formatting, but assuming that one is automatic, this looks good. Will hopefully reduce formatting noise from PRs too! 🎉

PR #312, opened 2023-01-03
Move polyglot piraha code into `piranha`

ketkarameya:
our `src`, `tests`, `test-resources` are all at `root`.  There is also the `legacy` folder at the the root. So to make it somewheat similar, I move all poyglot-piranha code into folder named `piranha`


lazaroclapp (commented):
Nothing against the code changes in this diff per se, but... do we want to do this? One advantage of having all the polyglot piranha structure directly under the repo root is that once we are sure there are no active internal uses and no major external uses of the collection of old Piranha implementations, we can just remove `legacy/` and end with a very clean/standards repo structure.

ketkarameya:
I see what you mean. Yea, once we remove `legacy` we will end up flattening the structure again. Then let's continue with this current structure and create a plan to decommission those implementations.

PR #314, opened 2023-01-04
Remove support for Strings

ketkarameya:
Remove support for `strings`. We were supporting this format (language). 

Reason: 
Currently I am using a experimental grammar that I developed for strings format for a internal use case. This internal usecase is post poned. So I am removing the support for this format. I highly doubt any external folks are using this . :|  I will add it back when we are tackling that usecase


lazaroclapp (approved):
LGTM. Not sure what the plan is exactly that would make us re-add this or what the use case was besides being an example, but as far as the clean up diff goes, all seems good!

PR #315, opened 2023-01-04
Update-dependencies

ketkarameya:
Cleanup and update the dependencies. 
Remove github links where possible (replace with crate.io release)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ env_logger = "0.10.0"
 tempdir = "0.3"
 serde_json = "1.0.82"
 # TODO: Update after https://github.com/fwcd/tree-sitter-kotlin/pull/71 lands
-tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git", branch = "expose_node_function_value_parameters"}
+tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git", branch = "expose_node_function_value_parameters" }
 tree-sitter-java = "0.20.0"
 tree-sitter-swift = { git = "https://github.com/ketkarameya/tree-sitter-swift.git", branch = "add_parser" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -42,20 +42,21 @@ serde_derive = "1.0.136"
...
 tempdir = "0.3"
 serde_json = "1.0.82"
-tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git" }
-tree-sitter-java = { git = "https://github.com/tree-sitter/tree-sitter-java.git" }
-tree-sitter-swift = { git = "https://github.com/ketkarameya/tree-sitter-swift.git", branch = "add_parser" }
-tree-sitter-python = { git = "https://github.com/tree-sitter/tree-sitter-python.git" }
-tree-sitter-typescript = { git = "https://github.com/tree-sitter/tree-sitter-typescript.git" }
-# TODO: tree-sitter-go should be changed either to upstream or Uber main branch
+# TODO: Update after https://github.com/fwcd/tree-sitter-kotlin/pull/71 lands
+tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git", branch = "expose_node_function_value_parameters" }
```
lazaroclapp, line 52: Are these imports just pulling the latest commit in that branch? That's kind of horrifying from the point of view of reproducible builds. Is there a way to lock the remaining ones to a specific commit, rather than the head of a branch?

Either way, happy we are doing less of that...

lazaroclapp (approved):
An improvement, but wonder if we can do something to get rid of "snapshot"/"branch head" imports altogether.

Cargo.toml
ketkarameya: Addressed https://github.com/uber/piranha/pull/315/commits/418ebde9f182a78fbe6113f9dade6a06f07d2857

lazaroclapp (approved):

PR #316, opened 2023-01-05
Deprecate `run_piranha_cli` in favor of `execute_piranha`

ketkarameya:
This PR deprecate `run_piranha_cli` in favor of `execute_piranha`. I think the changes to `POYGLOT_README.md` gives a good idea around what this PR is about.  

Note that `execute_piranha` already exists. 

Benefits of this change :
* eliminates the need of specifying `piranha_arguments.toml` when using our _Python(/Rust) API_. `PiranhaArguments` can be specified via the API itself. This makes it easy to execute the same rule-set with different arguments.
* Significantly simplifies our core code

Changes: 

* The important changes are under `src` .
    * Delete `piranha_input.rs`. We no longer need this `enum` to handle the library entry point and command line entry point. Now all these entry-points in some sense directly populate the PiranhaArguments  
    * Generally adapt `piranha_arguments.rs` to this change : 
       *  Add `py_new` which acts as `__init__` for constructing `PiranhaArguments` in `Python` 
       *  exposes some more command line options 
       *  adapted and cleaned up the `merge` method
* The `tests` (rust) and `demos` (python) are adapted to the new API by : 
   * deleting `piranha_arguments.toml`
   * inlining its content into the _test_ or _demo_ code 
   * Replace `run_piranha_cli` call with `execute_piranha`

------

I will follow up this PR with another change which actually simplifies our tests by _executing the same rule-set with different arguments_.


POLYGLOT_README.md
```diff
@@ -114,25 +114,37 @@ Polyglot Piranha can be used as a python library or as a command line tool.
...
+    path_to_configurations = "...",
+    language= "java",
+    substitutions = {},
+    dry_run = False, 
+    cleanup_comments = True
+)
+piranha_summary = execute_piranha(piranha_arguments)
 ```
-<h5> Arguments </h5>
+The API `execute_piranha` accepts a `PiranhaArgument`
```
lazaroclapp, line 132:
```suggestion
The API `execute_piranha` accepts a `PiranhaArguments`
```

POLYGLOT_README.md
```diff
@@ -114,25 +114,37 @@ Polyglot Piranha can be used as a python library or as a command line tool.
...
-- `dry_run` : Disables in-place rewriting of code
+  * `edges.toml` : expresses the flow between the rules
+- (*required*) `language` (`str`) : Target language (`java`, `py`, `kt`, `swift`, `py`, `ts` and `tsx`)
+- (*required*) `substitutions` (`dict`): Substitutions to instantiate the initial set of feature flag rules
+- (*optional*) `dry_run` (`bool`) : Disables in-place rewriting of code
+- (*optional*) `cleanup_comments` (`bool`) : Enables deletion of associated comments
+- (*optional*) `cleanup_comments_buffer` (`usize`): The number of lines to consider for cleaning up the comments
+- (*optional*) `number_of_ancestors_in_parent_scope` (`usize`): The number of ancestors considered when `PARENT` rules
+- (*optional*) `delete_file_if_empty` (`bool`): User option that determines whether an empty file will be deleted
+- (*optional*) `delete_consecutive_new_lines` (`bool`) : Replaces consecutive `\n`s  with a `\n`
```
lazaroclapp, line 146:
```suggestion
- (*optional*) `delete_consecutive_new_lines` (`bool`) : Replaces consecutive `\n`s  with a single `\n`
```

demo/find_replace_custom_cleanup_demos.py
```diff
@@ -1,6 +1,6 @@
 from collections import Counter
 from os.path import join, dirname, getmtime
-from polyglot_piranha import run_piranha_cli
+from polyglot_piranha import run_piranha_cli, execute_piranha,PiranhaArguments
```
lazaroclapp, line 3:
```suggestion
from polyglot_piranha import run_piranha_cli, execute_piranha, PiranhaArguments
```

src/models/piranha_arguments.rs
```diff
@@ -61,172 +68,237 @@ pub struct PiranhaArguments {
...
-  // a list of file extensions
-  // #[get = "pub"]
-  #[builder(default = "default_languages()")]
+  // the target language
+  #[builder(default = "default_language()")]
   #[clap(skip)]
-  #[serde(default = "default_languages")]
-  language: Vec<String>,
+  #[serde(default = "default_language")]
+  language: String,
```
lazaroclapp, line 83: We are switching from a vector of languages to a single language. Does this mean that before we supported running a set of rules to fix point on multi-language projects? Do we plan to support that? For Android, this might be important, since there are java+kotlin projects both internally at Uber and more broadly externally, and it might not be the only case where it applies.

src/models/piranha_arguments.rs
```diff
@@ -61,172 +68,237 @@ pub struct PiranhaArguments {
...
 
   #[get = "pub"]
   #[builder(default = "default_piranha_language()")]
   #[clap(skip)]
   #[serde(skip)]
   piranha_language: PiranhaLanguage,
 
-  // User option that determines whether an empty file will be deleted
+  /// User option that determines whether an empty file will be deleted
```
lazaroclapp, line 91: Why are some of these `///` rather than `//`? Does that have an special meaning in Rust comments?

src/models/piranha_arguments.rs
```diff
@@ -61,172 +68,237 @@ pub struct PiranhaArguments {
...
+
+impl PiranhaArguments {
+  pub fn get_language(&self) -> String {
+    self.language.clone()
+  }
+
+  pub fn merge(&self, other: PiranhaArguments) -> Self {
+    /// Accepts field name (e.g. `language`) and function name (e.g. `default_language`) which returns default value for that field.
+    /// It checks if the value `self.language` is same as value returned by `default_language()`.
+    /// If the value is same it returns the value from `other (e.g. `other.language`) else it returns it from `self`
```
lazaroclapp, line 240: So, the default is the same explicitly setting this to `java`. Does it mean that when merging two instances of `PiranhaArguments`, one of which has `language = "java"` explicitly and the other some other language, the non-java one will be preferred? When would we merge two sets of arguments, specially for two separate languages?

lazaroclapp (commented):
This PR is a bit too large to review in full detail, specially without a lot of knowledge about Rust configuration mechanisms and python bindings, but the explanation and what I saw at a superficial pass makes sense. Below are a few nits. My main question is about switching from a vector of languages to a single language. Does Piranha support or plan to support mixed language build targets? E.g. Kotlin+Java

src/models/piranha_arguments.rs
ketkarameya: `//` means a single line comment 
`///` is used for documenting a method/field/class 

Standard libraries that use documentation to create messages (for instance `clap` our command line parsing tool) recognize `///` for getting comments associated to elements.

src/models/piranha_arguments.rs
ketkarameya: Yes. So actually `language: Vec<String>`  was always unnecessary syntactic sugar.  I was using `language[0]` at the [getter](https://sourcegraph.com/github.com/uber/piranha/-/blob/src/models/piranha_arguments.rs?L141). 
I cleaned this up by migrating it from `Vec<String>` to `String`.

At this moment (or previously) we do not have support for 2 languages at once.
But internally we manage this mix language support by interleaving cleanup steps in each language.
This is still possible now (because each step essentially transforms code of one language). 
For instance, we run one step that updates `BUCK` files and follow that step up by migrating the API (`java` and `kotlin`). 

However, there are scenarios when interleaving at such coarse level will not be possible. We would want to invoke a Java rule after a Kotlin or vice versa. But we do not have support for this right now. 
I always wanted to do this, hence I had `language` as `vector`. But I guess its a bit lower on our priority list.

src/models/piranha_arguments.rs
ketkarameya: We will merge them, once we start supporting `language: Vec<String>` (look at my comment above)

lazaroclapp (approved):
Clarifications make sense to me! LGTM as far as I can judge this change.

PR #317, opened 2023-01-06
Add python formatting as pre-commit hook

ketkarameya:
We have quite a few `py` files .
This PR adds `black` formatter as a pre-commit hook for python file formatting. 
https://github.com/psf/black


lazaroclapp (approved):
LGTM. One question, though: do we use this or any other standard formatter for python code either internally or on other Uber OSS projects? Just wondering if we are using as standard a toolchain as we can (for a project mainly in Rust, but still...)

PR #320, opened 2023-01-09
Use `Edit` object instead of passing around range and replacement

ketkarameya:
In stead of passing around `replacement_string` and `range` pass the `edit` object (that contains this information)


lazaroclapp (approved):
LGTM, given existing tests for this functionality pass after changing the API :)

PR #321, opened 2023-01-09
Enhance match struct with the `matched_string` field 

ketkarameya:
Add the field `matched_string` to `Match` struct. This field will store the code snippet matched by the rule.

### DIFF
--- a/src/models/edit.rs
+++ b/src/models/edit.rs
@@ -46,9 +46,13 @@ impl Edit {
     }
   }
 
-  pub(crate) fn delete_range(replacement_range: Range) -> Self {
+  pub(crate) fn delete_range(code: &str, replacement_range: Range) -> Self {
     Self {
-      p_match: Match::new(replacement_range, HashMap::new()),
+      p_match: Match::new(
+        code[replacement_range.start_byte..replacement_range.end_byte].to_string(),
+        replacement_range,
+        HashMap::new(),
+      ),
       replacement_string: String::new(),
       matched_rule: "Delete Range".to_string(),
     }
--- a/src/models/matches.rs
+++ b/src/models/matches.rs
@@ -20,6 +20,9 @@ use serde_derive::Serialize;
 #[derive(Serialize, Debug, Clone, Getters)]
 #[pyclass]
 pub(crate) struct Match {
+  // Code snippet that matched
+  #[get = "pub"]
+  matched_string: String,
   // Range of the entire AST node captured by the match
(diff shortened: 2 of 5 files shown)


lazaroclapp (approved):
Makes sense to me. 

That said, question: Are these substrings eager copies in memory or are they stored some other way? Any concern about memory usage ballooning due to multiple copies of substrings of the code under analysis? Specially under, say, a long chain of simplifications on the same huge boolean expression? (Seems unlikely, but still, one alternative is to keep just the start and end position and a pointer to the full source - or even the file, to be read on demand and cached - and then just constructing the substring when requested through a method of the match struct, no?)

ketkarameya:
I totally agree to what u say. I think we should not do this. 
I was leaning towards usability over performance. But i feel this will get really bad when we capture these large constructs like `class` or `file`. 
With storing references comes lifetimes :|

ketkarameya:
Another problem is that, we are using this `Match` struct for our python bindings too. 
And `pyo3` library doesnt support struct with lifetimes.

dvmarcilio:
In the case of edits/rewrites, I wonder what would happen if there is _no_ matched string.
I may be misunderstanding, but we might not be able to get the original match content if rewrites (e.g., delete a function) happen.

ketkarameya:
yes. And Diego makes a good point too. If we dont do this eagerly we will not be able to report it at all.

ketkarameya:
I think I am gonna merge this. It might have performance implications (in memory) but I guess its alright. Neither do we have any benchmarks to empirically prove how bad it is. @lazaroclapp  what do u think ? (I know u have already accepted the PR)

lazaroclapp:
> I think I am gonna merge this. It might have performance implications (in memory) but I guess its alright. Neither do we have any benchmarks to empirically prove how bad it is. @lazaroclapp what do u think ? (I know u have already accepted the PR)

I think this makes sense to me, given the complications involved in designing something that would work well with deletions (I am sure there is some involved copy-on-write way to handle it, but it feels like premature optimization). Separate from this PR, a good question is whether we have any way to measure internally Piranha's memory usage. But, also, I wouldn't worry too much about it unless we start seeing issues, since it's not like Piranha is running during a standard build or anything like that. It is very much an "off main pipeline" tool 🙂

ketkarameya:
Thanks a lot. I will merge this PR. 
Filed the issue https://github.com/uber/piranha/issues/322 to track this.

PR #326, opened 2023-01-13
Use the default value for `dry_run` and fix bug related to `delete_file_if_empty`

ketkarameya:
Dependent on #325 

There were some entangled bugs : 

1. `delete_file_if_empty` did not run as expected when set to `false` 
2. `delete_consecutive_new_lines` was only performed during persist, but did not work for our API 

One thing to note here is that, these two flags can only be tested when `dry_run = false` 
Our unit-test for `delete_consecutive_new_lines` exclusively ran `persist(...)`  to test for the functionality, but we did not have end to end tests checking for this behavior.  We did not have e2e tests for this because our text comparison is white-space agnostic. 
Moreover, we run all our e2e tests in `dry_run` mode. 

---
So for this purpose we do the following: 
*  fix the entangled bugs : 
   *  Move  `delete_consecutive_new_lines` inside the cleanup 
   *  fix the bug - _`delete_file_if_empty` did not run as expected when set to `false`_ 
* Earlier our `check_result` compared API result to the files in `path_to_expected`
   * The check now compares files in `path_to_codebase` (which is updated by piranha) to the files in `path_to_expected`
   * Why are we adding those 4 java files? 
      *  So earlier just because how the logic was we did not require to have **unchanged** files in the `path_to_expected`
      * Now since we compare directories, we have to add the  **unchanged** files in the `path_to_expected`


src/tests/test_piranha_swift.rs
```diff
@@ -29,12 +29,12 @@ create_rewrite_tests! {
     substitutions = substitutions! {
       "stale_flag_name" => "Premium-Icon"
     },
-    cleanup_comments = true;
+    cleanup_comments = true, delete_file_if_empty= false;
```
lazaroclapp, line 32: Why the inconsistent spacing around `=` here? (between this parameter and the previous two) And why isn't that auto-formatted? (I thought we were running an auto-formatter on every commit now)

src/tests/test_piranha_java.rs
```diff
@@ -32,14 +32,14 @@ create_rewrite_tests! {
       "treated"=>  "true",
       "treated_complement" => "false",
       "namespace" => "some_long_name"
-    };
+    }, delete_file_if_empty= false;
```
lazaroclapp, line 35: Do we have a test case for `delete_file_if_empty = true`? Or is that the default for all other tests?

src/tests/mod.rs
```diff
@@ -63,23 +62,37 @@ fn copy_folder(src: &Path, dst: &Path) {
...
+    let mut count = 0;
+    for dir_entry in fs::read_dir(path).unwrap() {
+      if dir_entry.unwrap().path().is_file() {
+        count += 1;
+      }
+    }
+    count
+  };
+
+  assert_eq!(count_files(path_to_codebase), count_files(path_to_expected));
```
lazaroclapp, line 78: Won't this break for cases where we _are_ deleting empty files? Don't we want to be able to test for that too?

src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -182,33 +182,30 @@ fn test_apply_edit_comma_handling_via_regex() {
...
+  source_code_unit.perform_delete_consecutive_new_lines();
+  source_code_unit.persist(piranha_arguments);
   check_predicate(&tmp_dir)
 }
 
 #[test]
 fn test_persist_delete_file_when_empty() -> Result<(), io::Error> {
   let args = PiranhaArgumentsBuilder::default()
     .delete_consecutive_new_lines(true)
-    .delete_file_if_empty(true)
```
lazaroclapp, line 211: Ok, but we should also have a test where this _is_ set to `true`, no?

lazaroclapp (commented):
So, I didn't fully follow the description, but if we want to test `delete_file_if_empty`, shouldn't we be testing it under both conditions? As far as I can tell, all the tests are for when `delete_file_if_empty = false`, but we probably want to make sure `delete_file_if_empty = true` also works as intended (i.e. it removes empty files [after rewrite] from the output). Am I missing something?

src/tests/test_piranha_swift.rs
ketkarameya: I will add auto-formatting for swift files in a follow up commit.

src/tests/test_piranha_java.rs
ketkarameya: `delete_file_if_empty = true` is default. So we don't explicitly test that.

src/tests/test_piranha_swift.rs
lazaroclapp: This file isn't swift, though, it's `.rs` :/

src/tests/test_piranha_java.rs
lazaroclapp: Do we? Because I see elsewhere (see the next comment) that we check that the count of files at `path_to_codebase` and `path_to_expected` is the same. Not sure if that check holds for _every_ test, but, if it does, then we are never actually testing that empty files are being deleted correctly...

ketkarameya:
[This](https://github.com/uber/piranha/blob/master/src/models/piranha_arguments.rs#L93) sets the default value for the builder pattern. As mentioned in my comments above, the deault value of  `delete_file_if_empty` is true so we only have tests where we explicitly set to to `false`. Note also have unit tests for each of these features-flags. I plan to move [these](https://github.com/uber/piranha/blob/73e84d1c4829196032584d281621a38c41d6057b/src/models/unit_tests/source_code_unit_test.rs#L184) tests to our recently developed macro soon.

src/tests/mod.rs
ketkarameya: https://github.com/uber/piranha/blob/73e84d1c4829196032584d281621a38c41d6057b/src/models/unit_tests/source_code_unit_test.rs#L208 

We test this functionality here in this "unit test" (Atleast thats what I called this before). 
This unit test does kinda what `create_rewrite_rules` macro does. I plan to move these unit test to the macro.
These unit tests run via this `execute_persist_in_temp` method which is equivalent to `create_rewrite_rules` now.

ketkarameya (commented):
I have left my comment addressing where is the test which turns on `delete_file_if_empty`. But I also agree that its worth having one of these current e2e tests with `delete_file_if_empty=true`. I checked we do not have a e2e test where `input/` files > `expected` files. I will do that.

src/tests/mod.rs
lazaroclapp: Ok, still think it might make sense to test this at least once in our "integration" tests to verify that it indeed counts things correctly when `|input/| < |expected/|`, but at this point I am fairly convinced this isn't _required_, sorry for the confusion 🙂

lazaroclapp (approved):
Fine to land by me, after the discussion!

@ketkarameya Up to you if the integration test with file deletion should be here or on a follow up. In the former case, let me know so I can do a final pass on this PR. On the later, feel free to land this and add me to that follow up instead 🙂

ketkarameya:
Please do a final pass on the PR (I have updated one integration test to verify this behavior)

lazaroclapp (approved):
LGTM! :) (I see the test case was even simpler than I though!)

PR #330, opened 2023-01-18
Introduce `InstantiatedRule` 

ketkarameya:
Based on #329 

Introduce `InstantiatedRule` which is composed of `Rule` and `substitutions` (previously `grep_heuristics`).
An `InstantiatedRule` is constructed when a `Rule` is instantiated with `substitutions`. 
* Removed `grep_heuristics` property from `Rule` 
* Change type of `query`, `replace` and `replace_node` from `Option<String>` -> `String` from Rule . This change required updating the methods `is_dummy`, is_match_only`. (This is an un-related refactoring 😢 , but its small so its alright ? 😵‍💫 ) 
* Use `InstantiatedRule` throughout `source_code_unit` 

Earlier, we were updating `grep_heursitic` property of the Rule. This made it ambiguous at a type-level whether the rule is instantiated or contains holes.

-------
I am open to rename `InstantiatedRule`. Maybe it sounds a bit awkward. `ResolvedRule` ? `FilledRules` ?

### DIFF
--- a/src/models/default_configs.rs
+++ b/src/models/default_configs.rs
@@ -82,3 +82,15 @@ pub fn default_piranha_language() -> PiranhaLanguage {
 pub fn default_delete_consecutive_new_lines() -> bool {
   false
 }
+
+pub fn default_query() -> String {
+  String::new()
+}
+
+pub fn default_replace_node() -> String {
+  String::new()
+}
+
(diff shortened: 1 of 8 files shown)


lazaroclapp (approved):
Makes sense to me!

PR #331, opened 2023-01-18
Remove unnecessary field `global_tags` from RuleStore 

ketkarameya:
Earlier we were collecting all `global_tags` (as and when we found them) in a property of `RuleStore`
When we matched a rule, we checked if it contains `global_tags` and added them to the rule store that we are passing around. 

On further investigation I found that, we do not really need to do this.  It can be done more cleverly and cleanly.

Once a source code unit has been processed we simply retain all the substitutions that are "global" i.e. substitutions whose keys start with `global_tag_prefix`. Lol, idk why i did not think of this before. :\

Removed the functions `add_global_tags` and `add_to_substitutions`. Moved the"logic" that identifies global tags from `add_global_tags` to `global_substitutions` inside `source_code_unit.rs`

-----
P.S. I know `source_code_unit.rs` is becoming this gigantic file :|.  It is not that bad.  I will clean it up.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,10 +136,12 @@ impl Piranha {
   fn perform_cleanup(&mut self) {
     // Setup the parser for the specific language
     let mut parser = Parser::new();
+    let piranha_args = self.rule_store.piranha_args().clone();
     parser
-      .set_language(*self.rule_store.piranha_args().piranha_language().language())
+      .set_language(*piranha_args.piranha_language().language())
       .expect("Could not set the language for the parser.");
 
+    let mut current_global_substitutions = piranha_args.input_substitutions().clone();
     // Keep looping until new `global` rules are added.
     loop {
       let current_rules = self.rule_store.global_rules().clone();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -136,10 +136,12 @@ impl Piranha {
   fn perform_cleanup(&mut self) {
     // Setup the parser for the specific language
     let mut parser = Parser::new();
+    let piranha_args = self.rule_store.piranha_args().clone();
```
lazaroclapp, line 139: Question: What changed that requires `.clone()` here? The only place I see any reference within this object (not sure if the correct Rust terminology) being passed somewhere new is with `current_global_substitutions`, but we are calling `.clone()` there at the level of `piranha_args.input_substitutions()`. What am I missing?

lazaroclapp (approved):
Minor question about Rust that doesn't seem obvious to me but might be clear if you know the language 😅 , otherwise LGTM.

src/lib.rs
ketkarameya: Aah! I had to clone it because I extracted `piranha_args` as a variable. 

`self.rule_store.piranha_args()` returns `&PiranhaArguments`.  So if we don't clone it here, it is going to be a "Immutable borrow" for `self.rule_store`. Instead of borrowing I just cloned it :| 
If I drop the `clone()` here, rust compiler complains that I am using `self.rule_store` as immutable here and mutable (later below in the function). 

The work-around for this is to "inline" this variable.

PR #334, opened 2023-01-24
Move the source_code_unit_tests to e2e tests. 

ketkarameya:
This PR removes the tests in `source_code_unit_test.rs` using `execute_persist_int_temp` and convert them to e2e tests.

This PR has the following changes: 
1. Delete all the tests from `source_code_unit_test.rs` using `execute_persist_int_temp`
2. Added equivalent tests under `test_piranha_java.rs`
* We could not add all tests using the macro (`create_rewrite_test`) because this macro creates a test that compares files ignoring whitespaces difference. 
* the features `delete_consecutive_new_lines` cannot be correctly tested by `create_rewrite_test` macro (because the file comparison in the tests need to account for new line changes) 
   * For this reason I separately implement the +ve and -ve test for `delete_consecutive_new_lines`
      *  I literally copy pasted the macro, instantiated it manually and passed false to `execute_piranha_and_check_result`
3. I do some refactoring of the macro `create_rewrite_test` itself (I know it could have been in another diff :| ) 
    * This refactoring basically extracts some code from this macro into the function `execute_piranha_and_check_result`
    *  this refactoring also adds a functionality which allows file comparison with or without ignoring white spaces.

### DIFF
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -11,10 +11,13 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
+use crate::execute_piranha;
+use crate::models::piranha_arguments::PiranhaArguments;
 use crate::utilities::{eq_without_whitespace, read_file};
-use log::error;
+
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use tempdir::TempDir;
 
 mod test_piranha_java;
 mod test_piranha_kt;
@@ -47,22 +50,40 @@ fn initialize() {
 /// * dest: Path to destination
 ///
 /// This method causes side effects - writes new files to a directory
-fn copy_folder(src: &Path, dst: &Path) {
-  for entry in fs::read_dir(src).unwrap() {
+fn copy_folder_to_temp_dir(src: &PathBuf) -> TempDir {
+  // Copy the test scenario to temporary directory
+  let temp_dir = TempDir::new_in(".", "tmp_test").unwrap();
+  let temp_dir_path = &temp_dir.path();
+  for entry in fs::read_dir(src.as_path()).unwrap() {
     let entry = entry.unwrap();
     let path = entry.path();
     if path.is_file() {
       _ = fs::copy(
         path.to_str().unwrap(),
-        dst.join(path.file_name().unwrap()).to_str().unwrap(),
+        temp_dir_path
+          .join(path.file_name().unwrap())
+          .to_str()
+          .unwrap(),
       )
       .unwrap();
     }
   }
+  temp_dir
 }
 
 /// Checks if the file updates returned by piranha are as expected.
-fn check_result(path_to_codebase: &Path, path_to_expected: &Path) {
+fn execute_piranha_and_check_result(
+  piranha_arguments: &PiranhaArguments, path_to_expected: &Path, files_changed: usize,
+  ignore_whitespace: bool,
+) {
+  let path_to_codebase = Path::new(piranha_arguments.path_to_codebase());
+
+  let output_summaries = execute_piranha(piranha_arguments);
+
+  assert!(output_summaries.iter().any(|x| !x.rewrites().is_empty()));
+
+  assert_eq!(output_summaries.len(), files_changed);
+
   let mut all_files_match = true;
 
   let count_files = |path: &Path| {
@@ -86,10 +107,16 @@ fn check_result(path_to_codebase: &Path, path_to_expected: &Path) {
       let expected_file_path = path_to_expected.join(file_name);
       let expected_content = read_file(&expected_file_path).unwrap();
 
-      if !eq_without_whitespace(&cb_content, &expected_content) {
-        all_files_match = false;
-        error!("{}", &cb_content);
+      if (ignore_whitespace && eq_without_whitespace(&cb_content, &expected_content))
+        || (cb_content
+          .trim_end()
+          .eq(&expected_content.trim_end().to_string()))
+      {
+        continue;
       }
+
+      all_files_match = false;
+      print!("{}", &cb_content);
     }
   }
 
@@ -167,34 +194,18 @@ macro_rules! create_rewrite_tests {
     fn $test_name() {
       initialize();
       let _path= PathBuf::from("test-resources").join($language).join($path_to_test);
-      let path_to_codebase = _path.join("input").to_str().unwrap().to_string();
-      let path_to_configurations = _path.join("configurations").to_str().unwrap().to_string();
-      let path_to_expected = _path.join("expected");
-
-      // Copy the test scenario to temporary directory
-      let temp_dir = TempDir::new_in(".", "tmp_test").unwrap();
-      let temp_dir_path = &temp_dir.path();
-      copy_folder(
-        Path::new(&path_to_codebase),
-        temp_dir_path,
-      );
+      let temp_dir= copy_folder_to_temp_dir(&_path.join("input"));
 
       let piranha_arguments =  piranha_arguments!{
-        path_to_codebase = temp_dir_path.to_str().unwrap().to_string(),
-        path_to_configurations = path_to_configurations,
+        path_to_codebase = temp_dir.path().to_str().unwrap().to_string(),
+        path_to_configurations = _path.join("configurations").to_str().unwrap().to_string(),
         language= $language.to_string(),
         $(
-         $kw = $value,
+          $kw = $value,
         )*
       };
 
-      let output_summaries = execute_piranha(&piranha_arguments);
-      // Checks if there are any rewrites performed for the file
-      assert!(output_summaries.iter().any(|x|!x.rewrites().is_empty()));
-
-      assert_eq!(output_summaries.len(), $files_changed);
-      check_result(temp_dir_path, &path_to_expected);
-
+      execute_piranha_and_check_result(&piranha_arguments, &_path.join("expected"), $files_changed, true);
       // Delete temp_dir
       _ = temp_dir.close().unwrap();
     }
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -88,8 +88,8 @@ fn execute_piranha_and_check_result(
 
   let count_files = |path: &Path| {
     let mut count = 0;
-    for dir_entry in fs::read_dir(path).unwrap() {
-      if dir_entry.unwrap().path().is_file() {
+    for dir_entry in fs::read_dir(path).unwrap().flatten() {
+      if dir_entry.path().is_file() && ".placeholder" != dir_entry.file_name().to_str().unwrap() {
         count += 1;
       }
     }
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -40,7 +40,7 @@ fn initialize() {
   });
 }
 
-// We use a `.placeholder` file because git does not allows us to commit an empty directory
+// We use a `.placeholder` file because git does not allow us to commit an empty directory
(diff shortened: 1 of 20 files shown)


src/tests/mod.rs
```diff
@@ -47,28 +50,46 @@ fn initialize() {
...
 }
 
 /// Checks if the file updates returned by piranha are as expected.
-fn check_result(path_to_codebase: &Path, path_to_expected: &Path) {
+fn execute_piranha_and_check_result(
+  piranha_arguments: &PiranhaArguments, path_to_expected: &Path, files_changed: usize,
+  ignore_whitespace: bool,
+) {
+  let path_to_codebase = Path::new(piranha_arguments.path_to_codebase());
+
```
lazaroclapp, line 80: Any reason for all the empty lines here?

src/tests/mod.rs
```diff
@@ -47,28 +50,46 @@ fn initialize() {
...
+  let output_summaries = execute_piranha(piranha_arguments);
+
+  assert!(output_summaries.iter().any(|x| !x.rewrites().is_empty()));
```
lazaroclapp, line 83: What is this line checking for? Maybe worth a comment about what this check means?

There was a comment before: `// Checks if there are any rewrites performed for the file`, though I'd use the opportunity to explain it more clearly. Is the check here to make sure that at least one file was changed? (i.e. that the test as a whole actually resulted in a rewrite) Why do we need this test in addition to the next one, then? (Not saying we don't, just not clear to me from reading the code)

src/tests/mod.rs
```diff
@@ -47,28 +50,46 @@ fn initialize() {
...
+  assert_eq!(output_summaries.len(), files_changed);
+
   let mut all_files_match = true;
 
   let count_files = |path: &Path| {
     let mut count = 0;
-    for dir_entry in fs::read_dir(path).unwrap() {
-      if dir_entry.unwrap().path().is_file() {
+    for dir_entry in fs::read_dir(path).unwrap().flatten() {
+      if dir_entry.path().is_file() && ".placeholder" != dir_entry.file_name().to_str().unwrap() {
```
lazaroclapp, line 92: What is `".placeholder"`? I am assuming that we add these to empty output directories so that git allows to commit the directory? Is that true? Either way this could use a comment explaining.

Also, maybe worth it being a constant defined at the top of the file?

src/tests/mod.rs
```diff
@@ -86,10 +107,16 @@ fn check_result(path_to_codebase: &Path, path_to_expected: &Path) {
       let expected_file_path = path_to_expected.join(file_name);
       let expected_content = read_file(&expected_file_path).unwrap();
 
-      if !eq_without_whitespace(&cb_content, &expected_content) {
-        all_files_match = false;
-        error!("{}", &cb_content);
+      if (ignore_whitespace && eq_without_whitespace(&cb_content, &expected_content))
+        || (cb_content
+          .trim_end()
+          .eq(&expected_content.trim_end().to_string()))
```
lazaroclapp, line 113: This is some weird formatting overall for the full conditional expression, but since code is now auto-formatted, I am fine with it if this is what the tool does 😅

src/tests/test_piranha_go.rs
```diff
@@ -12,15 +12,15 @@ Copyright (c) 2022 Uber Technologies, Inc.
 */
 
 use super::{
-  check_result, copy_folder, create_match_tests, create_rewrite_tests, initialize, substitutions,
+  copy_folder_to_temp_dir, create_match_tests, create_rewrite_tests,
+  execute_piranha_and_check_result, initialize, substitutions,
```
lazaroclapp, line 16: We add `execute_piranha_and_check_result`, but we aren't actually using it in this file for now, are we?

src/tests/test_piranha_kt.rs
```diff
@@ -16,11 +16,12 @@ use crate::models::{
   piranha_arguments::{piranha_arguments, PiranhaArgumentsBuilder},
 };
 
-use super::{check_result, copy_folder, create_rewrite_tests, initialize, substitutions};
-use crate::execute_piranha;
+use super::{
+  copy_folder_to_temp_dir, create_rewrite_tests, execute_piranha_and_check_result, initialize,
```
lazaroclapp, line 20: Same comment as for `test_piranha_go.rs`

src/tests/test_piranha_python.rs
```diff
@@ -12,15 +12,15 @@ Copyright (c) 2022 Uber Technologies, Inc.
 */
 
 use super::{
-  check_result, copy_folder, create_match_tests, create_rewrite_tests, initialize, substitutions,
+  copy_folder_to_temp_dir, create_match_tests, create_rewrite_tests,
+  execute_piranha_and_check_result, initialize, substitutions,
```
lazaroclapp, line 16: Same comment as for `test_piranha_go.rs`

src/tests/test_piranha_swift.rs
```diff
@@ -11,14 +11,16 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
-use super::{check_result, copy_folder, create_rewrite_tests, initialize, substitutions};
-use crate::execute_piranha;
+use super::{
+  copy_folder_to_temp_dir, create_rewrite_tests, execute_piranha_and_check_result, initialize,
```
lazaroclapp, line 15: Same comment as for `test_piranha_go.rs`

src/tests/test_piranha_java.rs
```diff
@@ -59,13 +59,55 @@ create_rewrite_tests! {
...
+  test_user_option_do_not_delete_if_empty : "user_option_do_not_delete_if_empty", 1, delete_file_if_empty=false;
 }
 
 create_match_tests! {
   JAVA,
   test_java_match_only: "structural_find", 20;
 }
 
+#[test]
+
```
lazaroclapp, line 72: Remove the empty line? (Not even sure if this breaks the pragma/annotation above... and even if it doesn't I don't see why we have the space)

src/tests/test_piranha_java.rs
```diff
@@ -59,13 +59,55 @@ create_rewrite_tests! {
...
+  initialize();
+  let _path = PathBuf::from("test-resources")
+    .join(JAVA)
+    .join("user_option_delete_consecutive_lines");
+  let temp_dir = copy_folder_to_temp_dir(&_path.join("input"));
+
+  let piranha_arguments = piranha_arguments! {
+    path_to_codebase = temp_dir.path().to_str().unwrap().to_string(),
+    path_to_configurations = _path.join("configurations").to_str().unwrap().to_string(),
+    language= JAVA.to_string(),
```
lazaroclapp, line 83:
```suggestion
    language = JAVA.to_string(),
```

src/tests/test_piranha_java.rs
```diff
@@ -59,13 +59,55 @@ create_rewrite_tests! {
...
+    delete_consecutive_new_lines= true,
```
lazaroclapp, line 84:
```suggestion
    delete_consecutive_new_lines = true,
```

Just for consistency between all four options being passed. Unless there is any significance of `a = b` vs `a= b` somehow that I am missing.

src/tests/test_piranha_java.rs
```diff
@@ -59,13 +59,55 @@ create_rewrite_tests! {
...
+  initialize();
+  let _path = PathBuf::from("test-resources")
+    .join(JAVA)
+    .join("user_option_do_not_delete_consecutive_lines");
+  let temp_dir = copy_folder_to_temp_dir(&_path.join("input"));
+
+  let piranha_arguments = piranha_arguments! {
+    path_to_codebase = temp_dir.path().to_str().unwrap().to_string(),
+    path_to_configurations = _path.join("configurations").to_str().unwrap().to_string(),
+    language= JAVA.to_string(),
```
lazaroclapp, line 103:
```suggestion
    language = JAVA.to_string(),
```

src/tests/test_piranha_java.rs
```diff
@@ -59,13 +59,55 @@ create_rewrite_tests! {
...
+    delete_consecutive_new_lines= true,
+  };
+
+  execute_piranha_and_check_result(&piranha_arguments, &_path.join("expected"), 1, false);
+  // Delete temp_dir
+  temp_dir.close().unwrap();
+}
+
+#[test]
+fn test_user_option_do_not_delete_consecutive_lines() {
```
lazaroclapp, line 93: So, while they don't fit the existing macro... aren't these two tests identical except for the path (i.e. `"user_option_delete_consecutive_lines"` vs `"user_option_do_not_delete_consecutive_lines") and the value of "delete_consecutive_new_lines"? Couldn't both test cases call the same function which takes this two arguments and avoid some code duplication that way? (I guess sometimes it's good to have code duplication in tests if it makes them more readable, but in this case I feel it would both reduce code _and_ highlight what is actually different between the two cases in terms of configuration)

lazaroclapp (commented):
Finally got a chance to do a pass. Mostly minor nits, I think.

src/tests/mod.rs
ketkarameya: I had this check before we had updated all of this code. I guess now it seems a bit redundant.
It also restricts us to test scenarios where we always expect a rewrite. Maybe I want to add a test case where I dont expect a rewrite :) 
I am going to just remove this.

src/tests/test_piranha_go.rs
ketkarameya: This is because the `create_rewrite_tests` macro expands to use the `execute_piranha_and_check_result` method.

ketkarameya (commented):
Addressed comments  [here](https://github.com/uber/piranha/pull/334/commits/edefe9c162b6f9f5e1ac430a6154484e6964644f)

src/tests/mod.rs
lazaroclapp: No action needed here, as far as I am concerned 🙂

src/tests/mod.rs
```diff
@@ -47,28 +53,47 @@ fn initialize() {
...
+
   let mut all_files_match = true;
 
   let count_files = |path: &Path| {
     let mut count = 0;
-    for dir_entry in fs::read_dir(path).unwrap() {
-      if dir_entry.unwrap().path().is_file() {
+    for dir_entry in fs::read_dir(path).unwrap().flatten() {
+      if dir_entry.path().is_file() {
+        // If the directory contains a file named `.placeholder` we assume that the directory is empty
```
lazaroclapp, line 93: Should we assume the directory is empty? Or just not count that particular file? Or check that the directory is otherwise empty and throw an exception if it isn't? The first of the three options seems like the most prone to silent errors...

src/tests/test_piranha_java.rs
```diff
@@ -85,3 +84,37 @@ fn test_scenarios_find_and_propagate_panic() {
...
+  _helper_user_option_delete_consecutive_lines(false);
+}
+
+fn _helper_user_option_delete_consecutive_lines(delete_consecutive_new_lines: bool) {
+  initialize();
+  let _path = PathBuf::from("test-resources")
+    .join(JAVA)
+    .join(if delete_consecutive_new_lines {
+      "user_option_delete_consecutive_lines"
```
lazaroclapp, line 103: I'd still think it would be clearer if we passed these as part of the call to `_helper_user_option_delete_consecutive_lines`, but honestly a matter of taste at this point...

lazaroclapp (approved):
LGTM! Remaining comments are very much optional, but worth considering 🙂

src/tests/mod.rs
ketkarameya: i think lets go with 2. ITs more intuitive.

PR #336, opened 2023-01-25
Created polyglot rules for  cleaning up logical operations in swift

Satyam1749:
Created swift rules in demo folder at the moment.
Rules Created:
 - AND operation with true/false
 - OR operation with true/false
 - NOT operation with true/false
 - Accessing an enum case's variable


ketkarameya:
Based on our offline discussion I am closing this PR.

PR #340, opened 2023-01-31
Set up test-resources and rust test for swift cleanup rules tests.

Satyam1749:
Setting up swift folder in test-resources. Added a rule to replace `enum variable` access with true.


test-resources/swift/cleanup_rules/configurations/rules.toml
```diff
@@ -0,0 +1,37 @@
...
+            suffix: (navigation_suffix
+                suffix: (simple_identifier) @param))
+        suffix: (navigation_suffix
+            suffix: (simple_identifier) @access_identifier)) @parameter_access
+(#eq? @param "@stale_flag")
+(#eq? @access_identifier "isEnabled")
+)"""
+replace_node = "parameter_access"
+replace = "true"
+holes = ["stale_flag"]
```
ketkarameya, line 37:
```suggestion
holes = ["stale_flag"]
groups= ["replace_expression_with_boolean_literal"]
```

test-resources/swift/cleanup_rules/expected/TestEnum.swift
```diff
@@ -0,0 +1,17 @@
...
+// License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing permissions and
+// limitations under the License.
+
+enum TestEnum {
+    case stale_flag_one
+    var isEnabled: Bool {
+        return true
+    }
+}
```
ketkarameya, line 17: please add new line. I wonder how did this PR make it. It should have been caught by our pre-commit rule

ketkarameya (changes requested):
At a higher level this PR looks good to me. 
Please add `\n` characters for each file and make sure that there is some description for the PR

ketkarameya:
I wonder why `Fix End of Files` hook did not throw a fit. :|

Satyam1749:
> 

Maybe, because it was a swift file

ketkarameya:
Aaah ! I see . Its under `test-resource`. We only apply the new line linter on files under `src`.

ketkarameya (approved):
We don't need the `TestEnums.swift` right? I see there is no change in those files.

PR #341, opened 2023-01-31
Fix Cargo Clippy problems

ketkarameya:
Our `master` is failing because of some `cargo clippy` issue. `Clippy` is our linter. 
We have set up clippy such that, it fails if there are any linting issues for which `auto-fix` is available. 

When I look at the CI logs it shows that there are 5 warnings and 1 fix. 
When I look run cargo clippy locally i see 6 warnings and no fix. 

I am taking this opportunity to fix all the cargo clippy problems (These are not auto-fixable by cargo clippy). 
Now we have no warnings from this tool 

After I updated my local rust up toolchain, I could see those problems. I fixed them in the second commit.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,6 +11,7 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
+#![allow(deprecated)] // This prevents cargo clippy throwing warning for deprecated use.
 use models::{
   edit::Edit, matches::Match, piranha_arguments::PiranhaArguments,
   piranha_output::PiranhaOutputSummary, source_code_unit::SourceCodeUnit,
--- a/src/main.rs
+++ b/src/main.rs
@@ -49,8 +49,5 @@ fn write_output_summary(
       return;
     }
   }
-  panic!(
-    "Could not write the output summary to the file - {}",
-    path_to_json
-  );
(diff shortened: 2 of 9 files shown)


lazaroclapp (approved):
LGTM!

Did we run into this because we landed some change while GitHub CI was down for Uber? Or why were these issues not caught before landing?

ketkarameya:
I did not land anything when GitHub was down. I think there was a new version of clippy that our actions used.

lazaroclapp:
> I did not land anything when GitHub was down. I think there was a new version of clippy that our actions used.
> 
> We did not catch this before landing, because I did not retrigger build for PR job before it landed (which is not an usualy practice either) but we had this PR laying around for a while (and in that time i guess clippy version was updated)

Sorry, I did not understand this explanation.

1. Aren't cargo and clippy versions fixed unless we update them? (They should be)
2. "We did not catch this before landing" ... before landing what?
3. Shouldn't PRs need to be fully rebased, and therefore retrigger CI, before landing?

lazaroclapp:
Ok, makes sense to me, I didn't realize we were getting clippy from GitHub actions and not fixed to a specific version. Seems a bit weird that we are getting changing lint rules out of our CI without manual changes, though, but I get what happened at least 🙂 . Either way, PR looks good to me.

PR #346, opened 2023-02-01
Added rules for boolean_simplification of something_or_true and true_or_something

Satyam1749:
Rules added:

1.  `abc || true => true`
2. `true || abc => true`

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -51,6 +51,49 @@ groups = ["boolean_expression_simplify"]
 replace_node = "conjunction_expression"
 replace = "@lhs"
 
+#
+# Before 
+#   true || abcd()
+# After 
+#   true
+#
+[[rules]]
+name = "true_or_something"
+query = """(
+(disjunction_expression
+        lhs: [(boolean_literal) @true  
+            (tuple_expression 
+                value: (boolean_literal) @true)]
+        rhs: (_) 
+    ) @disjunction_expression
+(#eq? @true "true")
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "disjunction_expression"
+replace = "true"
+
+
+#
+# Before 
+#   abcd || true
+# After 
+#   true
+#
+[[rules]]
+name = "some_identifier_or_true"
+query = """(
+(disjunction_expression
+        lhs: (simple_identifier)
+        rhs: [(boolean_literal) @true  
+            (tuple_expression 
+                value: (boolean_literal) @true)]
+    ) @disjunction_expression
+(#eq? @true "true")
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "disjunction_expression"
+replace = "true"
+
 # Dummy rule that acts as a junction for all boolean based cleanups
 # Let's say you want to define rules from A -> B, A -> C, D -> B, D -> C, ... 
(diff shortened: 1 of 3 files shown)


src/cleanup_rules/swift/rules.toml
```diff
@@ -51,6 +51,49 @@ groups = ["boolean_expression_simplify"]
...
+        lhs: (_)
+        rhs: [(boolean_literal) @true  
+            (tuple_expression 
+                value: (boolean_literal) @true)]
+    ) @disjunction_expression
+(#eq? @true "true")
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "disjunction_expression"
+replace = "true"
```
ketkarameya, line 95:
```suggestion
replace = "@lhs"
```

Since _something_ can have side effects we dont want to remove it.

Satyam1749: But if we replace `true || sample_function()` with `sample_function()` and if `sample_function()` returns false, wouldn't that be wrong?

Satyam1749: Made changes as per our offline discussion.

src/cleanup_rules/swift/rules.toml
```diff
@@ -51,6 +51,49 @@ groups = ["boolean_expression_simplify"]
...
+#   true
+#
+[[rules]]
+name = "true_or_something"
+query = """(
+(disjunction_expression
+        lhs: [(boolean_literal) @true  
+            (tuple_expression 
+                value: (boolean_literal) @true)]
+        rhs: (simple_identifier)
```
ketkarameya, line 67: Here the rhs could be wildcard right?
```suggestion
        rhs: (_) 
```

ketkarameya (changes requested):

src/cleanup_rules/swift/rules.toml
Satyam1749: we decided to change only `true || variable` right and not `true || func()` with `true`

ketkarameya: but `func()` gets short-circuited in this scenario. Here the developer does not expect that `func()` will get invoked if `lhs` is `true`.

Satyam1749: Okay, I get it. I was thinking if code is something like `TestEnum.stale_flag.isEnabled || func()` and polyglot rules convert it to `true || func()`, Should we remove `func()`.

src/cleanup_rules/swift/rules.toml
```diff
@@ -51,6 +51,49 @@ groups = ["boolean_expression_simplify"]
...
+(#eq? @true "true")
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "disjunction_expression"
+replace = "true"
+
+
+#
+# Before 
+#   abcd() || true
```
ketkarameya, line 78:
```suggestion
#   abcd || true
```

src/cleanup_rules/swift/rules.toml
```diff
@@ -51,6 +51,49 @@ groups = ["boolean_expression_simplify"]
...
+# After 
+#   true
+#
+[[rules]]
+name = "something_or_true"
```
ketkarameya, line 83:
```suggestion
name = "some_identifier_or_true"
```

ketkarameya (approved):
LGTM, minor nits in comments

PR #347, opened 2023-02-01
Added rules for boolean_simplification of something_or_false and false_or_something

Satyam1749:
Rules added:

1.  `abc || false => abc`
2. `false || abc => abc`


ketkarameya (approved):
Looks good to me. These two added rule do not short-circuit anything unnecessarily. :)

PR #348, opened 2023-02-01
Added rules for boolean_simplification of something_and_false and false_and_something

Satyam1749:
Rules added:

1.  `abc && false => false`
2. `false && abc => false`

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -52,6 +52,7 @@ groups = ["boolean_expression_simplify"]
 replace_node = "conjunction_expression"
 replace = "@lhs"
 is_seed_rule = false
+
 #
 # Before 
 #   true || abcd()
@@ -73,6 +74,8 @@ groups = ["boolean_expression_simplify"]
 replace_node = "disjunction_expression"
 replace = "true"
 is_seed_rule = false
+
+#
 #   false || abcd()
 # After 
 #   abcd()
@@ -115,6 +118,7 @@ replace_node = "disjunction_expression"
 replace = "true"
 is_seed_rule = false
 
+# Before
 #   abcd() || false
 # After 
 #   abcd()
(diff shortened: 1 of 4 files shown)


src/cleanup_rules/swift/rules.toml
```diff
@@ -94,6 +94,48 @@ groups = ["boolean_expression_simplify"]
...
+                value: (boolean_literal) @false)]
+        rhs: (_) 
+    ) @conjunction_expression
+(#eq? @false "false")
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "conjunction_expression"
+replace = "false"
+
+#
```
ketkarameya, line 118: I know we have had an offline discussion around this. But I think it might be worth noting the intuition behind why we have this `some_identifier_and_false` rule, as a comment? You can refer to the `disjunction` rule from yesterday too.

ketkarameya (commented):
Find my comment inline.

src/cleanup_rules/swift/rules.toml
```diff
@@ -131,6 +133,49 @@ query = """(
...
 replace_node = "disjunction_expression"
 replace = "@lhs"
+#
+# Before 
+#   false && abcd()
+# After 
+#   false
+#
+[[rules]]
+name = "false_and_something"
```
ketkarameya, line 143:
```suggestion
name = "false_and_anything"
```

ketkarameya: I think since we are using `(_)`  (the wild card pattern) we should call it `anything` rather than `something`. I know this is legacy naming convention. But just something i thought about when reviewing.

src/cleanup_rules/swift/rules.toml
```diff
@@ -131,6 +133,49 @@ query = """(
...
+replace_node = "conjunction_expression"
+replace = "false"
+
+#
+# Before 
+#   abcd && false
+# After 
+#   false
+#
+# f1() && false wouldn't be simplified as f1() can have some internal processing which we don't want to clean up
```
ketkarameya, line 163:
```suggestion
# Note that this rule only applies when the lhs of the conjunction is an identifier. 
# In other scenarios (i.e. is a method invocation or some other expression) we would not simplify at all, because such expressions could have side effects
```

src/cleanup_rules/swift/rules.toml
ketkarameya: We can have a separate PR which renames `something` to `anything` when necessary too.

ketkarameya (changes requested):
Some small nits.

src/cleanup_rules/swift/rules.toml
Satyam1749: I will create a seprate PR to resolve this since alot of code with `something` has landed

ketkarameya (approved):
LGTM

PR #355, opened 2023-02-05
[Refactoring] Cleanup source code unit. Move instance methods into the best matching module

ketkarameya:
(You might have opinions related to this refactoring :P )

Our `impl` for `SourceCodeUnit` was getting too big and clunky . 
I split the `impl` block into multiple `impl` blocks. I move each `impl` block to the related module. 

I think it better organizes the code.  (Though its very _un-javaish_)
* in **mod** `constraint` we implement  the associated fn`is_satisfied` and `_is_satisfied` (semantics of constraint in some sense)
* in **mod** `edit` we implement  the associated fn `get_edit` and `get_edit_for_context` (semantics of edit in some sense)
* in **mod** `piranha_argument` we specify to implement the associated fn to implement semantics of piranha arguments like `dry_run`, `delete_file_if_empty`, `delete_consecutive_lines`, ... 

I feel it keeps the type and its semantics spatially close (making it easier to think, i guess, at least for me ).
Also this is kinda idiomatic in rust (according to few cyber warriors on reddit)

-----
* __->__ #355


src/lib.rs
```diff
@@ -85,7 +85,7 @@ pub fn execute_piranha(piranha_arguments: &PiranhaArguments) -> Vec<PiranhaOutpu
   let source_code_units = piranha.get_updated_files();
 
   for scu in source_code_units.iter() {
-    scu.persist(piranha_arguments);
```
ketkarameya, line 88: we were unnecessarily passing `piranha_arguments` (`SourceCodeUnit` already has `piranha_arguments`). So I updated this signature. 

This is the only signature I changed.

src/models/constraint.rs
```diff
@@ -31,3 +40,63 @@ impl Constraint {
     Self { matcher, queries }
   }
 }
+
+// Implements instance methods related to applying a constraint
+impl SourceCodeUnit {
+  pub(crate) fn is_satisfied(
```
lazaroclapp, line 46: Ok, other than `pub(crate)`, this is the same code that was in `src_code_unit.rs` before, already under `impl SourceCodeUnit`, right? We are basically moving stuff from an object-oriented structure to an aspect-oriented structure (not exactly, but so to speak). Is my understanding here correct?

lazaroclapp (approved):
GitHub diffing UI could have made reviewing this a lot easier, but a couple copy pastes and manual runs of diff show this to be mostly code motion with just a few safe-looking changes 😅 

I am fine with the "trait"/"aspect" based organization here, even if it breaks class implementation across multiple modules.

src/models/constraint.rs
ketkarameya: yes your understanding is correct. It is aspect oriented in some sense (Cross cutting and separation of concern). Actually that was kinda in my mind when i was doing this refactoring. Just wanted to make the main `source_code_unit` less crowded

ketkarameya:
I will merge this stack after the i merge that larger stack

ketkarameya:
Addressed the comment here - https://github.com/uber/piranha/pull/356/commits/6915f6b075a6331440d690817306cbd01d5d027d

PR #356, opened 2023-02-05
[Refactoring] Do not use traits unnecessarily

ketkarameya:
I have no clue why was I using `PiranhaHelper` trait.  It adds indirection (documentation is near the trait declaration, and implementation is elsewhere). 
Maybe, one benefit was using it as an associated method. But I don't see the `PiranhaHelper` behavior being shared with any other struct, so I guess using it as a trait has no real benefit. Other than the fact that _associated methods_ are cool. I guess these were the initial days when I was kinda learning rust. 

Anyhoo,  this PR basically removes the traits and implements stuff as just standalone functions.

* __->__ #356
* #355

### DIFF
--- a/src/utilities/tree_sitter_utilities.rs
+++ b/src/utilities/tree_sitter_utilities.rs
@@ -24,121 +24,108 @@ use tree_sitter::{InputEdit, Node, Point, Query, QueryCapture, QueryCursor, Rang
 
 use super::eq_without_whitespace;
 
-#[rustfmt::skip]
-pub(crate) trait PiranhaHelpers {
-
-     /// Applies the query upon `self`, and gets the first match
-    /// # Arguments
-    /// * `source_code` - the corresponding source code string for the node.
-    /// * `query` - the query to be applied
-    ///
-    /// # Returns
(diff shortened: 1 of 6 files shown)


src/utilities/tree_sitter_utilities.rs
```diff
@@ -24,121 +24,108 @@ use tree_sitter::{InputEdit, Node, Point, Query, QueryCapture, QueryCursor, Rang
...
+  output.sort_by(|a, b| a.range().start_byte.cmp(&b.range().start_byte));
+  output.reverse();
+  output
+}
 
-  fn get_query_capture_groups(
-    &self, source_code: &str, query: &Query,
-  ) -> HashMap<Range, Vec<Vec<QueryCapture>>> {
-    let mut cursor = QueryCursor::new();
+/// Applies the query upon `self`, and gets the first match
```
lazaroclapp, line 99: Same as the two above.

ketkarameya:
Addressed the comments here - https://github.com/uber/piranha/pull/356/commits/6915f6b075a6331440d690817306cbd01d5d027d

lazaroclapp (approved):
LGTM!

Issue #358, opened 2023-02-07
Fresh CLI build not working with demo examples


conormurray95:
**Summary**
I pulled the latest piranha codebase and attempted to compile and run the CLI binary against the `feature_flag_cleanup/java` example and couldn't get it to work using the documented inputs. 

**Steps to reproduce**
1. Added a piranha_arguments.toml file in `demo/feature_flag_cleanup/java/configurations` that looks like this:
```
language = ["java"]
substitutions = [
    ["stale_flag_name", "SAMPLE_STALE_FLAG"],
    ["treated", "true"]
]
```
2. I added a Dockerfile that had the required dependencies of python and rust in the base directory of the repo. If you have a working local setup steps 2-7 can likely be skipped but I'll leave them here for ease of reproduction:
```
FROM python:3.9.16

COPY . ./piranha

# Install curl
RUN \
    apt update && \
    apt upgrade -y && \
    apt install -y curl


# Install rust \
RUN \
    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > install_rust.sh && \
    sh install_rust.sh -y && \
    . "$HOME/.cargo/env"
```

3. Build dockerfile `docker build -t piranha:latest -f ./Dockerfile .`
4. Run docker image `docker run -d -t piranha:latest`
5. `docker exec -it ${CONTAINER_ID} /bin/bash`
6. `cd piranha/`
7. Build binary:  `cargo build --release`
8. Run binary: `target/release/polyglot_piranha --path-to-codebase demo/feature_flag_cleanup/java/ --path-to-configurations demo/feature_flag_cleanup/java/configurations/`

**Error**
RUST_BACKTRACE=1 target/release/polyglot_piranha --path-to-codebase demo/feature_flag_cleanup/java/ --path-to-configurations demo/feature_flag_cleanup/java/configurations/
```
thread 'main' panicked at 'Language not supported', src/models/language.rs:187:12
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fc594f15669680fa70d255faec3ca3fb507c3405/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/fc594f15669680fa70d255faec3ca3fb507c3405/library/core/src/panicking.rs:64:14
   2: <polyglot_piranha::models::language::PiranhaLanguage as core::convert::From<&str>>::from
   3: polyglot_piranha::models::piranha_arguments::PiranhaArguments::merge
[...]

PR #359, opened 2023-02-07
Fix the CLI problem

ketkarameya:
(This looks like a large one, but real change is to just one file `piranha_arguments.rs`, most files have changes like adding a derive, updating api name, updating imports).

**Problem** #358 
**Background**  Few days ago we had introduced the new `piranha_arguments!` macro which we used to construct PiranhaArguments via the Rust API and the Python API. Using this macro (or builder) is the standard way of constructing PiranhaArguments. Note using this macro "enriches" piranha arguments i.e. it initializes the `input_substitutions` and `piranha_language` field. 
 However, our command line entry point did not use it.  Currently our command line simply parses the piranha arguments. 

**Solution** Introduce a `from_cli` method (equivalent to [py_new](https://github.com/uber/piranha/blob/master/src/models/piranha_arguments.rs#LL159C7-L159C7)) which basically invokes the `piranha_arguments!` macro.  
When doing this cleanup I observed that I had two redundant fields in `PiranhaArguments` - `substitutions:HashMap` and `language:str`. I could directly parse the command line arguments into `PiranhaLanguage` and a `Vec<(String, String)>` directly. So I did that. Basically I merged the attributes `language` and `piranha_language`, and `substitutions` and `input_substitutions`.  (Actually the more I am using rust, I am realizing its capabilities. This is the way things should have been from day 1, but i didnt know :| and things were too intimidating back then to do it very correctly ) 

**Test Plan** Earlier we were not testing the CLI interface at all :) (Hence this bug :|) . Added a test which runs the cli interface too.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2022 Uber Technologies, Inc.
+ Copyright (c) 2022 Uber Technologies, Inc.
 
  <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  except in compliance with the License. You may obtain a copy of the License at
@@ -14,11 +14,9 @@ Copyright (c) 2022 Uber Technologies, Inc.
 //! Defines the entry-point for Piranha.
 use std::{fs, time::Instant};
 
-use clap::Parser;
 use log::{debug, info};
 use polyglot_piranha::{
-  execute_piranha,
-  models::piranha_arguments::{PiranhaArguments, PiranhaArgumentsBuilder},
+  execute_piranha, models::piranha_arguments::PiranhaArguments,
   models::piranha_output::PiranhaOutputSummary,
 };
 
@@ -28,7 +26,7 @@ fn main() {
 
   info!("Executing Polyglot Piranha");
 
-  let args = PiranhaArguments::parse().merge(PiranhaArgumentsBuilder::default().build());
+  let args = PiranhaArguments::from_cli();
 
   debug!("Piranha Arguments are \n{:#?}", args);
   let piranha_output_summaries = execute_piranha(&args);
--- a/src/models/language.rs
+++ b/src/models/language.rs
@@ -11,6 +11,8 @@
  limitations under the License.
 */
 
+use std::str::FromStr;
+
 use getset::Getters;
 use serde_derive::Deserialize;
 use tree_sitter::{Parser, Query};
@@ -51,13 +53,6 @@ pub enum SupportedLanguage {
   Ts,
   Tsx,
   Python,
-}
-
-impl std::str::FromStr for PiranhaLanguage {
-  type Err = String;
-  fn from_str(language: &str) -> Result<Self, Self::Err> {
-    Ok(PiranhaLanguage::from(language))
-  }
 }
 
 impl Default for SupportedLanguage {
@@ -105,11 +100,18 @@ impl Default for PiranhaLanguage {
 
 impl From<&str> for PiranhaLanguage {
   fn from(language: &str) -> Self {
+    PiranhaLanguage::from_str(language).unwrap()
+  }
+}
+
+impl std::str::FromStr for PiranhaLanguage {
+  type Err = &'static str;
+  fn from_str(language: &str) -> Result<Self, Self::Err> {
     match language {
       JAVA => {
         let rules: Rules = parse_toml(include_str!("../cleanup_rules/java/rules.toml"));
         let edges: Edges = parse_toml(include_str!("../cleanup_rules/java/edges.toml"));
-        Self {
+        Ok(Self {
           name: language.to_string(),
           supported_language: SupportedLanguage::Java,
           language: tree_sitter_java::language(),
@@ -121,12 +123,12 @@ impl From<&str> for PiranhaLanguage {
           .scopes()
           .to_vec(),
           comment_nodes: vec!["line_comment".to_string(), "block_comment".to_string()],
-        }
+        })
       }
       GO => {
         let rules: Rules = parse_toml(include_str!("../cleanup_rules/go/rules.toml"));
         let edges: Edges = parse_toml(include_str!("../cleanup_rules/go/edges.toml"));
-        PiranhaLanguage {
+        Ok(PiranhaLanguage {
           name: language.to_string(),
           supported_language: SupportedLanguage::Go,
           language: tree_sitter_go::language(),
@@ -136,12 +138,12 @@ impl From<&str> for PiranhaLanguage {
             .scopes()
             .to_vec(),
           comment_nodes: vec!["comment".to_string()],
-        }
+        })
       }
       KOTLIN => {
         let rules: Rules = parse_toml(include_str!("../cleanup_rules/kt/rules.toml"));
         let edges: Edges = parse_toml(include_str!("../cleanup_rules/kt/edges.toml"));
-        PiranhaLanguage {
+        Ok(PiranhaLanguage {
           name: language.to_string(),
           supported_language: SupportedLanguage::Kotlin,
           language: tree_sitter_kotlin::language(),
@@ -151,9 +153,9 @@ impl From<&str> for PiranhaLanguage {
             .scopes()
             .to_vec(),
           comment_nodes: vec!["comment".to_string()],
-        }
+        })
       }
-      PYTHON => PiranhaLanguage {
+      PYTHON => Ok(PiranhaLanguage {
         name: language.to_string(),
         supported_language: SupportedLanguage::Python,
         language: tree_sitter_python::language(),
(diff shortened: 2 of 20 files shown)


src/main.rs
```diff
@@ -28,7 +26,7 @@ fn main() {
 
   info!("Executing Polyglot Piranha");
 
-  let args = PiranhaArguments::parse().merge(PiranhaArgumentsBuilder::default().build());
```
ketkarameya, line 31: The "bug fix"

src/models/piranha_arguments.rs
```diff
@@ -46,23 +46,14 @@ pub struct PiranhaArguments {
...
   path_to_codebase: String,
 
-  // Input arguments provided to Piranha, mapped to tag names -
-  // @stale_flag_name, @namespace, @treated, @treated_complement
-  // These substitutions instantiate the initial set of feature flag rules
-  #[get = "pub"]
-  #[builder(default = "default_input_substitutions()")]
-  #[clap(skip)]
-  #[serde(skip)]
-  input_substitutions: HashMap<String, String>,
```
ketkarameya, line 56: Merged this into `substitutions`

src/models/piranha_arguments.rs
```diff
@@ -46,23 +46,14 @@ pub struct PiranhaArguments {
...
-
-  // Substitutions to instantiate the initial set of feature flag rules
+  /// These substitutions instantiate the initial set of rules.
+  /// Usage : -s stale_flag_name=SOME_FLAG -s namespace=SOME_NS1
   #[builder(default = "default_substitutions()")]
-  #[clap(skip)]
+  #[clap(short= 's',value_parser = parse_key_val)]
   #[serde(default = "default_substitutions")]
-  substitutions: Vec<Vec<String>>,
```
ketkarameya, line 62: Change type so that clap parses it as key-value

src/models/piranha_arguments.rs
```diff
@@ -75,18 +66,13 @@ pub struct PiranhaArguments {
   #[clap(short = 'j', long)]
   #[serde(skip)]
   path_to_output_summary: Option<String>,
-
-  // the target language
-  #[builder(default = "default_language()")]
-  #[clap(skip)]
-  #[serde(default = "default_language")]
-  language: String,
```
ketkarameya, line 83: merge into `piranha_language` (below)

src/models/piranha_arguments.rs
```diff
@@ -278,29 +284,14 @@ impl PiranhaArgumentsBuilder {
   /// * merge the two PiranhaArguments
   pub fn build(&self) -> PiranhaArguments {
     let _arg = &self.create().unwrap();
-    let input_substitutions = _arg
```
ketkarameya, line 281: Now no need for this hackery :) yayyy!

src/tests/test_piranha_python.rs
```diff
@@ -11,18 +11,47 @@ Copyright (c) 2022 Uber Technologies, Inc.
...
+use tempdir::TempDir;
+
+use super::create_match_tests;
+
+use crate::{
+  models::{default_configs::PYTHON, piranha_output::PiranhaOutputSummary},
+  utilities::{eq_without_whitespace, read_file},
+};
+
+#[test]
```
ketkarameya, line 26: An equivalent test case that uses CLI

src/models/piranha_output.rs
```diff
@@ -14,15 +14,15 @@ Copyright (c) 2022 Uber Technologies, Inc.
...
+use serde_derive::{Deserialize, Serialize};
 
 use crate::utilities::gen_py_str_methods;
 
 use super::{edit::Edit, matches::Match, source_code_unit::SourceCodeUnit};
 use pyo3::{prelude::pyclass, pymethods};
 
 /// A class to represent Piranha's output
-#[derive(Serialize, Debug, Clone, Default)]
+#[derive(Serialize, Debug, Clone, Default, Deserialize)]
```
ketkarameya, line 25: Had to add `Deserialize` for `PiranhaOutPutSummary` since in the new test case for `cli`. In this test we parse the output json (which is a `Vec<PiranhaOutputSummary>`) and compare with expected. Parsing the output json requires us to derive `Deserialize` for `PiranhaOutputSummary`

ketkarameya (commented):
Annotated.

src/tests/test_piranha_java.rs
```diff
@@ -79,7 +81,7 @@ fn test_scenarios_find_and_propagate_panic() {
   let piranha_arguments = piranha_arguments! {
     path_to_codebase = path_to_codebase,
     path_to_configurations = path_to_configurations,
-    language= JAVA.to_string(),
+    language= PiranhaLanguage::from(JAVA),
```
lazaroclapp, line 84:
```suggestion
    language = PiranhaLanguage::from(JAVA),
```

src/tests/test_piranha_python.rs
lazaroclapp: Equivalent to what? (Also, maybe that should be a comment in the source rather than on the PR?)

lazaroclapp (commented):
Minor nits and questions, but overall this fix looks pretty cool to me! (Though I am probably still missing some context on the inner-workings of Rust attributes, ie. `Deserialize`)

src/models/piranha_arguments.rs
```diff
@@ -11,19 +11,19 @@ Copyright (c) 2022 Uber Technologies, Inc.
...
-    default_piranha_language, default_substitutions,
+    default_global_tag_prefix, default_name_of_piranha_argument_toml,
+    default_number_of_ancestors_in_parent_scope, default_path_to_codebase,
+    default_path_to_configurations, default_path_to_output_summaries, default_piranha_language,
+    default_substitutions, GO, JAVA, KOTLIN, PYTHON, SWIFT, TSX, TYPESCRIPT,
   },
   language::PiranhaLanguage,
 };
+use crate::utilities::{parse_key_val, read_toml};
+use clap::builder::TypedValueParser as _;
```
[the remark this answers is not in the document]
ketkarameya: Removed. It was part of the tutorial, I copy pasted (didnt update) [Ref](https://doc.rust-lang.org/reference/items/use-declarations.html#:~:text=Underscore%20Imports,may%20conflict%20with%20another%20symbol). They do this so that when folks copy paste into their snippets in their own code base they dont get namespace conflicts

src/models/piranha_arguments.rs
```diff
@@ -229,7 +215,25 @@ impl PiranhaArguments {
...
+
+  pub fn from_cli() -> Self {
+    let p = PiranhaArguments::parse();
+    piranha_arguments! {
+      path_to_codebase= p.path_to_codebase().to_string(),
+      substitutions = p.substitutions.clone(),
+      language = p.language().clone(),
+      path_to_configurations = p.path_to_configurations().to_string(),
+      path_to_output_summary = p.path_to_output_summary().clone(),
+      delete_file_if_empty= *p.delete_file_if_empty(),
```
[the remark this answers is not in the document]
ketkarameya: I will make a widespread change later which makes this consistent for the entire repository. I think i am gonna look into a pre-commit hook for this . :P

ketkarameya (commented):
Addressed the comments.

src/models/language.rs
```diff
@@ -98,11 +100,20 @@ impl Default for PiranhaLanguage {
 
 impl From<&str> for PiranhaLanguage {
   fn from(language: &str) -> Self {
+    PiranhaLanguage::from_str(language).unwrap()
```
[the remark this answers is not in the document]
ketkarameya: Ok. So the deal is that earlier I I was implementing this `From<&str>` trait, and was simply panic-ing when language was not supported. 

Upon adding clap (Command line argument parsing) support for `language: PiranhaLanguage`, I needed to implement the `trait std::str::FromStr` (because `clap` requires me to do , since invokes it to parse the input string into `PiranhaLanguage`). Note, `FromStr` has a return type `Result<X,Err>` while `From` has a return type `X`. 

So I moved the implementation of `From` into `FromStr`. 
Then I call `FromStr` with unsafe unwrap (inside `From`). When the language name is not supported we just panic.

Alternatively, I could remove the implemetation of `From` trait  and update all usages to `FromStr`. But this will involve handling `Result<PiranhaLanguage, Err>` at each callee of `FromStr`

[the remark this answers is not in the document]
lazaroclapp: Shouldn't we return a graceful error to the user and a CLI help description of how to use the tool when they pass an unknown `--language`? It sounds from the above that we would just panic in that case, which makes sense when building from the python/rust API, but probably not graceful enough for CLI input...

src/models/piranha_arguments.rs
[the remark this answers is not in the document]
lazaroclapp: Ok, do we have any sort of check about unused imports / import order and formatting in Rust? I see a lot of bikeshedding on these import blocks (a fix for a different PR at any rate, just think aloud here).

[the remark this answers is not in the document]
ketkarameya: Actually, rust clippy takes care of unused imports. It is not aggressive about ordering the imports , but it has many settings which can fix the things you suggest. I will look into it (on some slow afternoon :P )

src/models/language.rs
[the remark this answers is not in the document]
lazaroclapp: That's pretty cool, actually 🙂

PR #360, opened 2023-02-09
`piranha_rule!` macro for creating a rule

ketkarameya:
**Background:**  Currently we use `new(..)` to instantiate `Rule` and `Constraint` which is messy. We want to add the Builder pattern to build it, and then wrap it with a macro. 
**Why?** We ultimately want to move `RuleGraph` into `PiranhaArguments`. Which implies that we will have to expose rules, constraints to the user.  So rather than doing that move and then retroactively fixing stuff, I am doing it the other way.
**Implementation:** Basically the `Rule::new()` and `Constraint::new()` are replaced by the macros `piranha_rule` and `constraint` respectively
**Test Plan:** Updated the tests to use the macro. No semantic change. 

(Trying to break down the change into smaller diffs)

### DIFF
--- a/src/models/unit_tests/rule_test.rs
+++ b/src/models/unit_tests/rule_test.rs
@@ -10,15 +10,16 @@ Copyright (c) 2022 Uber Technologies, Inc.
  express or implied. See the License for the specific language governing permissions and
  limitations under the License.
 */
-use crate::models::{default_configs::JAVA, language::PiranhaLanguage};
+use crate::{
+  constraint,
+  models::{default_configs::JAVA, language::PiranhaLanguage},
+  utilities::eq_without_whitespace,
+};
 use std::collections::HashSet;
 
 use super::InstantiatedRule;
 use {
-  super::Rule,
-  crate::models::{
-    constraint::Constraint, rule_store::RuleStore, source_code_unit::SourceCodeUnit,
-  },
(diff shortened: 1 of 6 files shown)


src/models/unit_tests/rule_test.rs
```diff
@@ -10,42 +10,56 @@ Copyright (c) 2022 Uber Technologies, Inc.
...
+  let rule = piranha_rule! {
+      name= "test",
+      query= "(
+        ((assignment_expression left: (_) @a.lhs right: (_) @a.rhs) @abc)
+        (#eq? @a.lhs \"@variable_name\")
+      )",
+      replace_node = "abc",
+      replace = "",
+      holes = ["variable_name"]
+  };
```
lazaroclapp, line 39: Ok, the actual macro definition is a bit clunky, but I have to admit this is **way** easier to read than the `Rule::new` example.

Any reason why using the builder directly is not a good option? Less idiomatic in Rust?

lazaroclapp (approved):
Refactoring makes sense to me. See question about macros vs builders, but I am fine with the macro approach.

src/models/unit_tests/rule_test.rs
ketkarameya: I think builders is as idomatic as it could get. I feel it gets too verbose. But folks can always use Builder if they want to. Its available for them.

PR #362, opened 2023-02-10
Move PiranhaArguments from RuleStore into Piranha

ketkarameya:
Moves `piranha_arguments: PiranhaArguments` from `RuleStore` to `Piranha` (the sort of main runner class).

--------
* #369
* #364
* #363
* __->__ #362

### DIFF
--- a/src/models/rule_store.rs
+++ b/src/models/rule_store.rs
@@ -25,7 +25,7 @@ use regex::Regex;
 use tree_sitter::Query;
 
 use crate::{
-  models::piranha_arguments::{PiranhaArguments, PiranhaArgumentsBuilder},
+  models::piranha_arguments::PiranhaArguments,
   models::{
     rule_graph::{RuleGraph, RuleGraphBuilder},
     scopes::ScopeQueryGenerator,
@@ -34,6 +34,7 @@ use crate::{
 };
 
 use super::{
+  language::PiranhaLanguage,
   outgoing_edges::Edges,
   rule::{InstantiatedRule, Rules},
 };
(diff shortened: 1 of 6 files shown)


src/models/rule_store.rs
```diff
@@ -34,14 +34,15 @@ use crate::{
...
+  language::PiranhaLanguage,
   outgoing_edges::Edges,
   rule::{InstantiatedRule, Rules},
 };
 
 pub(crate) static GLOBAL: &str = "Global";
 pub(crate) static PARENT: &str = "Parent";
 /// This maintains the state for Piranha.
-#[derive(Debug, Getters)]
+#[derive(Debug, Getters, Default)]
```
lazaroclapp, line 45: Just to make sure I get this right: Before this change we needed some custom logic on our `Default` implementation for `RuleStore` (namely the `PiranhaArgumentsBuilder::default().build()` line) and thus we couldn't use `#[derive(Default)]` here, but now we can? 

Is this correct? I see `PiranhaArguments` already has `#[derive(Default)]` on master. What about the main `Piranha` struct? No need to change that since there isn't a `Default` implementation?

lazaroclapp (approved):
Looks like an straightforward refactoring to me. One question about Rust below, though 🙂

src/models/rule_store.rs
ketkarameya: Yes. You are correct about why we do not require to `impl Default for RuleStore`  anymore because we do not need `PiranhaArgumentsBuilder::default().build()` (since we removed `PiranhaArguments` from `RuleStore`).

That's an awesome catch. While `Piranha` doesnt need `PiranhaArguments`'s `Default` trait, but the function `read_toml` relies on the `PiranhaArguments::default()` when the `piranha_arguments.toml` file is not provided, but we expect one.  Its more like a error scenario. ( I dont see that happening anymore), since we can fully compose `piranha_aguments.toml` from commandline/api now. 

But for consistency, and future bug proofing I have removed the `#[derive(Default)]` from `PiranhaArguments`  and implemented the `Default` for it by hand.

lazaroclapp (approved):
LGTM!

PR #363, opened 2023-02-10
Move RuleGraph from RuleStore to PiranhaArguments

ketkarameya:
* (to be filled)

### DIFF
--- a/src/models/language.rs
+++ b/src/models/language.rs
@@ -86,6 +86,13 @@ impl PiranhaLanguage {
     parser
   }
 
+  pub(crate) fn can_parse(&self, de: &jwalk::DirEntry<((), ())>) -> bool {
+    de.path()
+      .extension()
+      .and_then(|e| e.to_str().filter(|x| x.eq(&self.name())))
+      .is_some()
+  }
+
   #[cfg(test)]
   pub(crate) fn set_scopes(&mut self, scopes: Vec<ScopeGenerator>) {
     self.scopes = scopes;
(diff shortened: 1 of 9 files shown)


src/models/language.rs
```diff
@@ -86,6 +86,13 @@ impl PiranhaLanguage {
     parser
   }
 
+  pub(crate) fn can_parse(&self, de: &jwalk::DirEntry<((), ())>) -> bool {
+    de.path()
+      .extension()
+      .and_then(|e| e.to_str().filter(|x| x.eq(&self.name())))
```
lazaroclapp, line 92: Probably a nit for a separate PR, but... are using the field `name` to refer to the string representation of the language (i.e. `"kotlin"`) or the extension (i.e. `"kt")? Seems like a weird choice for "name" to be the second...

src/models/piranha_arguments.rs
```diff
@@ -283,14 +292,31 @@ impl PiranhaArgumentsBuilder {
...
 
-    let path_to_toml =
+    let rule_graph = built_in_rules.merge(&user_defined_rules);
+
+    let arg_rg = PiranhaArgumentsBuilder::default()
+      .rule_graph(rule_graph)
+      .create()
+      .unwrap();
+
+    _arg = _arg.merge(arg_rg);
```
lazaroclapp, line 311: Not knowledgable enough about Rust to know if this is idiomatic, but this feels like an awkward way to just pass the `rule_graph` to the `PiranhaArguments` being built. Its creating a different instance of `PiranhaArguments` with partial configuration and then sort of combining it into this one, no?

ketkarameya: Addressed

lazaroclapp (approved):

PR #366, opened 2023-02-13
Create a wrapper around strings carrying Tree-sitter queries

ketkarameya:
**Problem:**
I was casually reviewing the code and I noticed that its very confusing when it comes to determining whether a string carries a tree-sitter query or not. Currently we rely upon the variable names 😢  This can also lead to 🐛 .
For instance, when we replace tags with values inside strings, we perform a simple string replacement.
However, when we replace tags with values inside tree-sitter-queries we need to escape it `\n` -> `\\n`. 

**Solution:**
Create a wrapper _new type_ `TSQuery` around a String.
I also took this opportunity to extract  this `substitute_tag` behavior into a `Instantiate` trait. 
I implemented it for `String`, `TSQuery`, `Rule` and `Constraint`.

(Passing that `boolean` based on whether a string is a tree-sitter query or not was a bit "Risky")

-----
* #380
* #379
* #373
* #372
* #371
* #367
* __->__ #366


src/models/rule_store.rs
```diff
@@ -70,18 +70,18 @@ impl RuleStore {
       r.name().eq(&rule.name()) && r.replace().eq(&rule.replace()) && r.query().eq(&rule.query())
     }) {
       #[rustfmt::skip]
-      debug!("{}", format!("Added Global Rule : {:?} - {}", r.name(), r.query()).bright_blue());
+      debug!("{}", format!("Added Global Rule : {:?} - {}", r.name(), r.query().0).bright_blue());
```
lazaroclapp, line 73: As mentioned in a different comment, do we want to be referencing `.0` or encapsulate this in a meaningfully named method?

lazaroclapp (commented):
Overall makes sense, like the type encapsulation. Some nits below!

src/models/rule.rs
```diff
@@ -49,8 +50,7 @@ pub struct Rule {
   #[builder(default = "default_query()")]
   #[serde(default = "default_query")]
   #[get = "pub"]
-  #[pyo3(get)]
-  query: String,
+  query: TSQuery,
```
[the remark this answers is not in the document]
ketkarameya: It was a mistake. I didnt mean to do this change, it got deleted during one of the conflicts i was resolving :|

ketkarameya (commented):
Addressed the comments in this PR. [here](https://github.com/uber/piranha/pull/366/commits/3dbd8a9ca9a11827887547ad6cb2e544339d98fe)

src/models/rule.rs
```diff
@@ -247,6 +233,22 @@ impl InstantiatedRule {
   }
 }
 
+impl Instantiate for Rule {
+  /// Create a new query from `self` by updating the `query` and `replace` based on the substitutions.
+  fn instantiate(&self, substitutions_for_holes: &HashMap<String, String>) -> Rule {
+    if substitutions_for_holes.len() != self.holes().len() {
```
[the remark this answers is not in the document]
lazaroclapp: ^ Any update on this one?

src/models/rule.rs
```diff
@@ -236,11 +240,9 @@ impl InstantiatedRule {
 
 impl Instantiate for Rule {
   /// Create a new query from `self` by updating the `query` and `replace` based on the substitutions.
+  /// This functions assumes that each hole in the rule can be substituted .
```
lazaroclapp, line 243:
```suggestion
  /// This functions assumes that each hole in the rule can be substituted.
```

src/models/rule.rs
```diff
@@ -198,11 +198,15 @@ pub(crate) struct InstantiatedRule {
 
 impl InstantiatedRule {
   pub(crate) fn new(rule: &Rule, substitutions: &HashMap<String, String>) -> Self {
-    let substitutions_for_holes = rule
+    let substitutions_for_holes: HashMap<String, String> = rule
       .holes()
       .iter()
       .filter_map(|h| substitutions.get(h).map(|s| (h.to_string(), s.to_string())))
       .collect();
+    if substitutions_for_holes.len() != rule.holes().len() {
```
lazaroclapp, line 206: Again, question, what happens if the lengths match, but not the keys? Wouldn't you still end with a partially instantiated rule? Do we want to prevent this/check for it?

ketkarameya: As `substitutions`  we pass the entire substitution map we have in the "state". 
Then in this statement: 
```
let substitutions_for_holes: HashMap<String, String> = rule
      .holes()
      .iter()
      .filter_map(|h| substitutions.get(h).map(|s| (h.to_string(), s.to_string())))
      .collect();
```
I iterate over each hole and get its value from `substitutions` parameter map if possible, and collect. 
So when lengths don't match, it implies that we did not find all substitutions. 
If lengths match, implies We found a substitution for each hole in the rule (note we iterate over `rule.holes())
So to answer your question it will never happen that rule is partially instantiated though the lengths match

ketkarameya (commented):
Fixed the typo
On a side note, ghstack is not environment friendly. 
I made a small comment change in the first commit of the stack, and it has triggered build for each PR. :| 
(Though it can't be avoided because of the "paradigm" it uses)

src/models/rule.rs
lazaroclapp: Ok, I get it now. Fair enough!

I do feel this logic is complex enough that it merits a comment in the code. Probably not the full long explanation you just gave, but a sentence or two. Maybe:

```
// filter_map discards any element of `rules.holes()` for which there isn't a valid substitution,
// so checking that the lengths match is enough to verify all holes have a matching substitution.
```

As a side note, what goes Rust do on `m.get(k).map(...)` where `k` is missing from `m`? My intuition from Java/Go is that this would be an NPE/nil panic because `m.get(k)` returns `null`/`nil`. But presumably that isn't the case in Rust?

Do we have a unit test that expects this panic? I don't immediately see a test case for invalid instantiation in this PR, but not sure if we already had it before.

src/models/rule.rs
```diff
@@ -203,6 +203,8 @@ impl InstantiatedRule {
       .iter()
       .filter_map(|h| substitutions.get(h).map(|s| (h.to_string(), s.to_string())))
       .collect();
+    // Since filter_map (above) discards any element of `rules.holes()` for which there isn't a valid substitution,
+    // Checking that the lengths match is enough to verify all holes have a matching substitution.
```
lazaroclapp, line 207:
```suggestion
    // checking that the lengths match is enough to verify all holes have a matching substitution.
```

No need for upper case after a comma.

test-resources/java/find_and_propagate_invalid_substitutions/configurations/edges.toml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 1:
```suggestion
# Copyright (c) 2023 Uber Technologies, Inc.
```

test-resources/java/find_and_propagate_invalid_substitutions/configurations/rules.toml
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 1:
```suggestion
# Copyright (c) 2023 Uber Technologies, Inc.
```

test-resources/java/find_and_propagate_invalid_substitutions/expected/TestClass.java
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 2:
```suggestion
 * Copyright (c) 2023 Uber Technologies, Inc.
```

test-resources/java/find_and_propagate_invalid_substitutions/input/TestClass.java
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 2:
```suggestion
 * Copyright (c) 2023 Uber Technologies, Inc.
```

test-resources/java/find_and_propagate_invalid_substitutions/input/TestSubClass.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 2:
```suggestion
 * Copyright (c) 2023 Uber Technologies, Inc.
```

lazaroclapp (approved):
Spell check level nits at this point, good to land once those suggestions are added 🙂

PR #367, opened 2023-02-13
Fully qualify types inside the `piranha_rule` macro

ketkarameya:
Fairly straightforward refactoring.  Using `HashSet<T>` requires `T: Hash` which is an unnecessary constraint. 
Let's not use `HashSet` when using `Vec` is **ok**.

------

I have updated this PR. I have undone the previous changes, and kept some random cleanups (but sensible) for sake of not tinkering with the stack.

The change in this PR, fully qualifies `HashSet` in the `piranha_rule!` macro

----
* #380
* #379
* #373
* #372
* #371
* __->__ #367
* #366


lazaroclapp:
Ok, high level question here before I do a full review: are these fields for which we want to allow repetitions? My feeling is that if we want uniqueness, and don't care about iteration order, then hash sets are a better fit than vectors/lists. Is there anything I am missing about why having these types be hashable is a problem?

ketkarameya:
Hmmm. You are correct, we are language agnostic and repetition agnostic too. But obviously we would prefer no repetition (keeps rules cleaner). There is not problem with Hashable its just another type constraint to add to our struct which is already already stocked a lot of `@derives`. But I get your point. I just feel as a user vec/list is more intuitive to use than set. But yes, using `set` is not unintuitive in any sense.

lazaroclapp (approved):
Going to defer to you here on what the right data structure for all of these is. 

I do wonder what happens if the same rule/query/constraint is added multiple times by mistake (e.g. programmatically from a python client). I assume the case of a rule clashing against itself is very similar to that of two rules matching a subset of each other ASTs, so nothing should break for that case in particular. Right?

I think my general thinking is that it's safer to use sets than it's to use lists in general, except when you explicitly want to allow repetition (or care a lot about performance of iteration, but linked hashsets exist).

lazaroclapp (approved):
Code change makes sense to me, but please give the PR a reasonable title and description to be used for the commit/changelog/etc. Even if it is something like "Cleanup: Expand import in `piranha_rule` macro" (or whatever the idiomatic way to say this is in Rust)

PR #368, opened 2023-02-13
Added rules for if cleanup in swift

Satyam1749:
Rules created to cleanup if statements which are always true.
Rule 1:
```
Before 
if true {
   f1()
}
After 
{
   f1()
}
```

Rule 2:
```
Before
{
   f1()
}
After
f1()
```

Rule 2 will always be child of Rule 1, Since output of Rule 1 is not compilable in swift. This is done to make sure else if case is handled. Example of else if case: 
```
Before:

if var {
   f1()
} else if true {
   f2()
} else {
   f3()
}

After:

if var {
  f1()
} else {
  f2()
}
```

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -231,6 +231,59 @@ replace_node = "not_expression"
 replace = "true"
 is_seed_rule = false
 
+#
+# Next two rules take care of if-else cleanup, these 2 rules depend on their order in this file
+# 
+
+#
+# Before 
+#   if true {
+#     abcd()
+#   } else {
+#     hij()
+#   }
+# After 
+#   abcd()
+#
+[[rules]]
+name = "if_always_true"
+query = """ (
+(if_statement
+    condition: (boolean_literal) @condition_literal
+    .(statements) @if_block
+    ) @if_else_block
+(#eq? @condition_literal "true")
+)"""
+groups = ["if_cleanup"]
+replace_node = "if_else_block"
+replace = "@if_block"
+is_seed_rule = false
+
+# Before 
+#   else if true {
+#     abcd()
+#   } else {
+#     hij()
+#   }
+# After 
+#   abcd()
+#
+[[rules]]
+name = "else_if_always_true"
+query = """ (
+(else)
+(if_statement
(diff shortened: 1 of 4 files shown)


ketkarameya:
Please add a description and issue for this PR

ketkarameya (changes requested):
Please add a description and issue for this PR

ketkarameya:
Conflict to be resolved.

Cargo.toml
```diff
@@ -51,7 +51,8 @@ serde_json = "1.0.82"
 # TODO: Update after https://github.com/fwcd/tree-sitter-kotlin/pull/71 lands
 tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git", rev = "a87ddd003368e068563f1cc478a1b2a3f9d73b60" }
 tree-sitter-java = "0.20.0"
-tree-sitter-swift = "0.3.4"
+# TODO: Update after: https://github.com/alex-pinkus/tree-sitter-swift/issues/278 resolves
```
ketkarameya, line 54: should be rebased

src/cleanup_rules/swift/rules.toml
```diff
@@ -135,6 +136,54 @@ replace_node = "disjunction_expression"
 replace = "@lhs"
 is_seed_rule = false
 
+# Before
```
ketkarameya, line 139: Can we note the intuition here that  `if_always_true` is assumed to run before `else_if_always_true`

test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -37,9 +36,21 @@ class SampleClass {
...
+        isEnabled = v2 || placeholder_false 
+    }
+    func checkIfTrueCleanup() {
+        f1()
+        if TestEnum.stale_flag_one.isEnabled {
+            f2()
+        }
+
+        if isEnabled {
+            f2()
```
ketkarameya, line 51: can we have one more longer `if-else` ladder? Can we have isEnabled somewhere in the middle of the ladder?

ketkarameya (changes requested):

ketkarameya (approved):
LGTM.

PR #364, opened 2023-02-10
Expose Rust API for constructing a `RuleGraph`

ketkarameya:
Expose Rust API for constructing the `RuleGraph`.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #369
* __->__ #364
* #363
* #362

PR #369, opened 2023-02-13
Expose the Python API for RuleGraph construction

ketkarameya:
* __->__ #369
* #364
* #363
* #362

### DIFF
--- a/src/models/constraint.rs
+++ b/src/models/constraint.rs
@@ -11,23 +11,38 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
+use super::default_configs::{default_matcher, default_queries};
 use derive_builder::Builder;
 use getset::Getters;
+use pyo3::prelude::*;
 use serde_derive::Deserialize;
 
-use super::default_configs::{default_matcher, default_queries};
-
 #[derive(Deserialize, Debug, Clone, Hash, PartialEq, Eq, Getters, Builder)]
+#[pyclass]
 pub struct Constraint {
   /// Scope in which the constraint query has to be applied
   #[builder(default = "default_matcher()")]
   #[get = "pub"]
+  #[pyo3(get)]
   matcher: String,
   /// The Tree-sitter queries that need to be applied in the `matcher` scope
   #[builder(default = "default_queries()")]
   #[get = "pub"]
   #[serde(default)]
+  #[pyo3(get)]
   queries: Vec<String>,
+}
+
+#[pymethods]
+impl Constraint {
+  #[new]
+  fn py_new(matcher: String, queries: Option<Vec<String>>) -> Self {
+    ConstraintBuilder::default()
+      .matcher(matcher)
+      .queries(queries.unwrap_or_default())
+      .build()
+      .unwrap()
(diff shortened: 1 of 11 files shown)


tests/tests.py
```diff
@@ -67,6 +66,83 @@ def test_piranha_match_only():
...
+            (class_declaration name: (_)@class_name 
+                body : (class_body ((_)*) @class_members)  @class_body
+            ) @class_declaration
+            (#eq? @class_name "FooBar")
+        )
+        """,
+        replace_node="class_body",
+        replace="""{
+  private List<String> names;\n @class_members
+}""",
```
lazaroclapp, line 84: I guess one interesting caveat of the python API is that multi-line strings force us to have some somewhat awkward indentation here in order to get the right indentation in the final file. Finally one point in favor of the otherwise annoying EP varargs `lines` approach (see [here](https://github.com/uber/NullAway/blob/master/nullaway/src/test/java/com/uber/nullaway/NullAwayCoreTests.java#L74) for example)

tests/tests.py
```diff
@@ -67,6 +66,83 @@ def test_piranha_match_only():
...
+        )
+
+    args = PiranhaArguments(
+        "test-resources/java/insert_field_and_import/input",
+        "test-resources/java/insert_field_and_initializer/configurations",
+        "java",
+        {
+            "stale_flag_name": "STALE_FLAG",
+            "treated": "true",
+            "treated_complement": "false",
```
lazaroclapp, line 135: Why do we need any sort of configuration about stale flags to run the field insertion example above? Seems unrelated to flags.

src/models/constraint.rs
```diff
@@ -11,25 +11,40 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
+use super::default_configs::{default_matcher, default_queries};
 use derive_builder::Builder;
 use getset::Getters;
+use pyo3::prelude::*;
```
lazaroclapp, line 17: Not sure about Rust best practices, but for every other language our coding guidelines tend to be _against_ star imports.

src/models/outgoing_edges.rs
```diff
@@ -23,22 +23,43 @@ pub(crate) struct Edges {
 
 // Captures an entry from the `edges.toml` file.
 #[derive(Deserialize, Debug, Clone, Hash, PartialEq, Eq, Default, Getters, Builder)]
+#[pyclass]
 pub struct OutgoingEdges {
+  /// The source rule or group of rules
   #[get = "pub with_prefix"]
-  from: String,
+  #[serde(alias = "from")]
```
lazaroclapp, line 30: Interesting. This is due to `from` being a keyword in Python, right?

lazaroclapp (commented):
Some comments below, but overall makes sense to me.

ketkarameya:
Ow. I merged this by mistake. I reverted the commit( on master). I will make PR again by cherrypicking the commit.

src/models/outgoing_edges.rs
ketkarameya: Yes. :O

PR #374, opened 2023-02-16
Bumped up tree-sitter-swift dependency and resolved test failures

Satyam1749:
A new version of `tree-sitter-swift` was released to resolve issues reported by us ([issue-272](https://github.com/alex-pinkus/tree-sitter-swift/issues/272), [issue-271](https://github.com/alex-pinkus/tree-sitter-swift/issues/271)) .

The new version has an [issue](https://github.com/tree-sitter/tree-sitter-go/pull/103), which is critical for our usecase, So created a fork with fix and updated tree-sitter-swift dependency with the same.

Since the grammar for `!abc` has changed in new release, A seprate rule for `!TestEnum.stale_flag.isEnabled` is not required and this rule would be handled with `!true` and `!false` rules.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,8 @@ serde_json = "1.0.82"
 # TODO: Update after https://github.com/fwcd/tree-sitter-kotlin/pull/71 lands
 tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git", rev = "a87ddd003368e068563f1cc478a1b2a3f9d73b60" }
 tree-sitter-java = "0.20.0"
-tree-sitter-swift = "0.3.4"
+# TODO: Update after: https://github.com/alex-pinkus/tree-sitter-swift/issues/278 resolves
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -51,7 +51,7 @@ serde_json = "1.0.82"
 # TODO: Update after https://github.com/fwcd/tree-sitter-kotlin/pull/71 lands
 tree-sitter-kotlin = { git = "https://github.com/ketkarameya/tree-sitter-kotlin.git", rev = "a87ddd003368e068563f1cc478a1b2a3f9d73b60" }
 tree-sitter-java = "0.20.0"
-tree-sitter-swift = "0.3.4"
```
ketkarameya, line 54: Let's make an issue against the upstream repo and add it as a todo like we have done for kotlin/go

ketkarameya (changes requested):
please add description for your PR

ketkarameya (approved):
LGTM

PR #379, opened 2023-02-17
Remove  support for `run_piranha_cli` API 

ketkarameya:
This is a breaking change :| 

Removes support for the deprecated `run_piranha_cli` API. 

**Why?**
* I think its a lot of unnecessary and complicated code we are maintaining (just to support this API)
*  `run_piranha_cli` in some sense is incomplete and confusing to use. It allows passing some command line args, some args in tomls so on. 
* Polyglot Piranha is its infancy, so I guess we don't have any particular users too (Except us internally). Even if we have users, I guess this will be a welcomed change
* `execute_piranha` can support all the scenarios that are supported by `run_piranha_cli`

**Changes:**
* Remove `run_piranha_cli`
  * Delete the code which w as executed via `run_piranha_cli` only
* Remove `serde` support for `PiranhaArguments`(Not needed anymore)

-----
* #386
* #384
* #380
* __->__ #379


POLYGLOT_README.md
```diff
@@ -112,7 +112,7 @@ Polyglot Piranha can be used as a python library or as a command line tool.
 
 `pip install polyglot-piranha`
 
-Currently, we support one simple API (`run_piranha_cli`) that wraps the command line usage of Polyglot Piranha. We believe this makes it easy to incorporate Piranha in *"pipelining"*.
+Currently, we support one simple API (`execute_piranha`) that wraps the command line usage of Polyglot Piranha. We believe this makes it easy to incorporate Piranha in *"pipelining"*.
```
lazaroclapp, line 115: Just to clarify: 

1) Does this mean there is a separate CLI for Polyglot Piranha, unrelated to `run_piranha_cli`? Where `run_piranha_cli` is just a confusingly named alternative python API, now deprecated in favor of `execute_piranha` and has nothing to do with running Piranha as a CLI tool directly?

Or:

2) Are we actually dropping all support from running Polyglot Piranha as a command line tool directly? (i.e. it's now meant to be used exclusively as a Rust/Python library, and CLI tools will be built using that library).

This docs make me think the first interpretation is correct, but I want to be sure.

ketkarameya (commented):
ow! I forgot to post this :|

lazaroclapp (approved):
LGTM!

PR #380, opened 2023-02-17
Remove the "Cleanup Rule" hack

ketkarameya:
Instead of adding "Cleanup Rule" to group for marking seed/non-seed rules, we now pass a boolean.
Look at our earlier discussion here - https://github.com/uber/piranha/pull/329#pullrequestreview-1267892009 .

While the # files touched is large, its just a consistent pattern that replace `group = ["Cleanup Rule"]` with `is_seed_rule=False`.

----------
* #386
* #384
* __->__ #380
* #379


src/models/rule.rs
```diff
@@ -82,6 +79,13 @@ pub struct Rule {
   #[get = "pub"]
   #[pyo3(get)]
   constraints: HashSet<Constraint>,
+
+  /// Additional constraints for matching the rule
+  #[builder(default = "default_is_seed_rule()")]
+  #[serde(default = "default_is_seed_rule")]
+  #[get = "pub"]
+  #[pyo3(get)]
+  is_seed_rule: bool,
```
ketkarameya, line 88: Add the boolean property to `Rule`

src/models/rule_store.rs
```diff
@@ -53,9 +53,7 @@ impl RuleStore {
     };
 
     for rule in args.rule_graph().rules().clone() {
-      debug!("*** {:?}", rule);
```
ketkarameya, line 56: Dirty logs :| Cleanup

src/models/piranha_arguments.rs
```diff
@@ -268,20 +268,18 @@ impl PiranhaArgumentsBuilder {
...
   let built_in_rules = RuleGraphBuilder::default()
     .edges(piranha_language.edges().clone().unwrap_or_default().edges)
     .rules(piranha_language.rules().clone().unwrap_or_default().rules)
     .build();
 
+  // TODO: Move to `PiranhaArgumentBuilder`'s _validate
+  // Get the user-defined rule graph (if any) via the Python/Rust API
+  let mut user_defined_rules = _arg.rule_graph().clone();
   // In the scenario when rules/edges are passed as toml files
-  let mut user_defined_rules = if !_arg.path_to_configurations().is_empty() {
```
ketkarameya, line 277: Simplified the code. 

`add_to_seed_rules_group` is also removed now.

src/models/rule.rs
```diff
@@ -90,23 +94,10 @@ impl Rule {
...
-  pub(crate) fn is_seed_rule(&self) -> bool {
-    self.groups().contains(&SEED.to_string())
-  }
-
   /// Checks if a rule is `match-only` i.e. it has a query but no replace_node
   pub(crate) fn is_match_only_rule(&self) -> bool {
     *self.query() != default_query() && *self.replace_node() == default_replace_node()
   }
-
-  /// Adds the rule to a new group - "SEED" if applicable.
```
ketkarameya, line 103: No need for these shenanigans :)

ketkarameya (commented):
Self review.

src/models/piranha_arguments.rs
lazaroclapp: Might be worth linking from the comment on the source code to a small issue with this context (even just a copy of this comment's text), just in case that `TODO` lingers there for long enough that you forget this plan above 🙂

test-resources/java/insert_field_and_initializer/configurations/rules.toml
```diff
@@ -42,7 +42,6 @@ query = """
 """
 replace_node = "pkg_dcl"
 replace = "@pkg_dcl \n import java.util.List;"
-groups = ["Cleanup Rule"]
```
lazaroclapp, line 45: Is this a seed rule? Doesn't seem to be setting `is_seed_rule = false`, but it was in the cleanup rules group before...

test-resources/java/non_seed_user_rule/configurations/rules.toml
```diff
@@ -28,7 +28,6 @@ name = "add_import_statement"
 query = "(package_declaration (_)@package_name) @package_declaration"
 replace_node = "package_declaration"
 replace = "@package_declaration\nimport java.util.Collections;\n"
-groups = ["Cleanup Rule"]
```
lazaroclapp, line 31: Same question as above.

test-resources/java/insert_field_and_initializer/configurations/rules.toml
ketkarameya: Great catch. I am disappointed I missed this :| 
Anyways, what is shocking is that even without marking "Cleanup Rules", the tests passed. 
I have updated the tests, and added `is_seed_rule= False`

https://github.com/uber/piranha/pull/379/commits/3c93d040785a68836b84f167de336c75056d9578 - Fixed rust tests 
https://github.com/uber/piranha/pull/380/commits/415d23888d6a78bf5b6c774da1f882e10e530ac3 - Fixed PyTest 

At a high level, I added a file that shud not be transformed by the given rule graph.

src/models/piranha_arguments.rs
ketkarameya: Added issue, updated the todo in the comment 
https://github.com/uber/piranha/commit/d78602292539bbf43a68860008cb8c91c0c5ac6f 
https://github.com/uber/piranha/issues/387

test-resources/java/insert_field_and_initializer/configurations/rules.toml
lazaroclapp: > Anyways, what is shocking is that even without marking "Cleanup Rules", the tests passed.

Do we have an explanation as to why? Might be worth making sure we are not missing a test case there.

tests/tests.py
```diff
@@ -141,17 +141,27 @@ def test_insert_field_add_import():
...
+            ]
+            # Check if the file was rewritten
+            if updated_content:
+                if expected_content != updated_content[0]:
+                    return False
+            else:
+                # The scenario where the file is not expected to be rewritten
+                original_content= Path(join(input_dir, file_name)).read_text()
+                if expected_content != "".join(original_content.split()):
+                    return False
```
lazaroclapp, line 164: So, technically this check is just ensuring that we propagate files from `input/` to `expected/` when Piranha is not supposed to rewrite them, right? It allows us to have files in our tests that Piranha is not supposed to modify.

The actual new test cases are the added files above, which we just want to make sure don't get extraneous imports added to (because the rule will never run on them, since it's not a seed rule, nor triggered from another rule running on those same files).

Am I getting all of this right?

lazaroclapp (approved):
Looks good to me _assuming_ I got the explanation below right. If that's not the case, please let me know 🙂

tests/tests.py
lazaroclapp: Ok, this all makes sense and is what I thought (But nicely explained! Thanks!).

No further review needed from me on this one unless there are new changes, btw. LGTM and I already approved.

PR #384, opened 2023-02-21
Add python `__str__` and  `__repr__` methods for rule, rulegraph and edge

ketkarameya:
Adds `__repr__` and `__str__` pymethods `RuleGraph`, `Rule` and `OutgoingEdge`.

Updated the macro slightly to reason about the scenario when we just want to add just the methods in an existing `impl Self`.

----
* #386
* __->__ #384
* #380
* #379


src/utilities/mod.rs
```diff
@@ -124,6 +124,14 @@ macro_rules! gen_py_str_methods {
       }
     }
   };
+  () => {
```
ketkarameya, line 127: Oh boy! I spent some significant time trying to reuse the match arms of the macro_rule, didnt make progress. Will clean it up later when i figure it out. :|

ketkarameya (commented):
Self review and annotation.

src/utilities/mod.rs
lazaroclapp: Not sure I fully understand this comment or what the code is doing here. Isn't this duplicated with the code right above for `($struct_name:ident) =>`? Or is it still two cases?

Can you explain this change a bit more? (Potentially in the PR Description)

ketkarameya: I added this description and todo into the code itself, so that its easy for me or someone to understand what the desired cleanup is - https://github.com/uber/piranha/pull/384/commits/0ef979884eceecf1ee4cf35e0855a8f391a4968a

lazaroclapp (approved):
LGTM! Way clearer now what's going on there!

PR #386, opened 2023-02-21
Bug/Feature: Handle syntactically incorrect input trees

ketkarameya:
We**Background**: Currently Piranha accepts syntactically correct code, and panics when an applied a rule creates syntactically incorrect code. 

**Feature**: We want to allow Piranha to operate over partially correct ASTs when the user passes the option `--allow-dirty-ast` . We want this feature because it will (i) make code snippet mode better (not all code snippets are syntactically correct always , they might miss a brace here and there) (ii) allow us to operate over files with advanced language features that our grammar may not be supporting at the moment.

**Implementation**: 
* Add `--alow-dirty-ast` flag in `PiranhaArguments`
  *  This flag ensures that a source code unit cannot be constructed if the AST is errorneous
* Slightly tweak the core algorithm.  Instead of panicking when an erroneous AST is produced, we now panic only of the number of errors in the AST have increased.  This is a more general, than what we had earlier. 

**Test Plan**: Added three test cases for: 
  * Transform when input code is dirty, allow dirty is set and rule does not create more errors 
  * Panic when input code is dirty, allow dirty is set and rule creates more errors 
  * Panic when input code is dirty and allow dirty is **not** set 

-----------
* __->__ #386

### DIFF
--- a/src/models/source_code_unit.rs
+++ b/src/models/source_code_unit.rs
@@ -320,13 +320,6 @@ impl SourceCodeUnit {
     self.perform_delete_consecutive_new_lines();
   }
 
-  pub(crate) fn apply_edit(&mut self, edit: &Edit, parser: &mut Parser) -> InputEdit {
-    // Get the tree_sitter's input edit representation
-    let mut applied_edit = self._apply_edit(edit, parser);
-    self._delete_associated_comment(edit, &mut applied_edit, parser);
-    applied_edit
-  }
-
   /// Applies an edit to the source code unit
   /// # Arguments
   /// * `replace_range` - the range of code to be replaced
@@ -337,14 +330,14 @@ impl SourceCodeUnit {
   /// The `edit:InputEdit` performed.
   ///
   /// Note - Causes side effect. - Updates `self.ast` and `self.code`
-  pub(crate) fn _apply_edit(&mut self, edit: &Edit, parser: &mut Parser) -> InputEdit {
-    let mut edit_to_apply = edit.clone();
+  pub(crate) fn apply_edit(&mut self, edit: &Edit, parser: &mut Parser) -> InputEdit {
+    let mut edit: Edit = edit.clone();
     // Check if the edit is a `Delete` operation then delete trailing comma
-    if edit.replacement_string().trim().is_empty() {
-      edit_to_apply = self.delete_trailing_comma(edit);
+    if edit.is_delete() {
+      edit = self.delete_trailing_comma(&edit);
     }
     // Get the tree_sitter's input edit representation
-    let (new_source_code, ts_edit) = get_tree_sitter_edit(self.code.clone(), &edit_to_apply);
+    let (new_source_code, ts_edit) = get_tree_sitter_edit(self.code.clone(), &edit);
     // Apply edit to the tree
     let number_of_errors = self._number_of_errors();
     self.ast.edit(&ts_edit);
@@ -353,6 +346,12 @@ impl SourceCodeUnit {
     // Panic if the number of errors increased after the edit
     if self._number_of_errors() > number_of_errors {
       self._panic_for_syntax_error();
+    }
+    // Check if the edit is a `Delete` operation then delete associated comment
+    if edit.is_delete() && *self.piranha_arguments().cleanup_comments() {
+      if let Some(deleted_comment) = self._delete_associated_comment(&edit, parser) {
+        return deleted_comment;
+      }
     }
     ts_edit
   }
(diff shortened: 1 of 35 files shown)


src/models/source_code_unit.rs
```diff
@@ -339,9 +339,12 @@ impl SourceCodeUnit {
     // Get the tree_sitter's input edit representation
     let (new_source_code, ts_edit) = get_tree_sitter_edit(self.code.clone(), &edit_to_apply);
     // Apply edit to the tree
+    let number_of_errors = self._number_of_errors();
```
ketkarameya, line 342: It could maybe make sense to have this thing configurable via the command line? 
`--allow-dirty-tree`

ketkarameya (commented):
Self review.

src/models/unit_tests/rule_test.rs
```diff
@@ -154,7 +154,7 @@ fn test_get_edit_negative_recursive() {
       ]
   };
   let source_code = "class Test {
-          pub void foobar(){
```
ketkarameya, line 157: It seems that I had some syntactically incorrect test cases. I had to fix them because `allow-dirty-ast` is `false` by default

src/models/source_code_unit.rs
```diff
@@ -339,19 +346,33 @@ impl SourceCodeUnit {
...
-      panic!("{}", msg);
+
+    // Panic if the number of errors increased after the edit
+    if self._number_of_errors() > number_of_errors {
+      self._panic_for_syntax_error();
     }
     ts_edit
   }
 
+  fn _panic_for_syntax_error(&self) {
```
ketkarameya, line 360: extracted out the panic for syntax error!

ketkarameya (commented):
Self review and annotations.

src/tests/test_piranha_java.rs
```diff
@@ -254,3 +254,102 @@ fn test_consecutive_scope_level_rules() {
...
+                .rules(vec![rule])
+                .build(),
+  };
+
+  execute_piranha_and_check_result(&piranha_arguments, &_path.join("expected"), 1, true);
+  // Delete temp_dir
+  temp_dir.close().unwrap();
+}
+
+/// This test is to check if the Piranha is able to handle a syntactically incorrect tree.
```
lazaroclapp, line 323:
```suggestion
/// This test is to check if Piranha is able to handle a syntactically incorrect tree.
```

src/tests/test_piranha_java.rs
```diff
@@ -254,3 +254,102 @@ fn test_consecutive_scope_level_rules() {
...
+                .build(),
+    allow_dirty_ast = true,
+  };
+
+  execute_piranha_and_check_result(&piranha_arguments, &_path.join("expected"), 1, true);
+  // Delete temp_dir
+  temp_dir.close().unwrap();
+}
+
+/// This test is to check if the Piranha panics when it encounters a syntactically incorrect tree and
```
lazaroclapp, line 290:
```suggestion
/// This test is to check if Piranha panics when it encounters a syntactically incorrect tree and
```

src/tests/test_piranha_java.rs
```diff
@@ -254,3 +254,102 @@ fn test_consecutive_scope_level_rules() {
 
   execute_piranha_and_check_result(&args, _path.join("expected").as_path(), 1, true)
 }
+
+/// This test is to check if the Piranha is able to handle a syntactically incorrect tree.
```
lazaroclapp, line 258:
```suggestion
/// This test is to check if Piranha is able to handle a syntactically incorrect tree.
```

src/models/source_code_unit.rs
lazaroclapp: Wait. Not sure I understand this. Are you saying this is not configurable through the command line? I thought the whole point of PiranhaArguments was a similar interface between CLI and programatic invocation... what am I missing? (Or you mean having a different argument that allows the number of errors to increase?)

src/models/source_code_unit.rs
```diff
@@ -339,19 +346,33 @@ impl SourceCodeUnit {
...
+    let msg = format!(
+      "Produced syntactically incorrect source code {}",
+      self.code()
+    );
+    error!("{}", msg);
```
lazaroclapp, line 365: Why do we need to call both `error` and `panic`?

src/models/source_code_unit.rs
```diff
@@ -339,19 +346,33 @@ impl SourceCodeUnit {
...
+    panic!("{}", msg);
+  }
+
+  /// Returns the number of errors in the AST
+  fn _number_of_errors(&self) -> usize {
+    traverse(self.root_node().walk(), Order::Post)
+      .filter(|node| node.has_error())
+      .count()
+  }
```
lazaroclapp, line 374: One thing that can happen here, which I am not sure how to account for, is a Piranha change making a larger subset of the AST show as an error than before, while retaining the same or smaller error count (e.g. Piranha makes a change that breaks parsing for an entire class which already had at least one syntax error somewhere. Now the entire class shows as an error, and the error count is still 1, but a lot more of the code is unparsable/unanalyzable)

lazaroclapp (commented):
Some comments below, overall this looks good

PR #388, opened 2023-02-22
Added rules to clean if-false from swift

Satyam1749:
Rules to be created 

```
Before
if false {
   f1()
}
f2()
After
f2()
```

3 rules are created and we are depending on the order of rules to clean all corner cases of if always false.
 1. else if is always false and followed by else if
 2. else if is always false and followed by else or nothing
 3. if is always false.

These rules also handles cases with if let
i.e. below mentioned case will also be handled
```
Before
if let value = value, false {
   f1()
}
f2()

After
f2()
```

### DIFF
--- a/test-resources/swift/cleanup_rules/input/SampleClass.swift
+++ b/test-resources/swift/cleanup_rules/input/SampleClass.swift
@@ -93,4 +93,94 @@ class SampleClass {
             f5()
         }
     }
+    
+    func checkIfFalse() {
+        //test comments to be cleaned
+        if !TestEnum.stale_flag_one.isEnabled && abc {
+            f1()
+        }
+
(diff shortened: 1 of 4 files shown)


test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -39,4 +39,99 @@ class SampleClass {
...
+
+        if v1 {
+            f1()
+        } else if let v1 = v1, !TestEnum.stale_flag_one.isEnabled && abc {
+            f2()
+        } else {
+            f3()
+        }
+
+        if v1 {
```
ketkarameya, line 121: Can we have some comments, and check if they are cleaned up too ? You will have to pass piranha arguments `claenup_comments = True`

ketkarameya (approved):
LGTM.

PR #389, opened 2023-02-22
Added rule to cleanup guard statements which are always true

Satyam1749:
Rules created

```
Before 
guard true else {
   return
}
f1()
After
f1()
```

### DIFF
--- a/test-resources/swift/cleanup_rules/input/SampleClass.swift
+++ b/test-resources/swift/cleanup_rules/input/SampleClass.swift
@@ -183,4 +183,18 @@ class SampleClass {
             f2()
         }
     }
+    
+    func checkGaurdTrue() {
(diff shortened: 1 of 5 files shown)


test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -39,4 +39,11 @@ class SampleClass {
         isEnabled = f2() || placeholder_false 
         isEnabled = v2 || placeholder_false 
     }
+
+    func checkGaurdTrue() {
+        guard TestEnum.stale_flag_one.isEnabled || f1() else {
```
ketkarameya, line 44: Same comment as the other PR. Do we want to check for and too ?

ketkarameya (changes requested):

ketkarameya (approved):
LGTM

Issue #376, opened 2023-02-16
Go comments not being cleaned up correctly


conormurray95:
**Issue**
When running the tool with cleanup_comments = true against a go codebase I'm seeing some inconsistent behaviour around comment cleanup.

**Steps to reproduce**
I've added test cases demonstrating this issue in this pr https://github.com/uber/piranha/pull/375 

**Observations**:

- Multiline comments are cleaned up 
- Single line comments are not cleaned up
- Single line comments are cleaned up provided the next node above is also a comment

The test cases provided in the pr are for cleaning up if blocks however I've noticed the same behaviour when cleaning up constants in my own codebase e.g.

**Before**
```
const (
    // comment
    myFlag = "flag"

   // stale comment
   myStaleFlag = "stale flag"
)
```

**After**
```
const (
    // comment
    myFlag = "flag"

   // stale comment
)
```

**Expected**
```
const (
    // comment
    myFlag = "flag"
)
```

The constants example above exhibits the same behaviour around only cleaning up comments if they're (a) multiline or (b) have another comment node directly above them.

Any context on if this is a known issue in other languages or seems go specific etc would be great. It's the one piece left stopping autogenerated pr's we have being able to be directly merged.

ketkarameya:
Hey! Thanks a lot for reporting this. 
the go cleanup mechanism we have is pretty much experimental :| We have not battle tested it yet (by running it on our internal codebase) But thanks for doing this :) I will look into it. Like our java/kotlin cleanup is kinda robust. 
cc @dvmarcilio

ketkarameya:
@conormurray95 can you check this branch `go_comment_bug` https://github.com/uber/piranha/pull/392 on your go codebase, and let me know if it works as expected?

conormurray95:
Hey @ketkarameya apologies I just got a chance to look back at this, that pr looks good and solves the cases I'd spotted. Thanks!

PR #392, opened 2023-02-23
Fix go comments bug

ketkarameya:
**Bug Diagnosis** (#376):  This problem is only in`Go` (because of its grammar, in my observation). The problem is caused when the comment is associated to the last statement of a method body in `Go`. Diagnosis: 
Our comment fetching [algorithm](https://github.com/uber/piranha/blob/8edb8dbdc6da2c704efddc6894c7dcbf110414ad/src/models/piranha_arguments.rs#L357) looks for comments on lines on the same line as deleted element and some lines above it (determined by the `cleanup_comment_buffer` flag).  There is a caveat in this algorithm - (i) Get all the nodes that either start and end at [row] (ii) If **all** nodes are comments. 
Now lets take the `Go` case where the comment is associated to the last statement of a method body in `Go`. 
When we delete the last statement of the method body, now the last "statement" of the method body is a comment. 
For some reason, tree sitter reports that this last row is where the comment and the block end.  :| 

**Fix**: Add a field `ignore_node_for_comment` property to `PiranhaLanguage`, now our algorithm will ignore the node kinds when reasoning about **all** nodes that start/end at last row. 

**Test Plan** Added the test cases as reported in the issue. I added similar test cases for Java and Kotlin too, but I did not find this inconsistency for them (so did not have set the `ignore_node_for_comment` property for Java and Kotlin)

### DIFF
--- a/src/models/language.rs
+++ b/src/models/language.rs
@@ -15,7 +15,8 @@ use std::str::FromStr;
 
 use getset::Getters;
 use serde_derive::Deserialize;
-use tree_sitter::{Node, Parser, Query};
+use tree_sitter::{Node, Parser, Query, Range};
+use tree_sitter_traversal::{traverse, Order};
 
 use crate::utilities::parse_toml;
 
@@ -245,5 +246,72 @@ impl SourceCodeUnit {
         .language()
         .ignore_nodes_for_comments()
         .contains(&node.kind().to_string())
+  }
+
+  /// This function reports the range of the comment associated to the deleted element.
+  ///
+  /// # Arguments:
+  /// * row : The row number where the deleted element started
+  /// * buffer: Number of lines that we want to look up to find associated comment
+  ///
+  /// # Algorithm :
+  /// Get all the nodes that either start and end at [row]
+  /// If **all** nodes are comments
+  /// * return the range of the comment
+  /// If the [row] has no node that either starts/ends there:
+  /// * recursively call this method for [row] -1 (until buffer is positive)
+  /// This function reports the range of the comment associated to the deleted element.
+  ///
+  /// # Arguments:
+  /// * row : The row number where the deleted element started
+  /// * buffer: Number of lines that we want to look up to find associated comment
+  ///
+  /// # Algorithm :
+  /// Get all the nodes that either start and end at [row]
+  /// If **all** nodes are comments
+  /// * return the range of the comment
+  /// If the [row] has no node that either starts/ends there:
+  /// * recursively call this method for [row] -1 (until buffer is positive)
+  pub(crate) fn _get_comment_at_line(
+    &mut self, row: usize, buffer: usize, start_byte: usize,
+  ) -> Option<Range> {
+    // Get all nodes that start or end on `updated_row`.
+    let mut relevant_nodes_found = false;
+    let mut relevant_nodes_are_comments = true;
+    let mut comment_range = None;
+    // Since the previous edit was a delete, the start and end of the replacement range is [start_byte].
+    let node = self
+      .root_node()
+      .descendant_for_byte_range(start_byte, start_byte)
+      .unwrap_or_else(|| self.root_node());
+
+    // Search for all nodes starting/ending in the current row.
+    // if it is not amongst the `language.ignore_node_for_comment`, then check if it is a comment
+    // If the node is a comment, then return its range
+    for node in traverse(node.walk(), Order::Post) {
+      if self.should_ignore_node_for_comment_search(node) {
+        continue;
+      }
+
+      if node.start_position().row == row || node.end_position().row == row {
+        relevant_nodes_found = true;
+        let is_comment: bool = self.is_comment(node.kind().to_string());
+        relevant_nodes_are_comments = relevant_nodes_are_comments && is_comment;
+        if is_comment {
+          comment_range = Some(node.range());
+        }
+      }
+    }
+
+    if relevant_nodes_found {
+      if relevant_nodes_are_comments {
+        return comment_range;
+      }
+    } else if buffer > 0 {
+      // We pass [start_byte] itself, because we know that parent of the current row is the parent of the above row too.
+      // If that's not the case, its okay, because we will not find any comments in these scenarios.
+      return self._get_comment_at_line(row - 1, buffer - 1, start_byte);
+    }
+    None
   }
 }
--- a/src/models/language.rs
+++ b/src/models/language.rs
@@ -15,7 +15,7 @@ use std::str::FromStr;
 
 use getset::Getters;
 use serde_derive::Deserialize;
-use tree_sitter::{Parser, Query};
+use tree_sitter::{Node, Parser, Query};
 
 use crate::utilities::parse_toml;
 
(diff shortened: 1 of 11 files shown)


src/models/language.rs
```diff
@@ -28,20 +28,30 @@ use super::{
...
   edges: Option<Edges>,
+  /// Scope configurations for the language
   #[get = "pub(crate)"]
   scopes: Vec<ScopeGenerator>,
+  /// The node kinds to be considered when reasoning about comments
   #[get = "pub"]
   comment_nodes: Vec<String>,
+  /// The node kinds to be ignored when reasoning about comments
+  #[get = "pub"]
+  ignore_nodes_for_comments: Vec<String>,
```
ketkarameya, line 54: The property i mentioned in the PR description.

src/models/piranha_arguments.rs
```diff
@@ -386,6 +386,13 @@ impl SourceCodeUnit {
       .unwrap_or_else(|| self.root_node());
 
     for node in traverse(node.walk(), Order::Post) {
+      if self
```
ketkarameya, line 389: The bug fix.

ketkarameya (commented):
Self review and annotations

test-resources/go/feature_flag/builtin_rules/statement_cleanup/input/sample.go
```diff
@@ -106,3 +106,57 @@ func after_return4() string {
...
 }
+
+
+func simplify_if_statement_false_comment_demo_single_comment() {
+    if exp.BoolValue("false") {
+        fmt.Println("to be removed")
+    } else {
+        fmt.Println("remain")
+    }
+    // this comment doesnt get removed but it should
```
lazaroclapp, line 117: No longer true, no? Shouldn't the comment be updated to saying it will get removed (same with all the cases below)

src/models/language.rs
```diff
@@ -28,20 +28,30 @@ use super::{
...
+  /// Built-in edges for the language
   #[get = "pub(crate)"]
   edges: Option<Edges>,
+  /// Scope configurations for the language
   #[get = "pub(crate)"]
   scopes: Vec<ScopeGenerator>,
+  /// The node kinds to be considered when reasoning about comments
   #[get = "pub"]
   comment_nodes: Vec<String>,
+  /// The node kinds to be ignored when reasoning about comments
```
lazaroclapp, line 52:
```suggestion
  /// The node kinds to be ignored when searching for comments
```

Just to be clear, the algorithm will find the nearest node to the deleted position that is not one of these. Then, if that node is a comment, it will delete it. Correct? Am I understanding the logic?

src/models/language.rs
```diff
@@ -66,6 +76,13 @@ impl PiranhaLanguage {
     self.comment_nodes().contains(&kind)
   }
 
+  pub fn should_ignore_node_for_comment(&self, node: &Node) -> bool {
```
lazaroclapp, line 79: Documentation on this one, explaining what it does and how it's used.

Like, is the following correct?

> [Piranha] will find the nearest node to the deleted position that is not one of these. Then, if that node is a comment, it will delete it.

(Even if so, please rewrite the doc comment a bit with some more info about how this function is used to implement the algorithm above)

lazaroclapp (commented):
Change looks great (assuming I understood it correctly). Some minor requests regarding documentation and what seems to be a few stale comments 🙂

src/models/language.rs
ketkarameya: Yes. You have got it correct. I have made it more explicit now.

ketkarameya (commented):
Addressed comments. 
https://github.com/uber/piranha/pull/392/commits/c1086c94e1aaa327fa7c27d2d6ccf6bc62c16491 - Adds more comments, fixes comments, and slight cleanup 

https://github.com/uber/piranha/pull/392/commits/76bce7d32966f9051aac8bcd042c36bdd0fc3e2d - move `_get_comment_at_line` from `piranha_argument` to `language`

test-resources/go/feature_flag/builtin_rules/statement_cleanup/expected/sample.go
```diff
@@ -72,7 +72,6 @@ func simplify_if_statement_false_comment_demo_single_comment() {
 
 func simplify_if_statement_false_comment_demo_double_comment() {
     fmt.Println("remain")
-    // this comment doesnt get removed
```
ketkarameya: The bug was not fully solved :| . Now its solved.

src/models/piranha_arguments.rs
```diff
@@ -331,20 +331,13 @@ mod piranha_arguments_test;
...
-      let deleted_at = edit.p_match().range().start_point.row;
-      if let Some(comment_range) = self._get_comment_at_line(
-        deleted_at,
-        *self.piranha_arguments().cleanup_comments_buffer(),
-        edit.p_match().range().start_byte,
-      ) {
-        debug!("Deleting an associated comment");
-        *applied_edit = self._apply_edit(&Edit::delete_range(self.code(), comment_range), parser);
-      }
+    &mut self, edit: &Edit, parser: &mut tree_sitter::Parser,
```
ketkarameya, line 334: from `apply_edit` we call `_delete_associated_comment` and from `_delete_associated_comment` we recursively call `apply_edit`. This ensures that can delete a series of comments.

Let's say `apply_edit` deletes `int foo` in the below example.
```
// some 
// comment 
int foo;
```

becomes 

```
// some 
// comment 
```

now `_delete_associated_comment` is triggered, and now we produce
```
// some 
```

Since `_delete_associated_comment` is effectively called recursively 

```
<all comments are deleted>
```

src/models/source_code_unit.rs
```diff
@@ -313,13 +313,6 @@ impl SourceCodeUnit {
     self.perform_delete_consecutive_new_lines();
   }
 
-  pub(crate) fn apply_edit(&mut self, edit: &Edit, parser: &mut Parser) -> InputEdit {
```
ketkarameya, line 316: inlined `apply_edit`, this simplifies the code and also kinda solves this bug further. 

Earlier we had a function called `apply_edit` which called `_apply_edit`  **and then** called `_delete_associated_comment` . Now `_delete_associated_comment` called `_apply_edit` (not **`apply_edit`** ), therefore `_delete_associated_comment` was not getting invoked recursively, and therefore we were not deleting a series of comments. 

I just inlined `apply_edit`  into `_apply_edit` and renamed `_apply_edit` to `apply_edit`. Now when `apply_edit` is called from `_delete_associated_comment`, it ensures that `_delete_associated_comment` is called recursively. (I adjusted all the signatures and simplified the code)

ketkarameya (commented):
Aaah! I missed a corner case in this bug fix. :|

src/models/piranha_arguments.rs
lazaroclapp: Can we have this documented in a comment in the code in addition to in the PR review comments? Possibly in the documentation for this function.

lazaroclapp (approved):
LGTM!

PR #393, opened 2023-02-23
Added rule for always false guard block cleanup

Satyam1749:
Rule added:

```
Before 
guard false else {
   return
}
After
return
```

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -407,6 +407,31 @@ replace_node = "guard_block"
 replace = ""
 is_seed_rule = false
 
+#
+# Before
+#   guard false else {
+#       return f1()
+#   }
+#   f2()
+# After 
+#   return f1()
+#   f2()
+#
+[[rules]]
+name = "guard_always_false"
+query = """(
+(guard_statement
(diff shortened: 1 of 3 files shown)


test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -39,4 +39,10 @@ class SampleClass {
         isEnabled = f2() || placeholder_false 
         isEnabled = v2 || placeholder_false 
     }
+
+    func checkGuardFalse() {
+        guard placeholder_false || false else {
```
ketkarameya, line 44: do we want to check for conjunctions too ? And also for `true` ?

ketkarameya (changes requested):

src/cleanup_rules/swift/rules.toml
```diff
@@ -231,6 +231,31 @@ replace_node = "not_expression"
 replace = "true"
 is_seed_rule = false
 
+#
+# Before
+#   guard true else {
```
ketkarameya, line 236: update comment.

ketkarameya (approved):
(Requested Change in line)

PR #394, opened 2023-02-23
Added rules for ternary operator cleanup

Satyam1749:
Rules added

```
Before
true ? x : y
After
x
```

```
Before
false ? x : y
After
y
```


ketkarameya (approved):
LGTM.

PR #397, opened 2023-02-27
Rules for cleaning up case from switch statements of variables in an enum

Satyam1749:
This PR contains rule to cleanup any enum entry from all the variables in that enum. In swift, we use switch in enums to populate computed properties. These rules aims to cleanup any enum entry from all such variables.

example:
Before
```
enum TestEnum {
   case one
   case two
   case three

   var v1: String {
      switch self {
      case .one:
         return "one"
      case .two:
         return "two"
      case .three:
         return "three"
      }
   }
}
```

After
```
enum TestEnum {
   case two
   case three

   var v1: String {
      switch self {
      case .two:
         return "two"
      case .three:
         return "three"
      }
   }
}
```

`case one ` should be cleaned up by a custom rule which needs to belong to group `enum_entry_cleanup`, This will cleanup  entry of `case one` from `var v1`.

Limitation:
Because of limitations of grammar. We are not able to completely clean this specific case (where enum entry to be deleted is a part of case with multiple entries and is last in those entries), For example `.one` in below case:
```
switch self {
   case .two, .one:
       return "ret"
}
```

we modify this to 
```
switch self {
   case .two, .two:
       return "ret"
}
```

This is not pretty but it works and is compilable


ketkarameya:
I guess after https://github.com/uber/piranha/pull/410 is merged, the limitation would no longer exist. 
But it also implies that you might have to update the PR to not use the hack.

src/cleanup_rules/swift/rules.toml
```diff
@@ -431,6 +431,137 @@ query = """(
...
+#        }
+#    }
+# }   
+#
+# After
+# enum TestEnum {
+#    var v1: String {
+#        switch self {
+#        case .two, .two:
+#            return "one"
```
ketkarameya, line 496: We no longer need this right?

ketkarameya (commented):
Did my fix in 410 not work?

PR #406, opened 2023-03-21
Fix (identifier && false) and (identifier || true) for Java

Adarsh-NP:


Adarsh-NP:
Ah, my bad..looks like my comments are out of place. I'll update the PRs.

Issue #286, opened 2022-12-06
Fix disjunction_rule in Java and Kotlin


dvmarcilio:
## Something **or** `true`

The query `simplify_something_or_true` for Java and Kotlin is currently wrong.

https://github.com/uber/piranha/blob/de2d604e54f7a361b584586b49337aa511409bbb/polyglot/piranha/src/cleanup_rules/java/rules.toml#L217-L227

`@lhs` **or** `true` should always be true.

Simplifying just for `@lhs` is behavior-changing. 
In the code below, consider if `something()` (`@lhs`) ends up returning `false`.

```java
if (something() || true) {
    System.out.println("always taken");
}
```

Simplifying to `something()` would lead to the branch **not** be taken.

## Something **and** `false`

If we want to be safe in such side-effects scenarios, we also need to consider `simplify_something_and_false`.

https://github.com/uber/piranha/blob/de2d604e54f7a361b584586b49337aa511409bbb/polyglot/piranha/src/cleanup_rules/java/rules.toml#L196-L208

Simplifying only to `false` will skip `something()` from executing.

```java
if (something() && false) {}
```

## Discussion
 
My main idea is to simplify as before (without considering side-effects) **only** when we are dealing with identifiers-like nodes (e.g., identifier, field access).

This leads to larger cleanups as more if conditions would eventually reach `if_cleanup`. A plus is also that our rewrites are safe since we're not discarding calls.

I can think of two downsides:
- Piranha Polyglot users may not care about side effects. That could be the case if the goal is to maximize cleanups.
- We need to write additional `if_cleanup` rules. In the code below, we also need to clean up the else branch.

```java
if (something() || true) {
    // keep
} else if (somethingElse) {
    // whole `else if` should be removed
}
else {
    // whole `else` should be removed
}
```
---

Another idea, in addition to the proposed solution above, would be to have kind of a global argument to ignore side effects. If this argument is `true`, we could use our straightforward rules, just as before. If `false`, we use the solution above.
The downsides are that we need to maintain: 
-  two different sets of rules;
- some logic to select between the rules depending on the `ignore-side-effects` argument value.

PR #408, opened 2023-03-21
Fix (identifier && false) and (identifier || true) for Kotlin

Adarsh-NP:
This PR solves issues around (identifier && false) and (identifier || true) for Kotlin. No cleanups if there is function call in case of identifier for the two cases. Solves Kotlin part of #286

### DIFF
--- a/src/cleanup_rules/kt/rules.toml
+++ b/src/cleanup_rules/kt/rules.toml
@@ -269,7 +269,7 @@ replace_node = "conjunction_expression"
 is_seed_rule = false
 
 # Before :
-#  abc() && false
+#  abc && false
 # After :
 #  false
 #
@@ -278,7 +278,11 @@ groups = ["boolean_expression_simplify"]
 name = "simplify_something_and_false"
 query = """
 (
-(conjunction_expression (_) @lhs 
+(conjunction_expression [(simple_identifier)
+                          (parenthesized_expression (simple_identifier))
+                          (boolean_literal)
+                          (parenthesized_expression (boolean_literal))
+                        ] @lhs 
                         [((boolean_literal) @rhs) (parenthesized_expression (boolean_literal) @rhs)] ) @conjunction_expression
 (#eq? @rhs "false")  
 )
@@ -288,7 +292,7 @@ replace_node = "conjunction_expression"
 is_seed_rule = false
 
 # Before :
-#  abc() || true
+#  abc || true
 # After :
 #  true
 #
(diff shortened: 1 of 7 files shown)


src/cleanup_rules/kt/rules.toml
```diff
@@ -297,7 +301,11 @@ groups = ["boolean_expression_simplify"]
 name = "simplify_something_or_true"
 query = """
 (
-(disjunction_expression (_) @lhs 
+(disjunction_expression [(simple_identifier)
```
ketkarameya, line 304: look at my comment from the equivalent java PR

test-resources/kt/feature_flag_system_2/control/input/XPMethodChainCases.kt
```diff
@@ -47,6 +47,14 @@ internal class XPMethodChainCases {
             if (sp.isOtherFlag().cachedValue || sp.isStaleFeature().cachedValue) {
                 println("!!!")
             }
+            // simple_identifier && false, should be deleted
+            if (a && sp.isStaleFeature().cachedValue){
+                println("!!!")
```
ketkarameya, line 52: can we have different strings in the `println` ? So that its easier to figure out the clenaup?

test-resources/kt/feature_flag_system_1/control/expected/XPFlagCleanerCases.kt
```diff
@@ -57,7 +57,7 @@ internal class XPFlagCleanerPositiveCases {
         tBool = false
         tBool = false
 
-        tBool = false
+        tBool = true
```
ketkarameya, line 60: nice. This was a bug i guess. 
`tBool = experimentation.isToggleEnabled(TestExperimentName.STALE_FLAG) || true` was evaluating to false earlier :O

ketkarameya (changes requested):

test-resources/kt/feature_flag_system_2/treated/input/XPMethodChainCases.kt
```diff
@@ -47,6 +47,14 @@ internal class XPMethodChainCases {
             if (sp.isOtherFlag().cachedValue || sp.isStaleFeature().cachedValue) {
                 println("!!!")
             }
+            // simple_identifier && false, should be deleted
+            if (a && !sp.isStaleFeature().cachedValue){
```
ketkarameya, line 51: since you are at it, please update these `println` too

test-resources/kt/feature_flag_system_2/treated/expected/XPMethodChainCases.kt
```diff
@@ -39,9 +39,12 @@ internal class XPMethodChainCases {
             if (sp.isOtherFlag().cachedValue) {
                 println("!!!")
             }
-            if (sp.isOtherFlag().cachedValue) {
+            if (sp.isOtherFlag().cachedValue || true) {
```
ketkarameya, line 42: Let's update these `println`s too

ketkarameya (approved):
Approve but please address my comments

PR #409, opened 2023-03-21
Fix (something || true) and (something && false) for go

### LINKED ISSUE #286: Fix disjunction_rule in Java and Kotlin

Adarsh-NP:
This PR fixes Fix (something || true) and (something && false) for go. In case of function calls, there will be no cleanups for the two cases. Fixes #286 for Go.

### DIFF
--- a/src/cleanup_rules/go/rules.toml
+++ b/src/cleanup_rules/go/rules.toml
@@ -120,7 +120,7 @@ groups = ["boolean_expression_simplify"]
 is_seed_rule = false
 
 # Before :
-#  abc() && false
+#  abc && false
 # After :
 #  false
 #
@@ -131,9 +131,13 @@ query = """
     (binary_expression
         left : ([
                 (identifier)
+                (parenthesized_expression (identifier))
                 (true)
+                (parenthesized_expression (true))
                 (false)
+                (parenthesized_expression (false))
                 (selector_expression)
+                (parenthesized_expression (selector_expression))
             ]) @lhs
         operator : "&&"
         right: [(false) (parenthesized_expression (false))]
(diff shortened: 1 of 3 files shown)


src/cleanup_rules/go/rules.toml
```diff
@@ -131,9 +131,13 @@ query = """
     (binary_expression
```
ketkarameya, line 131: Please update the comment for this. 

`abc && false`

ketkarameya (approved):
LGTM. Please update the comment as mentioned inline.

PR #410, opened 2023-03-21
A cleaner way to handle commas and comments 

ketkarameya:
TLDR; Earlier we were handling commas and handling comments. 
Now we handle commas and comments. 
Goal : support trailing and leading comma/comment cleanup 

Earlier we had a way to detect trailing commas before performing a deletion and expanding the deletion to include it. We also had a mechanism that detected the trailing and leading comments for a deleted code snippet.  The problem was that we were doing something before deletion and some after deletion. It is very difficult to have any kind of information flow between comma and comment cleanup. This was required when we want to cleanup leading commas / comments
Now, we cleanup all of this mess. 

I have proposed a simple approach that simply visits siblings of each matched rule and collects its associated commas and comments.
For deletions, we then extend our match to the start (floor) and end (ceiling) of the associated matches => extending the deletion to include the comma / comment

### DIFF
--- a/src/models/matches.rs
+++ b/src/models/matches.rs
@@ -49,7 +49,7 @@ pub(crate) struct Match {
   #[get_mut]
   #[serde(skip)]
   associated_comma: Option<Range>,
-  // Captures the range of the associated comments
+  // Captures the range(s) of the associated comments
   #[get]
   #[get_mut]
   #[serde(skip)]
--- a/src/models/matches.rs
+++ b/src/models/matches.rs
@@ -11,9 +11,10 @@ Copyright (c) 2022 Uber Technologies, Inc.
  limitations under the License.
 */
 
-use std::collections::HashMap;
+use std::{cmp::Ordering, collections::HashMap};
 
-use getset::Getters;
+use getset::{Getters, MutGetters};
+use itertools::Itertools;
 use log::{debug, trace};
 use pyo3::prelude::{pyclass, pymethods};
 use serde_derive::{Deserialize, Serialize};
@@ -26,7 +27,7 @@ use crate::utilities::{
 
 use super::{rule::InstantiatedRule, rule_store::RuleStore, source_code_unit::SourceCodeUnit};
 
-#[derive(Serialize, Debug, Clone, Getters, Deserialize)]
+#[derive(Serialize, Debug, Clone, Getters, MutGetters, Deserialize)]
 #[pyclass]
 pub(crate) struct Match {
   // Code snippet that matched
@@ -40,6 +41,11 @@ pub(crate) struct Match {
   #[pyo3(get)]
   #[get = "pub"]
   matches: HashMap<String, String>,
+
+  #[pyo3(get)]
+  #[get]
+  #[get_mut]
+  associated_matches: HashMap<String, Range>,
 }
 gen_py_str_methods!(Match);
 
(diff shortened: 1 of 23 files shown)


src/models/matches.rs
```diff
@@ -49,20 +63,44 @@ impl Match {
...
+    let start_range = associated_ranges.first().cloned().unwrap_or(&self.range);
+    let end_range = associated_ranges.last().cloned().unwrap_or(&self.range);
+    (*start_range, *end_range)
+  }
+
+  /// Merge the associated matches of the given match into the current match.
+  /// It basically extends the range to include the first and last associated match of the given match.
+  pub(crate) fn expand_to_associated_matches(&mut self, code: &String) {
+    let (start_range, end_range) = self.get_first_and_last_associated_ranges();
+    if start_range.start_byte < self.range.start_byte {
```
lazaroclapp, line 95: Is there a case where `associated_ranges.first()` in `get_first_and_last_associated_ranges()` exists and yet it is after `self.range.start_byte`? Rather than before or equal? (Trying to figure out if the conditional checks here are needed or not.

src/models/matches.rs
```diff
@@ -80,6 +118,113 @@ impl Match {
...
+    loop {
+      // If we are looking for trailing elements, we start from the next sibling of the node
+      // Else we start from the previous sibling of the node
+      while let Some(sibling) = if trailing {
+        current_node.next_sibling()
+      } else {
+        current_node.prev_sibling()
+      } {
+        let content = sibling.utf8_text(code.as_bytes()).unwrap();
+        if !found_comma && content.trim().eq(",") {
```
lazaroclapp, line 148: Question: If we are removing leading and trailing commas, wouldn't deleting `b` in `foo(a, b, c)` lead to broken code like `foo(a c)` or are we relying on the whitespace to prevent that? And, if the later, what's an example of a comma before the element that we _do_ want to remove?

src/models/matches.rs
```diff
@@ -80,6 +118,113 @@ impl Match {
...
+      if self.overlaps(comment, &previous_node) {
+        return false;
+      }
+    }
+    // Check if there exists no node between the comment and the deleted node
+    if let Some(next_node) = comment.next_sibling() {
+      if next_node.start_byte() < deleted_node.start_byte() {
+        let (start_range, _) = self.get_first_and_last_associated_ranges();
+        if next_node.start_byte() < start_range.start_byte {
+          return false;
```
lazaroclapp, line 214: When do we hit this case? I thought we were going backwards from the deleted node and a) adding every comment node to the list used to calculate `get_first_and_last_associated_ranges()`, b) stopping on the first non-comment (non-comma) node we see. So we should never be calling this method with a non-comment node between the comment and the start of the current expanded range, no? What am I missing?

lazaroclapp (commented):
A few quick questions, but overall this LGTM and does remove a lot of extra work/duplicated code :)

ketkarameya:
Addressed all the comments.

ketkarameya (commented):
Addressed review comments.

test-resources/swift/leading_comma/input/Sample.swift
```diff
@@ -0,0 +1,15 @@
+enum TestEnum {
+   case one
+   case two
+   case three
+   case four
+
+   var v1: String {
+      switch self {
+      case .two, .one, .four:
```
Satyam1749, line 9: Can you please add a test case
```
input
case .two, .one, .four:
expected
case .two, .one:

Comment (ketkarameya) [reply test-resources/swift/leading_comma/input/Sample.swift:9]: done

File: src/models/matches.rs (hunk 2 above)
Comment (lazaroclapp) [reply src/models/matches.rs:149]: > Similarly if we find trailing comments we do not lookup leading comments.

Is this true? Or even what we want? I think we keep looking for comments even if we already found some. The only case in which we look at `found_comment` is to decide whether or not to expand our search to the parent node's siblings, and we still do that if we have found comments but not a comma. No?

This also seems to be reasonable, because we would expect both comments to be deleted in:

```
// The user type is used for [bla bla]
private UserType userType = DEFAULT_USER_TYPE; // DEFAULT_USER_TYPE == 0
```

src/models/matches.rs
lazaroclapp: Ah, basically, this exists because we jump to the parent without checking that the node being inspected is the first (non-comment) node inside said parent, correct?

lazaroclapp (approved):

src/models/matches.rs
ketkarameya: I stand corrected. 
If we find a trailing comment, we will still try to lookup a leading comment. 
Yes, found comment guards our `parent-level` search. 

I have updated our test cases to show that the current implementation handles the scenario you mention above: 

https://github.com/uber/piranha/pull/410/commits/d1c301c9cdc19b4f78a2ca55a2d8fc34b3216611
^ I added a comment on the line of enum (Which got cleaned up, hence the expected was not updated)

src/models/matches.rs
ketkarameya: yes correct. We do not do that check instead we have the check we are discussing above.

PR #417, opened 2023-03-29
Fixed release workflow for macOS

Satyam1749:
This PR addresses issue for release of polyglot on macOS, Polyglot depends on tree-sitter-python which tries to compile c++ files stdlibc++ which is depreciated in newer version of mac.

Passes commandline argument -stdlib=libc++ to fix the above mentioned issue.

Also, added strategy to run this workflow on both x86 and arm64 mac, Hence creating release for both platforms.

### DIFF
--- a/.github/workflows/polyglot_release.yml
+++ b/.github/workflows/polyglot_release.yml
@@ -8,7 +8,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: 3.9
+        python-version: 3.8
     - name: Build wheel with Maturin
       run: |
         pip install --upgrade maturin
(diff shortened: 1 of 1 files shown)


.github/workflows/polyglot_release.yml
```diff
@@ -18,29 +18,26 @@ jobs:
       run: |
         pip install twine
         twine upload --skip-existing -u __token__ -p ${{ secrets.PYPI_TOKEN }} dist/*
-  
-  build_mac_os:
+  macos_build:
```
ketkarameya, line 21: Can u ponder over the idea of adding another workflow to release from [older ubuntu versions](I would https://github.com/actions/runner-images)?
We want to specifically support debian 10 => ubuntu 18/20 idk. 

https://pypi.org/project/polyglot-piranha/0.3.0/#files These are our current releases for ubuntu. We have two of them. `_34` was produced from this `ubuntu-latest` architecture. But to support all internal uses cases we need to support `_28` architecture too (which i had manually released to Pypi from an internal debian 10 machine) . So, let's see if running on ubuntu 20/18 gives us this `_28` wheel.

Then the release will become fully automated . We can make a release daily. lol

ketkarameya (approved):
some comments. 
but 2 comments are blockers (new line and one adding comments to the file)

.github/workflows/polyglot_release.yml
Satyam1749: resolved

PR #418, opened 2023-03-29
Fixed function scope for swift

Satyam1749:
Function scope for swift was flaky as eq doesn't work in tree-sitter query for optional matched patterns. For example in this case the defined eq wouldn't work
```
(
   (simple_identifier)? @variable
   (#eq? @variable "var")
)
```

To resolve this, created separate matcher and generators for the below defined 4 cases.

```
class DummyClass{
    init(){
    }

    func dummyFunction(){
    }

    func dummyFunction(a:String) {
    
    }

    init(a: String) {

    }
}

```

### DIFF
--- a/src/cleanup_rules/swift/scope_config.toml
+++ b/src/cleanup_rules/swift/scope_config.toml
@@ -25,21 +25,52 @@ generator = """(
 name = "Function"
 [[scopes.rules]]
 matcher = """(
-    (function_declaration
-        name: (simple_identifier)? @name
-        (parameter)? @params
-        body: (function_body)
-    ) @function
+    [
+        (function_declaration
+            name: (simple_identifier) @m_function_with_param_name
+            (parameter) @m_function_with_param_params
+            body: (function_body)
+        )
+        (function_declaration
+            name: (simple_identifier) @m_function_without_param_name
+            .
+            body: (function_body)
+        )
+        (function_declaration
+            .
+            (parameter) @m_constructor_with_param_params
+            body: (function_body)
+        )
+        (function_declaration
+            .
+            body: (function_body) @m_constructor_without_param_body
+        )
+    ]@xdn
 )"""
 generator = """(
-    (function_declaration
-        name: (simple_identifier)? @func_name
-        (parameter)? @parameters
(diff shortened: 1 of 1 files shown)


src/cleanup_rules/swift/scope_config.toml
```diff
@@ -25,21 +25,52 @@ generator = """(
...
-    ) @function
+    [
+        (function_declaration
+            name: (simple_identifier) @m_function_with_param_name
+            (parameter) @m_function_with_param_params
+            body: (function_body)
+        )
+        (function_declaration
+            name: (simple_identifier) @m_function_without_param_name
+            .
```
ketkarameya, line 36: is this `.` necessary ? Same below. What does it do ?

ketkarameya, line 36: is this . necessary ? Same below. What does it do ?

src/cleanup_rules/swift/scope_config.toml
```diff
@@ -25,21 +25,52 @@ generator = """(
...
 matcher = """(
-    (function_declaration
-        name: (simple_identifier)? @name
-        (parameter)? @params
-        body: (function_body)
-    ) @function
+    [
+        (function_declaration
+            name: (simple_identifier) @m_function_with_param_name
+            (parameter) @m_function_with_param_params
```
ketkarameya, line 31:
```suggestion
            (parameter) @m_function_with_params
```

ketkarameya: rename ?

src/cleanup_rules/swift/scope_config.toml
```diff
@@ -25,21 +25,52 @@ generator = """(
...
+            body: (function_body)
+        )
+        (function_declaration
+            name: (simple_identifier) @m_function_without_param_name
+            .
+            body: (function_body)
```
ketkarameya, line 37: why are we matching the body to create the scope?
The body might change due to some `PARENT` edits.

ketkarameya (commented):
This looks neat to me. 
Left a few comments and questions.

PR #419, opened 2023-03-29
Fix Kotlin bug at handling if statement blocks containing comments and statements

ketkarameya:
We were not handling scenarios when blocks contained statements and comments.
The problem is that in tree-sitter-kotlin `comment` is not subtype of `statement`. Its a sibling type. (if it makes sense).
Therefore I had to make it `(_)*` because `(_)` would only match first statement(s) or comment(s).

----
* __->__ #419

### DIFF
--- a/src/cleanup_rules/kt/rules.toml
+++ b/src/cleanup_rules/kt/rules.toml
@@ -66,7 +66,7 @@ query = """
 (
 (if_expression (_) (_) 
                (control_structure_body ((if_expression [((boolean_literal) @condition) (parenthesized_expression (boolean_literal) @condition)] 
-               (control_structure_body (_) ) @consquent
+               (control_structure_body) @consquent
                (_) @alternative) @if_expression) )) @outer_if
 (#eq? @condition "false")
 )
@@ -86,7 +86,7 @@ query = """
 (
 (if_expression (_) (_) 
                (control_structure_body ((if_expression [((boolean_literal) @condition) (parenthesized_expression (boolean_literal) @condition)] 
-               (control_structure_body (_) ) @consquent)) ) @alternative) @if_expression
+               (control_structure_body) @consquent)) ) @alternative) @if_expression
 (#eq? @condition "false")
 )
 """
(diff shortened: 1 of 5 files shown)


src/cleanup_rules/kt/rules.toml
```diff
@@ -66,7 +66,7 @@ query = """
 (
 (if_expression (_) (_) 
                (control_structure_body ((if_expression [((boolean_literal) @condition) (parenthesized_expression (boolean_literal) @condition)] 
-               (control_structure_body (_) ) @consquent
```
ketkarameya, line 69: remove unnecessary match structure. (could be a different PR but ...)

src/cleanup_rules/kt/rules.toml
```diff
@@ -86,7 +86,7 @@ query = """
 (
 (if_expression (_) (_) 
                (control_structure_body ((if_expression [((boolean_literal) @condition) (parenthesized_expression (boolean_literal) @condition)] 
-               (control_structure_body (_) ) @consquent)) ) @alternative) @if_expression
```
ketkarameya, line 89: remove unnecessary match structure. (could be a different PR but ...)

src/cleanup_rules/kt/rules.toml
```diff
@@ -125,8 +125,8 @@ name = "simplify_if_true_with_alternative"
 query = """
 (
 (if_expression [((boolean_literal) @condition) (parenthesized_expression (boolean_literal) @condition)] 
-               (control_structure_body (_) @consequence ) 
-               (control_structure_body (_) ) @alternative) @if_expression
+               (control_structure_body (_)* @consequence )
```
ketkarameya, line 128: bug fix

src/cleanup_rules/kt/rules.toml
```diff
@@ -149,8 +149,8 @@ name = "simplify_if_false_with_alternative"
 query = """
 (
 (if_expression [((boolean_literal) @condition) (parenthesized_expression (boolean_literal) @condition)] 
-               (control_structure_body (_)? ) @consquent
-               (control_structure_body (_)? @alternative) ) @if_expression
+               (control_structure_body) @consquent
+               (control_structure_body (_)* @alternative) ) @if_expression
```
ketkarameya, line 153: bug fix

ketkarameya (commented):
Annotated the PR

lazaroclapp (approved):
I think I get it and the tests look good, so 👍

PR #420, opened 2023-03-30
Variable Inlining [1/n]: Local variable inlining

Adarsh-NP:
This PR is part-1 of the Variable inlining PRs. This aims to solve the problem of local variable inlining. The tests written are only testing for local variable deletion and replacement. Integration tests for resolving the booleans in if else and other statements will be done in upcoming PRs. Partially solves #382

### DIFF
--- a/src/cleanup_rules/swift/edges.toml
+++ b/src/cleanup_rules/swift/edges.toml
@@ -12,6 +12,12 @@
 
 # The edges in this file specify the flow between the rules.
 
+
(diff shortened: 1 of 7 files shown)


Adarsh-NP:
Can be landed after Satyam's PR get's landed. Please review my commit only. Cherry picking my commit and opening a PR from a different branch can cause test failures as Satyam's PR is important for setting up the Function scope.

PR #433, opened 2023-04-04
Add enum scope for Java

ketkarameya:
This PR adds enum as "Class" scope for Java. 
Test Plan: Added a test case 

Kotlin AST treats enums and classes as `(class_declaration)` so we do not need to update the "Class" scope query for Kotlin

### DIFF
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -150,23 +150,21 @@ fn execute_piranha_and_check_result(
 /// # Arguments:
 /// * test_name: Name of the test (identifier)
 /// * relative_path: relative path such that `test-resources/<language>/<relative_path>` leads to a directory containing the folders `input` and `configurations`
-/// * expected_number_of_matches: expression returning the expected number of matches
-/// * matches_frequency: The expected frequency of the matches
+/// * matches_frequency: The expected frequency for each match
 ///
 /// Usage:
 /// ```
 /// create_match_tests! {
 ///  "java",
-///  test_a1:  "relative/path_1", 2;
-///  test_a2:  "relative/path_2", 3
-///  HashMap::from([("match_class", 2)])
+///  test_a1:  "relative/path_1", HashMap::from([("match_class", 2));
+///  test_a2:  "relative/path_2", HashMap::from([("match_class", 2), ("match_class_1", 1)])
+///  
 /// ;
 /// }
 /// ```
 macro_rules! create_match_tests {
   ($language: expr,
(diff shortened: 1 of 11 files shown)


test-resources/java/structural_find/configurations/edges.toml
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 1: 2023

test-resources/java/structural_find/input/SampleEnum.java
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright (c) 2022 Uber Technologies, Inc.
```
lazaroclapp, line 2:
```suggestion
 * Copyright (c) 2023 Uber Technologies, Inc.
```

test-resources/java/structural_find/input/SampleEnum.java
```diff
@@ -0,0 +1,33 @@
...
+            return "bar";
+          }
+    }
+}
+
+public class SomeClass {
+    public void someMethod(){
+        System.out.println("Hello World!");
+    }
+}
```
lazaroclapp, line 33: So, the reason there isn't an `expected/` output for this test case, is that this is a find rule and not any sort of transformation, correct? And our testing infra only tests the total number of matches over all tests, rather than specifically how many matches on each test or where?

test-resources/java/structural_find/input/SampleEnum.java
ketkarameya: https://github.com/uber/piranha/pull/434

src/tests/mod.rs
```diff
@@ -133,19 +151,23 @@ fn execute_piranha_and_check_result(
 /// * test_name: Name of the test (identifier)
 /// * relative_path: relative path such that `test-resources/<language>/<relative_path>` leads to a directory containing the folders `input` and `configurations`
 /// * expected_number_of_matches: expression returning the expected number of matches
+/// * matches_frequency: The expected frequency of the matches
```
lazaroclapp, line 154:
```suggestion
/// * matches_frequency: The expected frequency for each of the matches
```

To make things more clear.

Also, do we still need `expected_number_of_matches` or is that just computable from some sort of sum of `matches_frequency`?

Edit: Looking below, I think this does make the `expected_number_of_matches` argument obsolete and we should remove it.

lazaroclapp (approved):
LGTM!

PR #446, opened 2023-04-21
Add Kotlin rules to simplify equality and inequality wherever possible

ketkarameya:
* (to be filled)

### DIFF
--- a/src/cleanup_rules/kt/rules.toml
+++ b/src/cleanup_rules/kt/rules.toml
@@ -374,6 +374,54 @@ replace = "@rhs"
 replace_node = "disjunction_expression"
 is_seed_rule = false
 
+
+[[rules]]
+groups = ["boolean_expression_simplify"]
+name = "simplify_equality_true"
+query = """(
+(equality_expression (_)@l1 "==" (_)@l2) @equality
+(#eq? @l1 @l2)
+)"""
+replace = "true"
+replace_node = "equality"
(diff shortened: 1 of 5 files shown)


src/cleanup_rules/kt/rules.toml
```diff
@@ -374,6 +374,73 @@ replace = "@rhs"
...
+replace_node = "equality"
+is_seed_rule = false
+
+# Before
+# true != true
+# After
+# false
+[[rules]]
+groups = ["boolean_expression_simplify"]
+name = "simplify_in_equality_false"
```
lazaroclapp, line 418:
```suggestion
name = "simplify_inequality_false"
```

This rule is about `!=`, so inequality rather than "in equality", I think.

src/cleanup_rules/kt/rules.toml
```diff
@@ -374,6 +374,73 @@ replace = "@rhs"
...
+is_seed_rule = false
+
+
+# Before
+# true != false
+# After
+# true
+[[rules]]
+groups = ["boolean_expression_simplify"]
+name = "simplify_in_equality_true"
```
lazaroclapp, line 434:
```suggestion
name = "simplify_inequality_true"
```

Same.

lazaroclapp (approved):
Minor nit about rule naming, otherwise LGTM.

PR #449, opened 2023-04-25
Update commit hash to include constructor match fix in swift

Adarsh-NP:
This PR updates the commit hash pointing to @Satyam1749's tree-sitter-swift fork. The new commit in his fork includes the provision of catching constructor `init` nodes in swift.


Satyam1749 (approved):

Satyam1749:
Changes LGTM, @ketkarameya Can you please take a look.

ketkarameya (commented):
I see you have updates `rules.toml` for some tests, but why is there no change in the corresponding `.swift` files?

Adarsh-NP:
> I see you have updates `rules.toml` for some tests, but why is there no change in the corresponding `.swift` files?

They already had constructor declarations that we are now catching with `init_declaration` which were earlier matching using `function_declaration`, hence no changes there.

ketkarameya:
> > I see you have updates `rules.toml` for some tests, but why is there no change in the corresponding `.swift` files?
> 
> They already had constructor declarations that we are now catching with `init_declaration` which were earlier matching using `function_declaration`, hence no changes there.

But if we were matching them using `function_declaration` why is there a new rule ? And not a rule update?

PR #450, opened 2023-04-25
Add if boolean, (_) cleanup rules

Adarsh-NP:
This PR solves the following categories of refactorings for Swift:

- if true, variable { } => if variable { }
- if variable, true { } => if variable { }
- if true, f() { } => if f() { }
- if f(), true { } => if f() { }
- if true, variable == f1() { } => if variable == f1() { }
- if variable == f1(), true { } => if variable == f1() { }
- if true, variable == someOtherVariable { } => if variable == someOtherVariable { }
- if variable == someOtherVariable, true { } => if variable == someOtherVariable { }
- if true, let variable = f1() { } => if let variable = f1() { }
- if let variable = f1(), true { } => if let variable = f1() { }
- if true, let variable == f1() { } => if let variable == f1() { }
- if let variable == f1(), true { } => if let variable = f1() { }

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -386,10 +386,10 @@ is_seed_rule = false
 
 # Before:
 # if true, anything { }
-# if true, someOptional { }
+# if anything, true { }
 # After: 
-# if someOptionalCall() { }
-# if someOptional { }
+# if anything { }
+# anything: [variable, function_call, variable_assignment, call_expression, equality check]
 [[rules]]
 name = "if_let_true_cleanup_literal_checks"
 query = """(
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -389,7 +389,8 @@ is_seed_rule = false
 # if anything, true { }
 # After: 
 # if anything { }
-# anything: [variable, function_call, variable_assignment, call_expression, equality check]
+# anything: [ variable, function_call, variable_assignment (to a function's return value, optional variable),
+# equality check (with a function's return value, variable ]
 [[rules]]
 name = "if_let_true_cleanup_literal_checks"
 query = """(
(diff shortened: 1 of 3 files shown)


test-resources/swift/cleanup_rules/expected/SampleClass.swift
```diff
@@ -163,4 +163,56 @@ class SampleClass {
         var value = 2
         var value2 = 3
     }
+
+    func checkIfShortCircuitStatementsWithBooleanPrefix() {
```
ketkarameya, line 167: Let's not add this to this file. Let's make a separate swift test case? This will add up to the slow down. 
Adding a separate test case will enable us to leverage `cargo test`  level parallelism

src/cleanup_rules/swift/rules.toml
```diff
@@ -383,6 +383,40 @@ replace_node = "if_else_block"
 replace = "@alternatives"
 is_seed_rule = false
 
+
+# Before:
+# if true, anything { }
+# if true, someOptional { }
```
ketkarameya: Let's add the construct matched by the third clause in the query

src/cleanup_rules/swift/rules.toml
```diff
@@ -383,6 +383,40 @@ replace_node = "if_else_block"
...
+# After: 
+# if someOptionalCall() { }
+# if someOptional { }
+[[rules]]
+name = "if_let_true_cleanup_literal_checks"
```
ketkarameya, line 394: i dont think all rules here are related to `let` . Can we have a better name ?

ketkarameya (changes requested):

ketkarameya:
Why does the PR description only contain lhs ? Can u put the RHS too ?

Adarsh-NP:
> Why does the PR description only contain lhs ? Can u put the RHS too ?

Done

src/cleanup_rules/swift/rules.toml
ketkarameya: Here in this rule there are three choices -`@ifstl` . We have documented two patterns here in the comment, can we document the missing one ?

ketkarameya (approved):
Ignore my comments related to creating smaller test cases. Things LGTM.

PR #451, opened 2023-04-25
Add new rules for Thrift RPC codes

rajbarik:
This PR adds new rules to Thrift. It adds more tests in sample.thrift to cover a number of refactoring scenarios. It also separates thrift demos from other demos by updating find_replace_demos.py file.


ketkarameya (approved):
LGTM

PR #452, opened 2023-04-26
Variable Inline [1/n]: Local variable inline

Adarsh-NP:
This PR aims to solve Variable inlining for variables within the function scope. Please take a look at the tests to understand the various cases. 

* #455
* #453
* __->__ #452

### DIFF
--- a/src/cleanup_rules/swift/edges.toml
+++ b/src/cleanup_rules/swift/edges.toml
@@ -24,7 +24,7 @@ from = "replace_expression_with_boolean_literal"
 to = ["boolean_literal_cleanup", "variable_inline_cleanup"]
 
 [[edges]]
-scope = "File"
+scope = "Class"
 from = "variable_inline_cleanup"
 to = ["delete_variable_declaration"]
 
--- a/src/cleanup_rules/swift/edges.toml
+++ b/src/cleanup_rules/swift/edges.toml
@@ -12,10 +12,21 @@
 
 # The edges in this file specify the flow between the rules.
 
+
+[[edges]]
+scope = "Function"
+from = "delete_variable_declaration"
+to = ["replace_identifier_with_value", "delete_parent_assignment"]
+
 [[edges]]
 scope = "Parent"
 from = "replace_expression_with_boolean_literal"
-to = ["boolean_literal_cleanup"]
+to = ["boolean_literal_cleanup", "variable_inline_cleanup"]
+
+[[edges]]
+scope = "File"
+from = "variable_inline_cleanup"
+to = ["delete_variable_declaration"]
 
 [[edges]]
 scope = "Parent"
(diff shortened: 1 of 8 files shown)


src/cleanup_rules/swift/edges.toml
```diff
@@ -12,10 +12,21 @@
...
+to = ["replace_identifier_with_value", "delete_parent_assignment"]
+
 [[edges]]
 scope = "Parent"
 from = "replace_expression_with_boolean_literal"
-to = ["boolean_literal_cleanup"]
+to = ["boolean_literal_cleanup", "variable_inline_cleanup"]
+
+[[edges]]
+scope = "File"
```
ketkarameya, line 27: Why file? Why not `Parent`?

ketkarameya, line 27: Why `File` and not `Parent` ?

src/cleanup_rules/swift/rules.toml
```diff
@@ -513,6 +513,155 @@ replace_node = "ternary_block"
 replace = "@block_false"
 is_seed_rule = false
 
+# delete variables declared in a function scope
+[[rules]]
+name = "delete_variable_declaration"
+query = """(
+    (function_declaration
```
ketkarameya, line 520: why do u need to match a `function_declaration`?

src/cleanup_rules/swift/rules.toml
```diff
@@ -513,6 +513,155 @@ replace_node = "ternary_block"
...
+        value: (boolean_literal) @value
+    )
+    (#eq? @variable "@identifier")
+)"""]
+
+# skip the rule if there is an assignment of the same variable
+[[rules.constraints]]
+matcher = "(class_declaration) @cd"
+queries = ["""(
+    (navigation_expression
```
ketkarameya, line 656: No sure if the comment and constraint match. 

Anyhoo, we can take also combine this with previous constraint (Since they use the same `matcher` like my recommendation for previous rule)

test-resources/swift/variable_inline/local_variable_inline/input/VariableInlining.swift
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
```
ketkarameya, line 1:
```suggestion
// Copyright (c) 2023 Uber Technologies, Inc.
```

ketkarameya (changes requested):

src/cleanup_rules/swift/rules.toml
Adarsh-NP: > why do u need to match a `function_declaration`?

This is because this rule is meant to detect just the function variable declaration and not any kind of declaration.

ketkarameya: But then either add a constraint

```
[[rules.constraints]]
matcher = "(function_declaration) @fd"
```

Adarsh-NP: How about I do this in a follow up PR? If this works, it can save from writing different rules for adhoc declarations.

PR #458, opened 2023-05-01
Variable Inlining [2/n]: Field Variable Inlining

Adarsh-NP:
This PR aims to solve variable inlining for field variables of a class. Please refer to the tests to understand the various cases it solves.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #459
* __->__ #458

PR #459, opened 2023-05-01
Var Inline [3/n]: Ad hoc var inline

Adarsh-NP:
Add tests for adhoc variable inline. The rules are not there because local variable inline automatically handles the adhoc declaration anywhere inside the function due to the matcher `function_declaration` in constraint without any query. 

* __->__ #459
* #458


ketkarameya (approved):
LGTM. Just adding tests

PR #464, opened 2023-05-03
Cleanup Rules Swift: Fix the issue in short circuit rule

Adarsh-NP:
Short circuit cleanup rules were not running properly in case of multiple cleanups that were caught as statements instead and were causing the rules to not replace the nodes properly. 

This PR adds the specific nodes in query to solve the issue for short circuit if statements.


ketkarameya:
In general this PR makes sense to me. But I do not understand the PR description.

PR #467, opened 2023-05-04
Ignore long running test cases

ketkarameya:
(Ok. I understand this is hacky, but this is the practice that they follow inside rust lang - [example](https://github.com/rust-lang/rust/blob/054b7b7/src/libcoretest/num/flt2dec/strategy/grisu.rs#L56) )

Context: 
Due to grammar issue, the tree-sitter swift queries take a lot of time to load (we are looking into it with the upstream owner). However, this implies that these swift tests cause a significant slow down in local development (was 20s before `swift` stuff was added now it is 2-4mins, which i a lot if u think about it).

So I am ignoring these tests. 
We will run these test in CI though (As updated in github workflow). These tests can be ran locally too by adding `--include-ignored` flag.

Oh, also! Instead of adding `'''ignore...` i use `doctest = false` in `Cargo.toml` (because I am ignoring all doc tests)

-----

* #468
* __->__ #467

### DIFF
--- a/src/models/constraint.rs
+++ b/src/models/constraint.rs
@@ -72,7 +72,7 @@ impl Constraint {
 ///
 /// Usage:
 ///
-/// ```ignore
+/// ```
 /// constraint! {
 ///   matcher = "(method_declaration) @md".to_string(),
 ///   queries=  ["(method_invocation name: (_) @name) @mi".to_string()]
@@ -81,7 +81,7 @@ impl Constraint {
 ///
 /// expands to
 ///
-/// ```ignore
+/// ```
 /// ConstraintBuilder::default()
 ///      .matcher("(method_declaration) @md".to_string())
 ///      .queries(vec!["(method_invocation name: (_) @name) @mi".to_string()])
--- a/src/tests/test_piranha_swift.rs
+++ b/src/tests/test_piranha_swift.rs
@@ -11,14 +11,18 @@ Copyright (c) 2023 Uber Technologies, Inc.
  limitations under the License.
 */
 
-use super::{create_rewrite_tests, substitutions};
+use std::path::PathBuf;
 
-use crate::models::default_configs::SWIFT;
+use super::{create_rewrite_tests, execute_piranha_and_check_result, substitutions};
+
+use crate::models::{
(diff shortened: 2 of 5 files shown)


src/models/constraint.rs
```diff
@@ -72,7 +72,7 @@ impl Constraint {
 ///
 /// Usage:
 ///
-/// ```ignore
+/// ```
```
lazaroclapp, line 75: What's this change doing? No longer ignoring these blocks for formatting?

src/tests/test_piranha_swift.rs
```diff
@@ -32,22 +36,102 @@ create_rewrite_tests! {
...
+    .join("variable_inline/local_variable_inline");
+  let temp_dir = super::copy_folder_to_temp_dir(&_path.join("input"));
+
+  let piranha_arguments = PiranhaArgumentsBuilder::default()
+    .path_to_codebase(temp_dir.path().to_str().unwrap().to_string())
+    .path_to_configurations(_path.join("configurations").to_str().unwrap().to_string())
+    .language(PiranhaLanguage::from(SWIFT))
+    .cleanup_comments(true)
+    .delete_file_if_empty(false)
+    .build();
```
lazaroclapp, line 88: This part repeats between the three tests below. Rather than a macro, my instinct here would be to have non-test helper function that you call as `make_default_piranha_arguments_with_path("variable_inline/local_variable_inline")` in this same file. Or even `execute_piranha_and_check_result_with_default_arguments(...)` if removing `temp_dir` is hard otherwise.

This is up to you, style-wise, though. Since for tests I'd usually lean on the side of test cases being clear vs avoiding repetition :)

lazaroclapp (commented):
Optional nit and one question about removed ignores in docstrings. Overall, LGTM as soon as I understand the removals :)

src/models/constraint.rs
ketkarameya: This `ignore` implies that cargo should ignore docstring tests. 
Now i specify that explicitly in the `Cargo.toml` file

lazaroclapp (approved):
LGTM if CI passes!

PR #472, opened 2023-05-16
Fix Conditional Statements: Add nodes to capture comments

Adarsh-NP:
This PR fixes the issue when the if cleanup rules were not working due to comments inside the blocks. Added comment nodes as optional to cover those scenarios.

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -250,7 +250,8 @@ name = "if_always_true"
 query = """ (
 (if_statement
     condition: (boolean_literal) @condition_literal
-    .(statements) @if_block
+    ((comment)*
+    .(statements)) @if_block
     ) @if_else_block
 (#eq? @condition_literal "true")
 )"""
@@ -275,7 +276,8 @@ query = """ (
 (else)
 (if_statement
     condition: (boolean_literal) @condition_literal
-    .(statements) @if_block
+    ((comment)*
+    .(statements)) @if_block
     ) @if_else_block
 (#eq? @condition_literal "true")
 )"""
@@ -299,8 +301,11 @@ query = """(
 (if_statement
     bound_identifier: (_)?
     condition: (boolean_literal) @condition_literal
-    .(statements) @if_block
-    [(statements) (if_statement)]? @alternatives 
+    ((comment)*
+    .(statements)) @if_block
+    ((comment)*
+    [(statements)
+     (if_statement)])? @alternatives 
     ) @if_else_block
 (#eq? @condition_literal "false")
 )"""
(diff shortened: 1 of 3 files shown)


src/cleanup_rules/swift/rules.toml
```diff
@@ -250,7 +250,8 @@ name = "if_always_true"
 query = """ (
 (if_statement
     condition: (boolean_literal) @condition_literal
-    .(statements) @if_block
+    ((comment)*
```
ketkarameya, line 253: how about just `(_)*`

test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -214,6 +214,62 @@ class SampleClass {
         var value2 =  placeholder_false ? 2 : 3
     }
 
+    func checkIfBooleanWithComments(){
+         if !TestEnum.stale_flag_one.isEnabled {
+            // to be deleted
+            toBeDeleted1()
+         }else{
+            // to be preserved
```
ketkarameya, line 222:
```suggestion
            // to be preserved 1
```

test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -214,6 +214,62 @@ class SampleClass {
...
+            toBePreserved1()
+         }
+
+         if !TestEnum.stale_flag_one.isEnabled{
+            // to be deleted
+            toBeDeleted2() 
+         } else if a {
+            // to be preserved
```
ketkarameya, line 230:
```suggestion
            // to be preserved 2
```

similarly for others for the sake of consistency

ketkarameya (commented):
Do u really need an alternation between comment and statement? Would wildcard not work ?

src/cleanup_rules/swift/rules.toml
Adarsh-NP: We want to capture the first (statements) block separately. If we do just (_)* we will catch all the if else cases since else and else if are also considered as children of the first if.

ketkarameya: But then this does not seem a sound rule either. How abut altering the grammar for swift to fix these `if-else` awkwardness from the root ?

Adarsh-NP: Waiting for that sounds a little unreasonable to me. Also, we would need to confirm with @Satyam1749 because if I remember he wanted to pick that up but it was an optional improvement.

ketkarameya: Its alright if you dont want to take this up, we will not land this either . When I get time i will look into it

Adarsh-NP: How about we merge this while we look to solve that? Once solved, we can revert this change because if we solve grammar a lot of rules will have to be re-written anyway. Especially these if else cleanups.

Satyam1749: @ketkarameya `(comments)* .(statements)` is a full proof way to handle if-blocks with current grammar and there can't be an edge case to it. So, in my opinion we should go ahead with this for now. Since, fixing the tree-sitter grammar is a 3rd party dependency. 

This usecase is a pretty basic usecase so we shouldn't keep this blocked on 3rd party dependency.

ketkarameya (approved):

PR #473, opened 2023-05-16
Fix demo files for the new PiranhaArguments signature

danieltrt:
I was trying to run the demos, but they were failing w/

```INFO root 2023-05-16 14:44:31,514 find_replace_demos.py:15 Running the Find/Replace demo for Swift
Traceback (most recent call last):
  File "/Users/danielrr/piranha/demo/find_replace_demos.py", line 72, in <module>
    swift_demo()
  File "/Users/danielrr/piranha/demo/find_replace_demos.py", line 22, in swift_demo
    args = PiranhaArguments(
TypeError: PiranhaArguments.__new__() got multiple values for argument 'path_to_codebase'
```

I'm guessing they were never updated to work with the new python bindings? It should be working now.
For reference I just used these rules:

```
comby "PiranhaArguments(:[lang], {:[subs]}, path_to_codebase=:[path], :[rest])" "PiranhaArguments(:[lang], :[path], substitutions={:[subs]}, :[rest])" -f .py
comby "PiranhaArguments(:[lang], {:[subs]}, :[rest], path_to_codebase=:[path], :[rest1])" "PiranhaArguments(:[lang], :[path], :[rest], substitutions={:[subs]}, :[rest1])" -f .py
comby "PiranhaArguments(:[lang], {:[subs]}, :[rest], path_to_codebase=:[path])" "PiranhaArguments(:[lang], :[path], :[rest], substitutions={:[subs]})" -f .py
```

### DIFF
--- a/demo/find_replace_custom_cleanup_demos.py
+++ b/demo/find_replace_custom_cleanup_demos.py
@@ -24,7 +24,7 @@ def java_demo():
 
     args = PiranhaArguments(
         "java",
-        {
+        substitutions={
             "input_type_name": "ArrayList",
         },
         path_to_configurations=configuration_path,
@@ -56,7 +56,7 @@ def python_demo():
 
     args = PiranhaArguments(
         "py",
-        {
+        substitutions={
             "str_literal": "dependency2",
             "str_to_replace": "dependency1",
             "str_replacement": "dependency1_1"
--- a/demo/find_replace_demos.py
+++ b/demo/find_replace_demos.py
@@ -21,7 +21,7 @@ def swift_demo():
 
     args = PiranhaArguments(
         "swift",
-        {
+        substitutions={
             "stale_flag_name": "test_second_experiment",
         },
         path_to_configurations=configuration_path,
@@ -52,7 +52,7 @@ def java_demo():
 
     args = PiranhaArguments(
         "java",
-        {
+        substitutions={
             "stale_flag_name": "STALE_FLAG",
         },
         path_to_configurations=configuration_path,
--- a/demo/match_only_demos.py
+++ b/demo/match_only_demos.py
@@ -11,7 +11,7 @@ def java_demo():
     info("Running the Match-only demo for Java")
     args = PiranhaArguments(
         "java",
-        {},
+        substitutions={},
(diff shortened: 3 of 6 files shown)


ketkarameya:
Thanks for pointing this out. Indeed i broke the code in a recent [PR](https://github.com/uber/piranha/pull/468). 

Actually, the change should be the other way around. We want to have `path_to_codebase` to be optional because we support `code_snippet` and `path_to_codebase`. So making it a positional argument is the bug i introduced :| 

So i guess you shud make `path_to_codebase`  as `Optional<String>` here (like it was before) - https://github.com/uber/piranha/blob/master/src/models/piranha_arguments.rs#L180 

(u might have to revert back to unwrapping `path_tocodebase` inside `py_new` (look here https://github.com/uber/piranha/pull/468/files#diff-2175302417330a100b6daa4d5e976745140033d8594d9f5194aa6fc087f9c4d7L186) basically undo this `path_to_codebase` part :| .
and then adapt the `pyi` file accordingly too.

danieltrt:
> Thanks for pointing this out. Indeed i broke the code in a recent [PR](https://github.com/uber/piranha/pull/468).
> 
> Actually, the change should be the other way around. We want to have `path_to_codebase` to be optional because we support `code_snippet` and `path_to_codebase`. So making it a positional argument is the bug i introduced :|
> 
> So i guess you shud make `path_to_codebase` as `Optional<String>` here (like it was before) - https://github.com/uber/piranha/blob/master/src/models/piranha_arguments.rs#L180
> 
> (u might have to revert back to unwrapping `path_tocodebase` inside `py_new` (look here https://github.com/uber/piranha/pull/468/files#diff-2175302417330a100b6daa4d5e976745140033d8594d9f5194aa6fc087f9c4d7L186) basically undo this `path_to_codebase` part :| . and then adapt the `pyi` file accordingly too.

I changed it back. Is this good? I also had to change the order of the arguments for the py_new to make the demos work

ketkarameya:
Please make the change like we discussed offline

danieltrt:
> Please make the change like we discussed offline

done

ketkarameya (approved):
LGTM.

PR #474, opened 2023-05-17
Swift: Fix Equality conditions evaluation

Adarsh-NP:
This PR fixes the equality expression evaluation where both lhs and rhs were boolean literals. The solution is to evaluate the equality and replace with the boolean literal so that it can be cleaned up by the other rules already in place.

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -874,6 +874,214 @@ not_contains = [
     )""",
 ]
 
+[[rules]]
+name = "de_parenthesize_literals"
+query = """(
+    (tuple_expression
+        [
+            (integer_literal)
+            (hex_literal)
+            (oct_literal)
+            (bin_literal)
+            (real_literal)
+            (boolean_literal)
+            (line_string_literal)
+            (regex_literal)
+        ] @tvalue
+    ) @texp
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "texp"
+replace = "@tvalue"
+is_seed_rule = false
+#
+# Before 
+# equality expression with same lhs and rhs for following:
+# integer_literal
+# hex_literal
+# oct_literal
+# bin_literal
+# real_literal
+# boolean_literal
+# _string_literal
+# regex_literal
+# After 
+#   true
+#
+[[rules]]
+name = "boolean_equality_expression_same_literals"
+query = """(
+    (equality_expression
+            lhs: [
+                (integer_literal)
+                (hex_literal)
+                (oct_literal)
+                (bin_literal)
+                (real_literal)
+                (boolean_literal)
+                (line_string_literal)
+                (regex_literal)
+            ] @left
+            rhs: [
+                (integer_literal)
+                (hex_literal)
+                (oct_literal)
+                (bin_literal)
+                (real_literal)
+                (boolean_literal)
+                (line_string_literal)
+                (regex_literal)
+            ] @right
+    ) @cond
+    (#eq? @left @right)
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "cond"
+replace = "true"
+is_seed_rule = false
+
+#
+# Before 
+# equality expression with different lhs and rhs for following:
+# integer_literal
+# hex_literal
+# oct_literal
+# bin_literal
+# real_literal
+# boolean_literal
+# _string_literal
+# regex_literal
+# After 
+#   false
+#
+[[rules]]
+name = "boolean_equality_expression_opposite_literals"
+query = """(
+    (equality_expression
+            lhs: [
+                (integer_literal)
+                (hex_literal)
+                (oct_literal)
+                (bin_literal)
+                (real_literal)
+                (boolean_literal)
+                (line_string_literal)
+                (regex_literal)
+            ] @left
+            rhs: [
+                (integer_literal)
+                (hex_literal)
+                (oct_literal)
+                (bin_literal)
+                (real_literal)
+                (boolean_literal)
+                (line_string_literal)
+                (regex_literal)
+            ] @right
+    ) @cond
+    (#not-eq? @left @right)
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "cond"
+replace = "false"
+is_seed_rule = false
+
+#
+# Before 
+# inequality expression (infix expression with custom_operator = '!='') with same lhs and rhs for following:
+# integer_literal
+# hex_literal
+# oct_literal
+# bin_literal
+# real_literal
+# boolean_literal
+# _string_literal
+# regex_literal
+# After 
+#   false
+#
+[[rules]]
+name = "boolean_inequality_expression_same_literals"
+query = """(
+    (infix_expression
+            lhs: [
+                (integer_literal)
+                (hex_literal)
+                (oct_literal)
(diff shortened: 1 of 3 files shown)


src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,90 @@ queries = [
...
+#   true
+#
+[[rules]]
+name = "boolean_equality_expression_both_true"
+query = """(
+    (equality_expression
+            lhs: (boolean_literal) @left
+            rhs: (boolean_literal) @right
+    ) @cond
+    (#eq? @left "true")
```
ketkarameya, line 890:
```suggestion
    (#eq? @left @right)
```

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,90 @@ queries = [
...
+    (#eq? @right "true")
```
ketkarameya, line 891:
```suggestion

```

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,90 @@ queries = [
...
+is_seed_rule = false
+
+#
+# Before 
+#   false == false
+# After 
+#   true
+#
+[[rules]]
+name = "boolean_equality_expression_both_false"
```
ketkarameya, line 905: no need for this rule, if u update like above.

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,90 @@ queries = [
...
+# Before 
+#   true == false
+# After 
+#   true
+#
+[[rules]]
+name = "boolean_equality_expression_true_false"
+query = """(
+    (equality_expression
+            lhs: (boolean_literal) @left
```
ketkarameya, line 929: introduce alternation here such that we match all literals

ketkarameya (changes requested):
Please look at a the kotlin complement of this PR 
https://github.com/uber/piranha/pull/446/files#diff-ee59591e1b8caa4e0f8b1a27f264d292291a7ab0e55d5acc9640e9541bf4d51f 

Try to keep rules as close to these as possible

src/cleanup_rules/swift/rules.toml
Adarsh-NP: Oh this is nice, I had the misconception that rhs of eq has to be a hole variable. This will simplify it.

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,150 @@ not_contains = [
     )""",
 ]
 
+#
+# Before 
+#   true == true
+#   (true) == true
+#   true == (true)
```
ketkarameya, line 881: can u write a rule to simplify a parenthesized expression to a non paramthesized expression ? 
The rule will become simpler then (look at Kotlin and Java rules). Basically one rule transforms `(true)` to `true`, then these equality / inequality rules do not have to account for parenthesis

ketkarameya: can u please add support for any expression for equaliity (look at the kotlin/java) rules . Equality rules do not have to be restricted to literals .

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,150 @@ not_contains = [
...
+#   (false) == true
+#   true == (false)
+#   false == (true)
+#   (false) == (true)
+#   (true) == (false)
+# After 
+#   false
+#
+[[rules]]
+name = "boolean_equality_expression_opposite_literals"
```
ketkarameya, line 926: Here please add support for all literals supported by swift - string, number, boolean, ...

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,150 @@ not_contains = [
...
+#   false != (false)
+#   (false) != (false)
+# After 
+#   false
+#
+[[rules]]
+name = "boolean_inequality_expression_same_literals"
+query = """(
+    (infix_expression
+            lhs: [(boolean_literal) @left
```
ketkarameya, line 964: same as above

src/cleanup_rules/swift/rules.toml
```diff
@@ -874,6 +874,150 @@ not_contains = [
...
+#   false != (false)
+#   (false) != (false)
+# After 
+#   false
+#
+[[rules]]
+name = "boolean_inequality_expression_opposite_literals"
+query = """(
+    (infix_expression
+            lhs: [(boolean_literal) @left
```
ketkarameya, line 1001: same as above

ketkarameya (commented):
.

ketkarameya (commented):
.

src/cleanup_rules/swift/rules.toml
```diff
@@ -890,6 +890,153 @@ not_contains = [
     )""",
 ]
 
+[[rules]]
+name = "de_parenthesize_literals"
+query = """(
+    (tuple_expression
+        (_) @tvalue
```
ketkarameya, line 897: let's only restrict this to literals 
Can we also have some documentation (before/after example) for this rule ?

src/cleanup_rules/swift/rules.toml
```diff
@@ -890,6 +890,153 @@ not_contains = [
...
+    ) @texp
+)"""
+groups = ["boolean_expression_simplify"]
+replace_node = "texp"
+replace = "@tvalue"
+is_seed_rule = false
+#
+# Before 
+#   true == true
```
ketkarameya, line 906: update comment. (do the same for all the rules )

test-resources/swift/cleanup_rules/expected/SampleClass.swift
```diff
@@ -251,4 +251,82 @@ class SampleClass {
...
+        if d {
+            toBePreservedEquality12a()
+        } else {
+        toBePreservedEquality12b()
+        }
+    }
+
+    func refactorBooleanLiteralInequalityExpressionInIfConditions(){
+
+        toBePreservedInequality1()
```
ketkarameya, line 296: can we have a atleast 1/2 test cases testing the other literals (i.e. string, number, ...) ?

src/cleanup_rules/swift/rules.toml
```diff
@@ -890,6 +890,153 @@ not_contains = [
...
+name = "boolean_inequality_expression_opposite_literals"
+query = """(
+    (infix_expression
+            lhs: [(boolean_literal) 
+                  (integer_literal)
+                  (line_string_literal)
+                 ] @left
+            op: (custom_operator) @ops
+            rhs: [(boolean_literal) 
+                  (integer_literal)
```
ketkarameya, line 1028: Swift has 9 kinds of literals - https://github.com/alex-pinkus/tree-sitter-swift/blob/main/grammar.js#L272

ketkarameya (changes requested):
Left comments inline

ketkarameya (changes requested):
Left comments inline

ketkarameya (changes requested):
Left comments inline

ketkarameya (approved):
LGTM

ketkarameya (approved):
LGTM

ketkarameya (approved):
LGTM

PR #478, opened 2023-05-19
Add contains, at least and at most to the the filter language

danieltrt:
I added the new functionality and a few new tests for contains.

I also added a new rule to delete interfaces based on contains, just to make sure everything was working fine on the python side. Let me know what you think

Edit: I realized `not-contains` is equivalent to doing `contains` with `at-most = 0`. I think keeping both constructs is a good idea anyway


tests/tests.py
```diff
@@ -137,6 +135,68 @@ def test_insert_field_add_import():
...
+                    (super_interfaces (type_list (type_identifier) @it)
+                    (#eq? @it "FooBar"))
+                )""",],
+            )
+        ]),
+    )
+
+    # This is wrong, should check the arity for implements :|.
+    # Works in this example
+    delete_implements = Rule (
```
danieltrt, line 163: This rule is not right, as it will delete all interfaces. But I didn't figure out a way of expressing that with the query language? Maybe we need two rules here for the case we have just one interface vs multiple?

tests/tests.py
```diff
@@ -137,6 +135,68 @@ def test_insert_field_add_import():
...
+            Filter(
+                enclosing_node= "(class_declaration ) @c_cd",
+                contains = ["""(
+                    (super_interfaces (type_list (type_identifier) @it)
+                    (#eq? @it "FooBar"))
+                )""",],
+            )
+        ]),
+    )
+
```
danieltrt, line 160: This rule is not right because it will delete all interfaces. But I didn't figure out a way of expressing that with the query language? Maybe we need two rules here for the case we have just one interface vs multiple?

polyglot_piranha.pyi
```diff
@@ -164,12 +164,20 @@ class Filter:
...
+    at_most: int
+    "The maximum number of times the contains query should match the enclosing node"
     def __init__(
         self,
         enclosing_node: str,
-        not_contains: list[str] = []
+        not_contains: list[str] = [],
+        contains: list[str] = [],
+        at_least: int = 1,
+        at_most: int = 1000
```
danieltrt, line 180: Should I just make this rust u32::MAX? I'm kind of averse to passing such large numbers as input.

ketkarameya: yes

polyglot_piranha.pyi
```diff
@@ -164,12 +164,20 @@ class Filter:
     enclosing_node: TSQuery
     "AST patterns that some ancestor node of the primary match should comply"
     not_contains: list[TSQuery]
-    "AST patterns that should not match any subtree of node matching `enclosing_node` pattern"
-
+    "AST patterns that SHOULD NOT match any subtree of node matching `enclosing_node` pattern"
+    contains: list[TSQuery]
+    "AST patterns that SHOULD match at least one subtree of node matching `enclosing_node` pattern"
```
ketkarameya, line 169: In this comment lets make it clear that the filter will return true based on the values of the fields `at_least` and `at_most`

src/models/filter.rs
```diff
@@ -45,12 +45,35 @@ pub struct Filter {
...
   not_contains: Vec<TSQuery>,
+  /// AST patterns that should match any subtree of node matching `enclosing_node` pattern
+  #[builder(default = "default_queries()")]
+  #[get = "pub"]
+  #[serde(default)]
+  #[pyo3(get)]
+  contains: Vec<TSQuery>,
+
+  /// Least number of matches we should find for the contains query
+  #[builder(default = "1")]
```
ketkarameya, line 56: extract this "magic number" to the module where all defaults are defined.

src/models/filter.rs
```diff
@@ -45,12 +45,35 @@ pub struct Filter {
   #[serde(default)]
   #[pyo3(get)]
   not_contains: Vec<TSQuery>,
+  /// AST patterns that should match any subtree of node matching `enclosing_node` pattern
+  #[builder(default = "default_queries()")]
+  #[get = "pub"]
+  #[serde(default)]
```
ketkarameya, line 51: missing default value?

src/models/filter.rs
```diff
@@ -45,12 +45,35 @@ pub struct Filter {
...
+  #[pyo3(get)]
+  contains: Vec<TSQuery>,
+
+  /// Least number of matches we should find for the contains query
+  #[builder(default = "1")]
+  #[get = "pub"]
+  #[serde(default)]
```
ketkarameya, line 58: missing default value

src/models/filter.rs
```diff
@@ -45,12 +45,35 @@ pub struct Filter {
...
+  #[pyo3(get)]
+  at_least: u32,
+
+  /// Most number of matches we should find for the contains query
+  #[builder(default = "u32::MAX")]
+  #[get = "pub"]
+  #[serde(default)]
```
ketkarameya, line 65: same as above

src/models/filter.rs
```diff
@@ -45,12 +45,35 @@ pub struct Filter {
   #[serde(default)]
   #[pyo3(get)]
   not_contains: Vec<TSQuery>,
+  /// AST patterns that should match any subtree of node matching `enclosing_node` pattern
+  #[builder(default = "default_queries()")]
+  #[get = "pub"]
+  #[serde(default)]
+  #[pyo3(get)]
+  contains: Vec<TSQuery>,
```
ketkarameya, line 53: Hmmmm. I guess lets make it unary . It is unclear which query is the `atleast` and `atmost` applicable.

src/models/filter.rs
```diff
@@ -60,6 +83,15 @@ impl Filter {
           .map(|x| TSQuery::new(x.to_string()))
           .collect_vec(),
       )
+      .contains(
+        contains
```
ketkarameya, line 87: make unary

tests/tests.py
ketkarameya: hmmm. I think this is a slight mis-use of the contains. I would rather just match `FooBar` it in the `enclosed_node`. I kinda see what you are trying to do. This is actually a limitation currently, its a bit awkward to delete a child node from a "sequence". If at all. were to do this I would - `(_)* (type_identifier) @it (_)*`. 

How about u update this example to delete unused private method or field or local var  ?

src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -123,8 +124,132 @@ fn test_apply_edit_negative() {
...
+
+#[test]
+fn test_satisfies_filters_at_most_negative() {
+  run_test_satisfies_filters(
+    filter! {
+        enclosing_node= "(method_declaration) @md",
+        contains= ["(
+                    ((method_invocation) @method)
+                )",],
+        at_most = 1
```
ketkarameya, line 244: can we add another test which asserts that `contains . atleast 0` == `not-contains`

src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -123,8 +124,132 @@ fn test_apply_edit_negative() {
   );
 }
 
+/// Tests for contains, at_least, and at_most
+
+fn run_test_satisfies_filters(
+  filter: Filter, // Replace with the actual type of the filter
```
ketkarameya, line 130:
```suggestion
  filter: Filter, // Replace with the filter to test 
```

src/models/filter.rs
```diff
@@ -166,6 +208,16 @@ impl SourceCodeUnit {
             return false;
           }
         }
+        for query_with_holes in filter.contains() {
```
ketkarameya, line 211: I am unable to comment a the appropriate location (because of limitation of Github UI I guess). ANyhoo, the documentation for the enclosing fn `_check` should be updated.

ketkarameya, line 211:
```suggestion
               // <put a comment> 
                for query_with_holes in filter.contains() {
```

do something similar for `not-contains` loop too

src/models/filter.rs
```diff
@@ -89,10 +121,13 @@ impl Filter {
 /// ```
 ///
 macro_rules! filter {
```
ketkarameya, line 123: macro documentation needs to be updated

ketkarameya (changes requested):

ketkarameya (changes requested):
The "logic" of this PR makes sense to me. Suggested some changes inline

ketkarameya (changes requested):
The "logic" of this PR makes sense to me. Suggested some changes inline

danieltrt:
Updated w/ the comments

tests/tests.py
```diff
@@ -174,4 +215,4 @@ def _is_readable(input_str: str) -> bool:
     Returns:
         bool: is human readable
     """
-    return not any(re.findall(r"\<(.*) object at (.*)\>", input_str))
+    return not any(re.findall(r"\<(.*) object at (.*)\>", input_str))
```
ketkarameya, line 218:
```suggestion
    return not any(re.findall(r"\<(.*) object at (.*)\>", input_str))
    
```

ketkarameya (approved):
Nits. Otherwise LGTM.

PR #480, opened 2023-05-19
Rename `name` to `extension` #365

### LINKED ISSUE #365: Rename `PiranhaLanguage.name` to `extension`
Probably a nit for a separate PR, but... are using the field `name` to refer to the string representation of the language (i.e. `"kotlin"`) or the extension (i.e. `"kt")? Seems like a weird choice for "name" to be the second...

_Originally posted by @lazaroclapp in https://github.com/uber/piranha/pull/363#discussion_r1103298558_

Field to be renamed - https://github.com/uber/piranha/blob/master/src/models/language.rs#L32 (rename to `extension`)

-----
This is also a good issue for someone new to rust and wants to familiarize themselves with rust tooling. At a higher level - setup vscode for with rust-analyzer,  trigger rename refactoring and regex-replace.
Check list: 
- [ ] update the field name, getter / setter 
- [ ] Update the code comments, variable names and method names with the pattern

How to check if all is well (locally)?
```
cd polyglot/piranha
cargo build
cargo test
```
For mac users - `cargo test --no-default-features`

danieltrt:
Fixes #365

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,7 +202,7 @@ impl Piranha {
     let temp_dir_path = temp_dir.path();
     let sample_file = temp_dir_path.join(format!(
       "sample.{}",
-      self.piranha_arguments.language().name()
+      self.piranha_arguments.language().extension()
     ));
     let mut file = File::create(sample_file).unwrap();
     file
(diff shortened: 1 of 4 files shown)


ketkarameya (approved):
Thanks. LGTM.

PR #482, opened 2023-05-26
Filter without enclosing node

danieltrt:


src/models/filter.rs
```diff
@@ -196,19 +197,30 @@ impl SourceCodeUnit {
...
-    }
 
     // No enclosing node is provided
     if filter.enclosing_node().get_query().as_str() == DEFAULT_ENCLOSING_QUERY {
       // Get the enclosing node matching the pattern specified in the filter (`filter.enclosing_node`)
-      panic!["Not implemented"]
+      self._check_current_node(filter, rule_store, initial_node)
     } else {
+      if node.child_count() > 0 {
+        initial_node = node.child(0).unwrap();
```
danieltrt, line 207: not sure if this should be here

ketkarameya (commented):
some comments inline.

ketkarameya (commented):
some comments inline.

src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -405,3 +405,89 @@ fn test_satisfies_filters_not_contains_negative() {
...
+    &HashMap::from([("method_name".to_string(), "foobar".to_string()),]),
+    &mut rule_store,
+  )));
+}
+
+#[test]
+fn test_not_contains_no_enclosing_negative() {
+  run_test_satisfies_filters_without_enclosing(
+    filter! {
+    enclosing_node = "",
```
ketkarameya, line 469:
```suggestion
```

src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -405,3 +405,89 @@ fn test_satisfies_filters_not_contains_negative() {
...
+    |result| !result,
+  );
+}
+
+// Tests for contains with enclosing
+#[test]
+fn test_contains_no_enclosing_positive() {
+  run_test_satisfies_filters_without_enclosing(
+    filter! {
+    enclosing_node = "",
```
ketkarameya, line 484:
```suggestion
```

src/models/filter.rs
```diff
@@ -196,19 +196,32 @@ impl SourceCodeUnit {
     // This ensures that the below while loop considers the current node too when checking for filters.
     // It does not make sense to check for filter if current node is a "leaf" node.
     let mut initial_node = node;
-    if node.child_count() > 0 {
-      initial_node = node.child(0).unwrap();
-    }
 
     // No enclosing node is provided
     if filter.enclosing_node().get_query().as_str() == DEFAULT_ENCLOSING_QUERY {
```
ketkarameya, line 201: should we make `enclosing_node: Optional[str]` ? Also let's update the `.pyi` file too then

ketkarameya (changes requested):
left comments inline

src/models/filter.rs
danieltrt: I think it;s already optional in the pyi. Must have done this in the previous PR

src/models/unit_tests/source_code_unit_test.rs
```diff
@@ -405,3 +405,36 @@ fn test_satisfies_filters_not_contains_negative() {
...
+    |result| !result,
+  );
+}
+
+// Tests for contains with enclosing
+#[test]
+fn test_not_contains_no_enclosing() {
+  run_test_satisfies_filters(
+    filter! {
+        enclosing_node= "",
```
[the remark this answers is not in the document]
danieltrt: I changed the macro to support this

src/models/filter.rs
ketkarameya: i see. 
Let's update this `#pynew` too 
https://github.com/uber/piranha/blob/master/src/models/filter.rs#L83 to use `default_enclosing_node`

And in pyi instead of `enclosing_node: str = ''` lets make it `enclosing_node: Optional[str] = None` Some sense I want to keep the signature of `#pynew` and `pyi` consistent (As much as we can)

(I know its logically still the same, but just lets reflect our intention in code)

ketkarameya (approved):
small nit. Else LGTM.

Issue #390, opened 2023-02-23
Unintuitive interface - To match a code snippet, we need to set `dry_run`


ketkarameya:
IF a user provides a match rule and code snippet, specifying `dry_run` explicitly seems rather unintuitive.
The bug is that, if we don't provide a `dry_run`, the persist call throws an exception. 

 
```
source = """package com.uber.piranha;
class Piranha {
    public void run() {
        if (isToggleDisabled("dark_mode")) {
            // disable dark mode
        }
    }
}
"""
rule = Rule(
    name= "find_toggle_disabled_dark_mode",
    query= """(
        (method_invocation 
            name: (identifier) @method_name 
            arguments: (argument_list (string_literal) @argument))
        (#eq? @method_name "isToggleDisabled")
        (#match? @argument "dark_mode")
    )""",
)

args = PiranhaArguments(
    language="java",
    code_snippet=source,
    rule_graph=RuleGraph(rules=[rule], edges= []),
    substitutions={},
    dry_run= True # THIS IS THE PROBLEM. THIS IS UNITUITVE
)

output = execute_piranha(args)
print(output)
```

ketkarameya:
Actually it seems that if code_Snippet is passed, dry_run has to be set to true

PR #489, opened 2023-06-02
Fix bug related to code snippet mode and dry run

ketkarameya:
Fixes bug #390. Code snippet requires us to pass dry_run as true, else its panics. This is unintuitive. This PR fixes that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,16 +1,15 @@
 /*
-Copyright (c) 2023 Uber Technologies, Inc.
+  Copyright (c) 2023 Uber Technologies, Inc.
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1,16 +1,15 @@
 /*
-Copyright (c) 2023 Uber Technologies, Inc.
+  Copyright (c) 2023 Uber Technologies, Inc.
```
ketkarameya, line 2: for some reason our license linter wanted me to add these white spaces. :| anyhoo

danieltrt (approved):
LGTM

PR #492, opened 2023-06-04
 Dataflow analysis to check query consistency 

danieltrt:
# Dataflow Analysis for Piranha

This PR introduces a dataflow analysis for Piranha. The goal is to ensure that all variables used in query predicates are defined before usage

I think it's best to explain this with an example. Consider the rule graph below, where each rule defines a set of tags via capture groups in it's query:

- `seed rule` with input the substitution `@stale_flag`, defines the capture group `@method_decl`
- `rule1` defines the capture group `@var_id` and `@class_n`
- `rule2` defines the capture group`@var_id`
- `final rule` with no capture groups.

After running the analysis, we would have the following set of variables available to use for each predicate of each rule:

```
                  +--------------+
                  |              |
                  |  seed rule   |  @method_decl
                  |              |  @stale_flag
                  +--------------+
                      /       \
                     /         \
                    /           \
                   /             \
                  /               \
         +--------------+   +--------------+
         |              |   |              |
         |    rule1     |   |    rule2     |
         |              |   |              |
         +--------------+   +--------------+
@method_decl     \                /     @method_decl
@var_id           \              /      @var_id
@class_n           \            /       @stale_flag
@stale_flag         \          /
                     \        /
                   +--------------+
                   |              |
                   |  final rule  |
                   |              |
                   +--------------+

                     @method_decl
                     @var_id
                     @stale_flag
```

This actual implementation of the analysis is done using Kildall's algorithm. The analysis is performed over the following lattice:

### Lattice 
The lattice is defined over the powerset of all tags defined in the rule graph. The partial order operator is $\supseteq$ (e.g., $\sigma_1$ is smaller than $\sigma_2$, iff $\sigma_1$ contains $\sigma_2$. Formally:

$\sigma \in \mathcal{P}(DEFS)$ 
$\sigma_1 \sqsubseteq \sigma_2 \iff \sigma_1 \supseteq \sigma_2$ 
$\sigma_1 \sqcup \sigma_2 \iff \sigma_1 \cap \sigma_2$ 
$\top = \varnothing$ 
$\bot = DEFS$

### Flow Function
The flow function specifies which tags are available to use in predicates after running the matcher/query. In this case, the tags available after executing a rule are the ones **previously defined** and the ones **defined within the rule itself.**

$f_{rule}(\sigma_{in}) = \sigma_{in} \cup \text{definitions}(rule)$

### Boundary Condition 
The boundary condition defines how we initiate the analysis for entry points. In this case, the seed rule is given the substitutions as user-input.

$\sigma_{in}[\text{seed rule}] = \text{substitutions}$

### Initial Value for Interior Points 
For each non-seed rule, we initialize it's input to bottom (the universal set). 

$\sigma_{in}[rule] = \bot$


src/df/README.md
```diff
@@ -0,0 +1,12 @@
+# Dataflow analysis for Piranha
+
+This folder implements a small data flow analysis framework `df.rs`.
+
+There's an to a "Definite Assignment Analysis" under `tag_analysis.rs`.
```
ketkarameya, line 5: typo?

src/df/df.rs
```diff
@@ -0,0 +1,103 @@
...
+impl<D: Direction> DataflowAnalysis<D> {
+  pub fn new(direction: D) -> Self {
+    DataflowAnalysis {
+      direction,
+      sigma_in: DfResults::new(),
+      sigma_out: DfResults::new(),
+    }
+  }
+
+  // From https://cmu-program-analysis.github.io/2023/index.html
```
ketkarameya, line 70: Lets document what this function is doing? (Also  add some inline comments here in this function)

src/df/tag_analysis.rs
```diff
@@ -0,0 +1,141 @@
...
+// in compilers. Instead of tracking variable definitions, it tracks "tags" as they propagate
+// through the graph. The goal of the analysis is to find, for each point in the graph,
+// the set of tags that will always reach that point. The result can then be used to check if the
+// query contains any variable tag that was not defined.
+
+#[derive(Debug, Clone, Getters)]
+pub struct DefiniteAssignmentSigma {
+  #[get = "pub"]
+  variables: HashSet<String>,
+  is_bottom: bool, // hack to prevent initializing variables with all elements for bottom
```
ketkarameya, line 34: not sure what u mean by this comment

src/df/df.rs
```diff
@@ -0,0 +1,103 @@
...
+ express or implied. See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+use getset::Getters;
+use std::collections::HashMap;
+use std::hash::Hash;
+
+/// Trait for states (sigma), which store the states of the analysis at each program point
+pub trait Sigma: Clone {
```
ketkarameya, line 19: maybe this trait could extend `Eq` and `PartialEq` ? then u will get the default equals for free, and will not have to create and override `is_equal`


To implement the `Eq` you can - https://doc.rust-lang.org/std/cmp/trait.Eq.html at the appropriate struct

src/df/tag_analysis.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+}
+
+impl Direction for ForwardDefiniteAssignment {
+  type Node = Rule;
+  type Sigma = DefiniteAssignmentSigma;
+
+  fn successors(&self, _rule: &Rule) -> Vec<Rule> {
+    // Get the rules and discard the edge type from the tuple
```
ketkarameya, line 89: lets try to use set intersection ?

src/df/tag_analysis.rs
```diff
@@ -0,0 +1,141 @@
...
+      .rules()
+      .iter()
+      .filter(|r| child_rules.contains(r.name()))
+      .cloned()
+      .collect()
+  }
+
+  // Since the join operator or this analysis is intersection, the initial value
+  // all intermediate points need to be set to the universal set (bottom)
+  fn initial_value(&self) -> DefiniteAssignmentSigma {
```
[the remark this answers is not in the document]
ketkarameya: We could keep this `initial_value` method instead call `::default()` inside ti

src/models/unit_tests/rule_graph_validation_test.rs
```diff
@@ -118,3 +119,20 @@ fn test_filter_bad_arg_contains_n_sibling() {
     .sibling_count(2)
     .build();
 }
+
+#[test]
+fn test_df_warnings() {
```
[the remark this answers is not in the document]
ketkarameya: or `should_panic` ?

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,225 @@
...
+    piranha_rule! {
+      name = "detect_method_bar_in_foo",
+      query = "(
+      (method_declaration
+          name: (identifier) @method_name
+          . (class_declaration
+                name: (_) @detected_class_name
+             )
+      )
+      (#eq? @detected_class_name @class_name)
```
ketkarameya, line 42: should'nt `class_name` be declared as a hole in this rule?

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,225 @@
...
+  analysis.run_analysis(rules_post_order, vec![entry_rule.clone()]);
+
+  // Check the sigma in of the 2nd rule
+  let sigma = analysis
+    .sigma_in()
+    .get(&rules[2])
+    .unwrap()
+    .variables
+    .clone();
+  let expected = vec!["@method_name", "@class_name", "@detected_class_name"]
```
ketkarameya, line 96: why not directly construct a hashset with `"@method_name".to_string() ...`

Also why do we have the `@` here?

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,225 @@
...
+
+  // Get the rules in post order for optimal performance of the analysis
+  let mut rules_post_order = rules.clone();
+  rules_post_order.reverse();
+  // The entry point of the rule graph
+  let entry_rule = &rules[0];
+  analysis.run_analysis(rules_post_order, vec![entry_rule.clone()]);
+
+  // Check the sigma in of the 2nd rule
+  let sigma = analysis
```
ketkarameya, line 90: lets write an assert for all three rules. 

Also we can abstract this small logic out into a small inner function if we want.

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,225 @@
...
+    variables: HashSet::new(),
+    is_bottom: false,
+  };
+
+  let new_sigma = forward.flow(&rule, &sigma);
+  assert_eq!(new_sigma.variables.len(), 4);
+}
+
+#[test]
+fn test_flow_function_0() {
```
ketkarameya, line 197: is this a -ve test case? 
can we add more -ve test cases with wrong flow n all ?

ketkarameya (changes requested):

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,239 @@
...
+      (call_expression
+          function: (identifier) @function_name
+          . (method_declaration
+                name: (_) @detected_method_name
+             )
+      )
+      (#eq? @detected_method_name @method_name)
+      (#eq? @function_name \"baz\")
+    )",
+      holes = ["@method_name"]
```
ketkarameya, line 68: holes do not need `@`

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,239 @@
...
+      (method_declaration
+          name: (identifier) @method_name
+          . (class_declaration
+                name: (_) @detected_class_name
+             )
+      )
+      (#eq? @detected_class_name @class_name)
+      (#eq? @method_name \"bar\")
+    )",
+      holes = ["@class_name"]
```
ketkarameya, line 54: same.  check other places too in case I mess any

src/models/rule_graph.rs
```diff
@@ -193,6 +199,48 @@ impl RuleGraph {
...
+
+    // Now validate for each rule, that all the tags used in the predicate are defined in the sigma out
+    // Sigma out represents all the capture groups available to use in the rule's predicates
+    for rule in self.rules() {
+      let defined_variables = analysis.sigma_out().get(rule).unwrap().variables();
+      let tags_in_predicates = get_tags_usage_from_matcher(rule);
+      // If there's any tag in the predicate that is not sigma out, then we might be using an undefined variable
+      for tag in tags_in_predicates {
+        if !defined_variables.contains(&tag) {
+          println!(
```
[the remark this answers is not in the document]
ketkarameya: Lets convert this to log.warn

src/df/unit_tests/tag_analysis_test.rs
```diff
@@ -0,0 +1,239 @@
...
+  let entry_rule = &rules[0];
+  analysis.run_analysis(rules_post_order, vec![entry_rule.clone()]);
+
+  // Call the inner function to check the sigma_in of each rule
+  check_sigma_in(&analysis, &rules[0], vec![]);
+  check_sigma_in(
+    &analysis,
+    &rules[1],
+    vec![
+      "@class_name",
```
ketkarameya, line 165: why are we requiring `@` for the tag names?

ketkarameya (commented):
some comments inline. Otherwise looks good to me. 

I guess u shud update rust by re-installing
and run `pre-commit run --all-files`, and the changes to the this PR

PR #493, opened 2023-06-04
Validate Rule graph

ketkarameya:
This PR introduces very basic validation for our RuleGraphs. At this moment we make sure that (i) All tree-sitter queries are syntactically correct (ii) Filter fields are configured correctly (this is was something we previously checked)

This PR also ensures that the the graph is validated irrespective of where it is loaded from - `toml`, Rust API, or Python API (Have added test for each API)

### DIFF
--- a/src/cleanup_rules/kt/rules.toml
+++ b/src/cleanup_rules/kt/rules.toml
@@ -197,7 +197,7 @@ query = """
 (
 (prefix_expression (boolean_literal) @exp) @prefix_expression
 (#eq? @exp "false")
-(#match @prefix_expression "!.*")  
+(#match? @prefix_expression "!.*")  
 )
 """
 replace = "true"
@@ -216,7 +216,7 @@ query = """
 (
 (prefix_expression (boolean_literal) @exp) @prefix_expression
 (#eq? @exp "true")
-(#match @prefix_expression "!.*")  
+(#match? @prefix_expression "!.*")  
 )
 """
 replace = "false"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,6 @@ use std::{collections::HashMap, fs::File, io::Write, path::PathBuf};
 
 use itertools::Itertools;
 use log::{debug, info};
-use tree_sitter::Parser;
 
 use crate::models::rule_store::RuleStore;
 
@@ -111,11 +110,9 @@ impl Piranha {
   /// Performs cleanup related to stale flags
   fn perform_cleanup(&mut self) {
     // Setup the parser for the specific language
-    let mut parser = Parser::new();
     let piranha_args = &self.piranha_arguments;
-    parser
-      .set_language(*piranha_args.language().language())
-      .expect("Could not set the language for the parser.");
+
+    let mut parser = piranha_args.language().parser();
 
     let mut path_to_codebase = self.piranha_arguments.path_to_codebase().to_string();
 
--- a/src/models/filter.rs
+++ b/src/models/filter.rs
@@ -26,7 +26,9 @@ use crate::utilities::{
   tree_sitter_utilities::{get_all_matches_for_query, get_match_for_query, get_node_for_range},
 };
 
-use super::{rule::InstantiatedRule, rule_store::RuleStore, source_code_unit::SourceCodeUnit};
+use super::{
+  rule::InstantiatedRule, rule_store::RuleStore, source_code_unit::SourceCodeUnit, Validator,
+};
 
 use crate::utilities::{tree_sitter_utilities::TSQuery, Instantiate};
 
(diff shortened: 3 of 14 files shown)


src/cleanup_rules/kt/rules.toml
```diff
@@ -197,7 +197,7 @@ query = """
 (
 (prefix_expression (boolean_literal) @exp) @prefix_expression
 (#eq? @exp "false")
-(#match @prefix_expression "!.*")  
+(#match? @prefix_expression "!.*")
```
ketkarameya, line 200: Found this bug (and the one below) :)

src/lib.rs
```diff
@@ -111,11 +110,9 @@ impl Piranha {
   /// Performs cleanup related to stale flags
   fn perform_cleanup(&mut self) {
     // Setup the parser for the specific language
-    let mut parser = Parser::new();
     let piranha_args = &self.piranha_arguments;
-    parser
```
ketkarameya, line 116: un related to this PR :| , but I observed that i was not using the `piranha_args.language()` to get the parser

src/models/filter.rs
```diff
@@ -103,47 +105,68 @@ impl Filter {
...
+    if *self.not_enclosing_node() != default_not_enclosing_node() {
+      self.not_enclosing_node().validate()?
+    }
+
+    if *self.contains() != default_contains_query() {
+      self.contains().validate()?
+    }
+
+    if *self.not_contains() != default_not_contains_queries() {
+      self.not_contains().iter().try_for_each(|x| x.validate())?
```
ketkarameya, line 150: Such ergonomic APIs make me like Rust  :)  ([`try_for_each`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each))

ketkarameya (commented):
self reviewed

ketkarameya (commented):
self reviewed

src/models/mod.rs
```diff
@@ -24,3 +24,7 @@ pub(crate) mod rule_graph;
 pub(crate) mod rule_store;
 pub(crate) mod scopes;
 pub(crate) mod source_code_unit;
+
+pub(crate) trait Validator {
+  fn validate(&self) -> Result<(), String>;
```
ketkarameya, line 29: as our validation enriches, I will modify this Type parameter

src/utilities/tree_sitter_utilities.rs
```diff
@@ -303,6 +304,21 @@ pub(crate) fn get_replace_range(input_edit: InputEdit) -> Range {
   }
 }
 
+pub(crate) fn get_ts_query_parser() -> Parser {
+  let mut parser = Parser::new();
+  parser
+    .set_language(tree_sitter_query::language())
+    .expect("Could not set the language for the parser.");
```
lazaroclapp, line 311: I don't understand what's going on here. Are we creating a new parser and expecting it to be set to an invalid language? Why? (This at the very least needs documentation)

src/utilities/tree_sitter_utilities.rs
```diff
@@ -317,6 +333,17 @@ impl TSQuery {
...
 }
 
+impl Validator for TSQuery {
+  fn validate(&self) -> Result<(), String> {
+    let mut parser = get_ts_query_parser();
+    parser
+      .parse(self.get_query(), None)
+      .filter(|x| number_of_errors(&x.root_node()) == 0)
+      .map(|_| Ok(()))
+      .unwrap_or(Err(format!("Cannot parse - {}", self.get_query())))
```
lazaroclapp, line 343: Do we report what exactly we failed to parse? i.e. for the missing parenthesis "Unexpected end of query string, unmatched `(` at position x:y" or similar?

src/models/rule_graph.rs
```diff
@@ -54,6 +55,15 @@ pub struct RuleGraph {
   graph: HashMap<String, Vec<(String, String)>>,
 }
 
+impl Validator for RuleGraph {
+  fn validate(&self) -> Result<(), String> {
+    match self.rules().iter().try_for_each(|rule| rule.validate()) {
+      Ok(()) => Ok(()),
+      Err(e) => Err(format!("Incorrect Rule Graph - {}", e)),
```
lazaroclapp, line 62: Question: Do we report the first failing rule or all syntax errors in every rule in the rule graph? (Neither would be right or wrong, specially at this point, just want to make sure. I'd think it's the first case looking at this, but nor sure what the semantics of `try_for_each` are.

src/utilities/tree_sitter_utilities.rs
ketkarameya: aah maybe its just bad api naming in rustlang. 
We are setting a language, and if the setter throws an exception we panic with "_Could not set the language for the parser._"

.expect() ~= .orElseThrow() in Java

src/utilities/tree_sitter_utilities.rs
ketkarameya: At this moment we do not report why the query is incorrect (or any diagnostics around it) 
I will add diagnostics for this as a follow up. I think it just reporting error does not make sense, without diagnostics. Tracking this issue #494

src/models/rule_graph.rs
ketkarameya: We are short-circuiting at the first exception. 
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each

ketkarameya (commented):
Replies inline

ketkarameya (commented):
Replies inline

ketkarameya (commented):
Replies inline

ketkarameya (commented):
Replies inline

ketkarameya (commented):
Replies inline

src/utilities/tree_sitter_utilities.rs
lazaroclapp: Got it. Also, a bit surprised that `tree_sitter_query` doesn't need any import. For a second I thought this was of the same type `piranha_args.language()`, i.e. `PiranhaLanguage`, and was wondering why this wasn't simply `tree_sitter_query::language().parser()`, but I see the difference now. I still think this all merits a short comment here explaining what this code does (uses tree sitter to construct a parser for its own DSL without all the extensions we have made to Piranha-supported language, correct?)

lazaroclapp (approved):
See comment above about documenting the `tree_sitter_query::language()` bit, but otherwise this LGTM.

src/utilities/tree_sitter_utilities.rs
ketkarameya: aah i see the the source of confusion.  
Yes u r correct, we are using ts to parse its own dsl.
I will address this as a comment in the code

PR #497, opened 2023-06-06
Bug Fix - Create dummy rule from py api

ketkarameya:
Currently, because of slight misconfuguration Piranha python API does not allow the user to create a dummy rule. This PR fixes this bug.


polyglot_piranha.pyi
```diff
@@ -210,11 +210,13 @@ class Rule:
     "Holes that need to be filled, in order to instantiate a rule"
     filters: set[Filter]
     "Filters to test before applying a rule"
+    is_seed_rule: bool
```
ketkarameya, line 213: not related to bug fix, Just saw that this field was missing here so added it

PR #499, opened 2023-06-07
Bug Fix: Cleanup sub-optimal ternary operator cleanup

ketkarameya:
We were not simplifying the code when the alternative and consequence of the ternary opertors are the same. This PR fixes that, and adds tests for it

### DIFF
--- a/test-resources/swift/cleanup_rules/input/SampleClass.swift
+++ b/test-resources/swift/cleanup_rules/input/SampleClass.swift
@@ -492,5 +492,7 @@ class SampleClass {
          } else {
             toBeDeletedInequality12b()
          }
+
+         x = condition() ? TestEnum.stale_flag_one.isEnabled : true
     }
 }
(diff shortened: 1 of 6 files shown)


test-resources/swift/cleanup_rules/input/SampleClass.swift
```diff
@@ -492,5 +492,7 @@ class SampleClass {
          } else {
             toBeDeletedInequality12b()
          }
+
+         x = condition() ? TestEnum.stale_flag_one.isEnabled : true
```
lazaroclapp, line 496: I think this is probably fine in practice, but an interesting caveat is that if `condition()` has side-effects, this actually does change program behavior beyond setting the flag value to a constant, since it skips the execution of `condition()`.

lazaroclapp (approved):
LGTM, but note the potential semantics change when the condition is a method call with potential side-effects!

test-resources/swift/cleanup_rules/input/SampleClass.swift
ketkarameya: yes, i see what u mean.
But in my experience folks want such aggressive simplifications. They do not want us to be conservative.

PR #500, opened 2023-06-07
Scripts for rule inference using LLM

danieltrt:
This PR includes a basic python script to infer simple piranha rules from code examples. 
I had to close the previous PR because it wasn't syncing with the remote branch


danieltrt:
Why is this not syncing with the branch :|

PR #505, opened 2023-06-08
Inference with multiple rounds in case the first attempt files

danieltrt:
This PR makes the inference "interactive". After asking the model to infer a rule, we test it. If the rule transforms the source into the target, we return it to the user. Otherwise, we provide execution feedback to the model and ask it to try again.

### DIFF
--- a/experimental/rule_inference/logger_formatter.py
+++ b/experimental/rule_inference/logger_formatter.py
@@ -2,8 +2,9 @@ import logging
 
 
 class CustomFormatter(logging.Formatter):
-    blue = "\033[36m"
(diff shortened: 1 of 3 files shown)


experimental/rule_inference/logger_formatter.py
```diff
@@ -2,8 +2,9 @@
 
 
 class CustomFormatter(logging.Formatter):
-    blue = "\033[36m"
+    blue = "\033[34m"
```
danieltrt, line 5: This was wrong

ketkarameya (approved):

PR #506, opened 2023-06-08
Add UI playground to rule inference

danieltrt:
This PR introduces a basic UI for the rule inference module


danieltrt:
Nvm I found a mirror hosted on cloudfare. goign to replace it

PR #509, opened 2023-06-11
Add support to delete a specific index child

ketkarameya:
Adds a basic feature to delete specific child of a specific node.

In addition to `replace_node` not users can define `child_idx` to further narrow down upon the exact sub-tree they want to replace. 

As a followup, we will 
1. allow passing this index via piranha arguments 
2. use index of nodes from previously matched rules


src/cleanup_rules/kt/rules.toml
```diff
@@ -488,7 +488,7 @@ name = "delete_file_with_no_declarations"
 query = """
 (source_file) @source_file"""
 replace = ""
-replace_node = "program"
```
ketkarameya, line 491: The changes  logic (below), now disallows wrong rules that reference non-existing variables

test-resources/swift/leading_comma/configurations/rules.toml
```diff
@@ -4,6 +4,6 @@ query = """(
 (switch_pattern (pattern (simple_identifier) @switch_case))@switch_pattern
 (#eq? @switch_case "@stale_flag")
 )"""
-replace_node = "switch_patttern"
```
ketkarameya, line 7: The changes logic (below), now disallows wrong rules that reference non-existing variables

test-resources/py/delete_cleanup_str_in_list/configurations/rules.toml
```diff
@@ -115,6 +115,6 @@ query = """
 (#match? @list "\\\\[\\\\s*,?\\\\]")
 )
 """
-replace_node = "@assignment"
```
ketkarameya, line 118: The changes logic (below), now disallows wrong rules that reference non-existing variables

test-resources/kt/feature_flag_system_2/treated/configurations/rules.toml
```diff
@@ -128,7 +128,7 @@ query = """
                                                                        (simple_identifier) @arg_method_receiver
                                                                     (navigation_suffix (_)@arg_method_name)
                                                                       )
-                                                              )))))@ce
```
ketkarameya, line 131: The changes logic (below), now disallows wrong rules that reference non-existing variables

test-resources/kt/feature_flag_system_2/control/configurations/rules.toml
```diff
@@ -128,7 +128,7 @@ query = """
                                                                        (simple_identifier) @arg_method_receiver
                                                                     (navigation_suffix (_)@arg_method_name)
                                                                       )
-                                                              )))))@ce
```
ketkarameya, line 131: The changes logic (below), now disallows wrong rules that reference non-existing variables

src/models/rule.rs
```diff
@@ -51,6 +51,12 @@ pub struct Rule {
   #[pyo3(get)]
   query: TSQuery,
   /// The tag corresponding to the node to be replaced
```
lazaroclapp, line 53: This doc is incorrect (duplicated from field `replace_node` below). Please change to a proper doc for this new `replace_idx` field :)

src/utilities/tree_sitter_utilities.rs
```diff
@@ -79,17 +80,23 @@ pub(crate) fn get_all_matches_for_query(
...
-      let mut replace_node_range = captured_node_range;
-
+      let mut final_replace_node_range = captured_node_range;
       if let Some(replace_node_name) = &replace_node {
-        if let Some(r) = get_range_for_replace_node(query, &query_matches, replace_node_name) {
-          replace_node_range = r;
+        if let Some(r) =
+          get_range_for_replace_node(query, &query_matches, replace_node_name, replace_node_idx)
+        {
+          final_replace_node_range = r;
```
lazaroclapp, line 88: Not sure why this is `final_replace_node_range`, if we are pushing it on every iteration through the loop, rather than just the last one... what am I missing here?

lazaroclapp (commented):
Pretty nice how this rudimentary validation already found a bunch of errors in our test resources (I think the new feature and the validation could have been separated into two PRs, but this isn't super long either way, so...)

Minor nits, below, but otherwise LGTM.

src/utilities/tree_sitter_utilities.rs
ketkarameya: aah its just bad naming practice on my end :| it should be just `replace_node_range`. I was doing some changes, renamed the var, forgot to un-rename the changes when i undid those changes:|

ketkarameya (commented):
addressed PR comments

lazaroclapp (approved):
LGTM!

PR #523, opened 2023-06-30
Support tree-sitter queries as Piranha Language

ketkarameya:
Adds support for matching and rewriting tree-sitter queries with Piranha. Using scm extention(scheme) because thats the convention. I would expect that in most cases we would use it in code snippet mode


test-resources/scm/simple_rename/configurations/rules.toml
```diff
@@ -0,0 +1,24 @@
...
+# <p>http://www.apache.org/licenses/LICENSE-2.0
+# 
+# <p>Unless required by applicable law or agreed to in writing, software distributed under the
+# License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Matchs an enum entry named `GLOBAL_TAG.enum_entry_name`
+[[rules]]
+name = "delete_enum_entry"
```
danieltrt, line 14: Why call it delete_enum_entry if it's just a rename?

danieltrt (approved):
LGTM

PR #527, opened 2023-07-05
Introduce CompiledCGPatterns capturing the TS-Query and regex (placeholder)

ketkarameya:
Introduces `CompiledCGPattern` which wraps `treesitter::Query` and `regex::Regex`

(A simple temporary test to show that regex support is added but not supported. This test is updated in the stacked diff). 

------
* #536
* #529
* __->__ #527


src/models/capture_group_patterns.rs
```diff
@@ -56,3 +66,50 @@ impl Instantiate for CGPattern {
...
 }
+
+#[derive(Debug)]
+pub(crate) enum CompiledCGPattern {
+  Q(Query),
+  R(Regex), // Regex is not yet supported
+}
+
+impl CompiledCGPattern {
+  /// Applies the query upon the given node, and gets all the matches
```
lazaroclapp, line 77: Is the documentation correct here? It seems this returns 0 or 1 matches. Is this meant to be the docstring for `get_matches`?

lazaroclapp (commented):
Minor documentation nit (also, might be worth having a full documentation comment for both `get_match` and `get_matches`)

lazaroclapp (approved):
👍

PR #529, opened 2023-07-05
Add regex stuff

ketkarameya:
Implement the logic for matching code with regex. 

Add tests for : 
* query as regex 
* filter as regex 
* propagate when hole is captured by regex 
* propagate built in rules when query is regex 

-------

* __->__ #529


test-resources/java/regex_based_matcher/configurations/rules.toml
```diff
@@ -0,0 +1,59 @@
...
+[[rules]]
+name = "replace_call_def_fed"
+query = """rgx (?P<n>abc\\(\\)\\.(?P<m_def>def)\\(\\)\\.ghi\\(\\))"""
+replace_node = "m_def"
+replace = "fed"
+
+
+# The below three rules do a dummy type migration from List<Integer> to NewList 
+
+# Updates the import statement from `java.util.List` to `com.uber.NEwList`
```
lazaroclapp, line 35:
```suggestion
# Updates the import statement from `java.util.List` to `com.uber.NewList`
```

test-resources/java/regex_based_matcher/configurations/rules.toml
```diff
@@ -0,0 +1,59 @@
...
+enclosing_node = "(method_declaration) @cmd"
+
+# Updates the relevant callsite from `add` to `addToNewList`
+[[rules]]
+name = "update_add"
+query = """rgx (?P<call>@name\\.(?P<m_name>add)\\(\\w+\\))"""
+replace_node = "m_name"
+replace = "addToNewList"
+holes = ["name"]
+is_seed_rule = false
```
lazaroclapp, line 59: Every `replace_node` in this test suite is a constant, even though the surrounding context is matched by more complex regexps. For generality, we might want a regex that matches a more general pattern on the replaced node. How about converting `OurMapOfX` to `HashMap<X>` for various `X`s (i.e. the code could have `OurMapOfInteger`, `OurMapOfString`, and `OurMapOfOurMapOfLong`)

Or is something like that not meant to be supported?

test-resources/java/regex_based_matcher/input/Sample.java
```diff
@@ -0,0 +1,37 @@
...
+
+    void someTypeChange() {
+        // Will get updated
+        List<Integer> a = getList();
+        Integer item = getItem();
+        a.add(item);
+        
+        // Will not get updated
+        List<String> b = getListStr();
+        Integer item = getItemStr();
```
lazaroclapp, line 33:
```suggestion
        String itemStr = getItemStr();
```

Or something like that. I know this code doesn't get compiled, but it still should make sense types-wise :)

src/models/capture_group_patterns.rs
```diff
@@ -38,12 +39,20 @@ impl CGPattern {
...
+    let mut _val = &self.pattern()[4..];
+    _val.to_string()
+  }
 }
 
 impl Validator for CGPattern {
   fn validate(&self) -> Result<(), String> {
     if self.pattern().starts_with("rgx ") {
-      panic!("Regex not supported")
+      let mut _val = &self.pattern()[4..];
```
ketkarameya, line 52:
```suggestion
      let mut _val = &self.extract_regex();
```

src/utilities/regex_utilities.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Applies the query upon the given `node`, and gets all the matches
+/// # Arguments
+/// * `node` - the root node to apply the query upon
+/// * `source_code` - the corresponding source code string for the node.
+/// * `recursive` - if `true` it matches the query to `self` and `self`'s sub-ASTs, else it matches the `query` only to `self`.
+/// * `replace_node` - node to replace
+///
+/// # Returns
+/// The range of the match in the source code and the corresponding mapping from tags to code snippets.
```
lazaroclapp, line 28: Is this correct? Return seems to just be a vector of matches. Where is this mapping encoded?

src/utilities/regex_utilities.rs
```diff
@@ -0,0 +1,70 @@
...
+pub(crate) fn get_all_matches_for_regex(
+  node: &Node, source_code: String, regex: &Regex, recursive: bool, replace_node: Option<String>,
+) -> Vec<Match> {
+  // let code_snippet = node.utf8_text(source_code.as_bytes()).unwrap();
```
lazaroclapp, line 32: Delete commented out line

src/utilities/regex_utilities.rs
```diff
@@ -0,0 +1,70 @@
...
+        captures.get(0).unwrap()
+      };
+      let matches = extract_captures(&captures, &names);
+      all_matches.push(Match::from_regex(&group_by_tag, matches, &source_code));
+    }
+  }
+  all_matches
+}
+
+// Creates an hashmap from the capture group(name) to the corresponding code snippet.
```
lazaroclapp, line 57:
```suggestion
// Creates a hashmap from the capture group (name) to the corresponding code snippet.
```

src/utilities/regex_utilities.rs
```diff
@@ -0,0 +1,70 @@
...
+  let names = regex.capture_names().collect_vec();
+  let mut all_matches = vec![];
+  for captures in all_captures {
+    // Check if the range of the self (node), and the range of outermost node captured by the query are equal.
+    let range_matches_node = node.start_byte() == captures.get(0).unwrap().start()
+      && node.end_byte() == captures.get(0).unwrap().end();
+    let range_matches_inside_node = node.start_byte() <= captures.get(0).unwrap().start()
+      && node.end_byte() >= captures.get(0).unwrap().end();
+    if (recursive && range_matches_inside_node) || range_matches_node {
+      let group_by_tag = if let Some(ref rn) = replace_node {
```
lazaroclapp, line 43: Group by tag is either the matched code string (plus location info) corresponding to the match group for the replace node if present or a the matched code string matching the full regex match if not, correct? Why is the name `group_by_tag`?

src/models/matches.rs
```diff
@@ -55,6 +55,18 @@ pub(crate) struct Match {
...
 impl Match {
+  pub(crate) fn from_regex(
+    mtch: &regex::Match, matches: HashMap<String, String>, source_code: &str,
+  ) -> Self {
+    Match {
+      matched_string: mtch.as_str().to_string(),
+      range: Range::from_regex_match(mtch, source_code),
+      matches,
+      associated_comma: None,
+      associated_comments: Vec::new(),
```
lazaroclapp, line 66: Do we populate associated comments later and does that work for regex matches?

src/models/capture_group_patterns.rs
lazaroclapp, line 52: Why do we have this logic here and in `extract_regex()` both? This probably should not be duplicated in case we ever change the identifier, and also should be something like `len(REGEX_QUERY_PREFIX)` and that prefix used elsewhere to match `"rgx "`

lazaroclapp (approved):
A bunch of documentation/comment nits, but LGTM once those are addressed!

ketkarameya (commented):
aah forgot to push my comments

ketkarameya:
Addressed comments https://github.com/uber/piranha/pull/529/commits/3ac21b03af456692c94d66ec4fc2eab8312afb08 and rebased

PR #532, opened 2023-07-06
Simplify conjunctions where LHS is false

ketkarameya:
Make the boolean simplification rules less conservative (based on the feedback received)

Earlier we used to simplify `<expr> && false` only if `expr` was `identifier` or `literal` (to account for side effect)
Now we don't do that.


test-resources/go/feature_flag/builtin_rules/boolean_expression_simplify/input/sample.go
```diff
@@ -91,7 +91,7 @@ func simplify_false_and_something(something bool) {
     } else {
         fmt.Println("else 5")
     }
-    // does not simplify binary_expression; left call may contain side-effects
+    // simplify
     if exp.BoolValue("random") && exp.BoolValue("false") {
         fmt.Println("keep 1")
```
lazaroclapp, line 96: Should no longer read "keep", right?

lazaroclapp: Also, "keep as it is" is stale in the `f1() || exp.BoolValue("true")` case below

lazaroclapp (commented):
Some strings are outdated, but otherwise sounds good.

test-resources/go/feature_flag/builtin_rules/boolean_expression_simplify/input/sample.go
ketkarameya: Good catch! Done.

lazaroclapp (approved):

PR #535, opened 2023-07-10
Fixing a small typo in the readme with a repeated parameter

BlakeItOut:
The dry_run parameter and it's associated description are repeated. This removes that duplication.


ketkarameya (approved):
Thanks for fixing this.

PR #537, opened 2023-07-11
Aggressively simplify something or true

ketkarameya:
Aggressively simplifies boolean disjunction expressions when RHS is  and LHS is any AST node. Earlier we were conservatively only simplifying when LHS was an identifier or literal. But now(after feedback from folks) we have decided to make this simplification more aggressive. This PR is similar to https://github.com/uber/piranha/pull/532/.


danieltrt (approved):
LGTM!

PR #540, opened 2023-07-12
Use crate::model::{Range, Point} over tree_sitter::{Range, Point}

ketkarameya:
Use our own Range and Point struct everywhere instead of using tree_sitter's. This will allow us to add more features into this Range struct.


danieltrt (approved):
This PR looks good to me

PR #541, opened 2023-07-14
Improve ternary simplification

ketkarameya:
Adds a hacky way to simplify ternary expressions when nil is used. This hack can be removed when https://github.com/alex-pinkus/tree-sitter-swift/issues/303 is resolved


danieltrt (approved):
LGTM

PR #546, opened 2023-07-17
Implement backend to support for alternation

danieltrt:
:[x].doSomething(:[y]) ->     doSomething(:[x: identifier, method_invocation], :[y]) now supported

### DIFF
--- a/experimental/piranha_playground/rule_inference/static_inference.py
+++ b/experimental/piranha_playground/rule_inference/static_inference.py
@@ -9,7 +9,8 @@
 # express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Dict
+from abc import ABC, abstractmethod
+from typing import Dict, List
 
 import attr
 from comby import Comby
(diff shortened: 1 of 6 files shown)


experimental/piranha_playground/rule_inference/tests/test_piranha_agent.py
```diff
@@ -40,13 +40,41 @@ def test_go_inference():
...
+    assert len(graph.rules) == 1 and graph.rules[0].query == "\n".join(expected)
+
+
+def test_java_inference():
+    # Source and target code samples
+    language = "java"
+    source_code = """ // 1
+    :[x].doSomething(:[y]); 
+    """
+    target_code = """ // 1
```
ketkarameya, line 63: target code is `after` right? Why do we have to give the "kind" list in rhs and not in LHS

experimental/piranha_playground/rule_inference/static_inference.py
```diff
@@ -18,7 +19,64 @@
...
+
+    @abstractmethod
+    def process(self, cursor: TreeCursor, depth: int) -> str:
+        pass
+
+
+@attr.s
+class SimplifyStrategy(ChildProcessingStrategy):
+    """
+    Strategy to simplify the processing of a named child node.
```
ketkarameya, line 39: what does simplify imply here? 

This strategy is related to adding alternations right?


aah I see this is that smart simplification ?

danieltrt: This is for when you want simplify queries (match only current node and assign groups to its children). It’s a refactoring because the code was messy

experimental/piranha_playground/rule_inference/tests/test_piranha_agent.py
danieltrt: You can do both, only need to do it once

experimental/piranha_playground/rule_inference/static_inference.py
ketkarameya: I see. 
Let's document this method a bit

experimental/piranha_playground/rule_inference/tests/test_piranha_agent.py
ketkarameya: I see. Let's still have it on LHS (in this test) . It is more intuitive :|

ketkarameya (approved):
LGTM (modulo the nit)

PR #552, opened 2023-07-19
Attempt to add concrete syntax

danieltrt:
This PR adds a concrete syntax similar to comby

The only constraint is that each hole `:[a]` has to match a meaningful AST node


src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,89 @@
...
+      get_all_matches_for_metasyntax_aux(nodes.clone(), source_code.clone(), &meta_adv, None)
+    {
+      let mut match_map = HashMap::new();
+      match_map.insert(
+        var_name.to_string(),
+        node.utf8_text(source_code.as_bytes()).unwrap().to_string(),
+      );
+
+      inner_matches.push(Match {
+        matched_string: var_name.to_string(),
```
ketkarameya, line 64:
```suggestion
        matched_string: match_map.get(var_name.to_string()),
```

This should be the actually matched string right ?

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,89 @@
...
+pub(crate) fn get_all_matches_for_metasyntax(
+  node: &Node, source_code: String, meta: &MetaSyntax, recursive: bool,
+  _replace_node: Option<String>,
+) -> (Vec<Match>, bool) {
+  let vec_deque = vec![node.clone()].into_iter().collect::<VecDeque<_>>();
+  get_all_matches_for_metasyntax_aux(vec_deque, source_code, meta, _replace_node)
+}
+
+pub(crate) fn get_all_matches_for_metasyntax_aux(
+  mut nodes: VecDeque<Node>, source_code: String, meta: &MetaSyntax, _replace_node: Option<String>,
```
ketkarameya, line 34: lets also think how we can account for `_replace_node`

Since we have no way to name the outermost node, we can have user insert `*`   to replace the entire expression or the appropriate tag name

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,89 @@
...
+) -> (Vec<Match>, bool) {
+  let mut matches: Vec<Match> = Vec::new();
+  let re_var = Regex::new(r"^:\[(?P<var_name>\w+)\]").unwrap();
```
ketkarameya, line 37: would it make sense to stick to our `@<var_name>` format ?

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,89 @@
...
+    {
+      let mut match_map = HashMap::new();
+      match_map.insert(
+        var_name.to_string(),
+        node.utf8_text(source_code.as_bytes()).unwrap().to_string(),
+      );
+
+      inner_matches.push(Match {
+        matched_string: var_name.to_string(),
+        range: Range::from(node.range()),
```
ketkarameya, line 65: This range shud be the range of the node corresponding to `_replace_node` tag.

danieltrt: Oh!

danieltrt: But then this syntax, we don’t have replace node. Should we just replace everything always

danieltrt: I guess we could have replace node

src/models/concrete_syntax.rs
danieltrt: Sounds good!

src/models/concrete_syntax.rs
ketkarameya: yea like `:[x]` is basically a node or range captured by something named `x` ?

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }
+
+    return (matches, false);
+
+  } else if node.child_count() == 0 {
+     if (syntx.starts_with(code.trim())) {
+      let advance_by = code.len();
+      let meta_substring = MetaSyntax(syntx[advance_by..].to_string().trim_start().to_owned());
+      while (!cursor.goto_next_sibling()) {
```
ketkarameya, line 92: I saw this pattern twice. Maybe we can extract it out as an inner function  (fn inside fn) ? Don't know if that makes it easier to read

(Rust has inner functions :P )

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,129 @@
...
+
+  } else if ! recursive {
+    cursor.goto_first_child();
+    return get_all_matches_for_metasyntax_aux(cursor.clone(), source_code.clone(), meta, recursive,None);
+  } else {
+
+    let mut success = false;
+    if cursor.goto_first_child() {
+      loop {
+        if let (mut inner_matches, true) =
```
ketkarameya, line 115: could be `while let (mut inner_matches, true) = ...` ?

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,129 @@
...
+      );
+    }
+    return (matches, false)
+
+  } else if ! recursive {
+    cursor.goto_first_child();
+    return get_all_matches_for_metasyntax_aux(cursor.clone(), source_code.clone(), meta, recursive,None);
+  } else {
+
+    let mut success = false;
```
ketkarameya, line 112: not needed . We can just short circuit return from the loop?

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,129 @@
...
+        });
+        return (inner_matches, true);
+      }
+
+      if !cursor.goto_first_child() {
+        break;
+      }
+    }
+
+    return (matches, false);
```
ketkarameya, line 86:
```suggestion
    return (Vec::new(), false);
```

I think returning empty vec explicitly makes it easier to read.

src/models/concrete_syntax.rs
```diff
@@ -0,0 +1,129 @@
...
+pub(crate) fn get_all_matches_for_metasyntax_aux(
+  mut cursor: TreeCursor, source_code: String, meta: &MetaSyntax, recursive: bool, _replace_node: Option<String>,
+) -> (Vec<Match>, bool) {
+  let mut matches: Vec<Match> = Vec::new();
+  let re_var = Regex::new(r"^:\[(?P<var_name>\w+)\]").unwrap();
+  let syntx = meta.0.trim_start();
+
+
+  if syntx.is_empty() {
+    return (matches, true);
```
ketkarameya, line 41:
```suggestion
    return (Vec::new(), true);
```

ketkarameya (commented):
I think this algo is neat. Just reading the code I kinda understood what I happening. 

Just left some inline comments

src/cleanup_rules/java_dyn/rules.toml
```diff
@@ -321,49 +249,56 @@ is_seed_rule = false
...
-    (binary_expression
-        left : [(false) @lhs (true) @lhs (string_literal) (decimal_integer_literal) @lhs (decimal_floating_point_literal) @lhs]
-        operator:"=="
-        right: [(false) @rhs (true) @rhs (string_literal) @rhs (decimal_integer_literal) @rhs (decimal_floating_point_literal) @rhs]
-    )
-@binary_expression
-(#not-eq? @lhs @rhs)
-)"""
+name = "simplify_equals_equals_false_1"
+query = """dyn false == true"""
```
ketkarameya, line 286: I see how a `where` clause could help. 

We could be like `:[x] == :[y] where x != y`

ketkarameya (commented):
round 1

ketkarameya (commented):
LGTM. One suggestion to avoid the 3rd party dependency

ketkarameya (approved):

PR #555, opened 2023-07-20
Enable graph validation by default

danieltrt:
This PR changes the default of graph validation to true.
It will now panic if the graph is not correct


ketkarameya (approved):
Please update the test to use `should_panic` (based on our offline discussion)

PR #558, opened 2023-07-21
This PR fixes a bug in the dataflow analysis

danieltrt:
Previously we were not getting the correct range for the capture group names in predicates.
In this PR we fix that. Also added a new test


danieltrt:
Tests are failing because the playground uses the pip version of piranha

ketkarameya (approved):
LGTM

PR #565, opened 2023-08-01
This PR introduces a feature flag to test dynamic rules

danieltrt:
The default is false. To enable dynamic rules in java set `experiment_dyn` to `true` in PiranhaArgument.s

### DIFF
--- a/src/models/piranha_arguments.rs
+++ b/src/models/piranha_arguments.rs
@@ -287,12 +287,15 @@ impl PiranhaArgumentsBuilder {
   pub fn build(&mut self) -> PiranhaArguments {
     let _arg: PiranhaArguments = self.create().unwrap();
 
-    if let Some(piranha_language) = self.language.clone() {
-      let language = piranha_language.extension();
-      if language == ".java" && self.experiment_dyn.unwrap_or(default_experiment_dyn()) {
-        self.language = Option::from(PiranhaLanguage::from("java_dyn"));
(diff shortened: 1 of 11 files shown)


src/models/piranha_arguments.rs
```diff
@@ -275,9 +284,16 @@ impl PiranhaArgumentsBuilder {
   /// * create PiranhaArgument from the builder
   /// * parse `piranha_arguments.toml` (if it exists)
   /// * merge the two PiranhaArguments
-  pub fn build(&self) -> PiranhaArguments {
+  pub fn build(&mut self) -> PiranhaArguments {
     let _arg: PiranhaArguments = self.create().unwrap();
 
+    if let Some(piranha_language) = self.language.clone() {
```
ketkarameya, line 290: lets do something like : 

```
let mut piranha_language = self.language;

if piranha_language.filter(|x|x.extesion== ".java" && experiment){

  piranha_language = Option::from(PiranhaLanguage::from("java_dyn"));
}
.
.
_arg = PiranhaArguments { rule_graph, piranha_language, .._arg };




```

ketkarameya (approved):
nit

src/models/piranha_arguments.rs
ketkarameya: Generally I avoid mutation, and keep things a bit functional :|

PR #572, opened 2023-08-11
Introduce plugin for migrating scalatest

ketkarameya:
- Adding plugin to migrate scalatest


plugins/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[tool.poetry]
+name = "scala_test"
+version = "0.0.1"
+description = "Rules to migrate 'scaletest'"
+# Add any other metadata you need
+
+[tool.poetry.dependencies]
+python = "^3.9"
+polyglot_piranha = "*"
```
lazaroclapp, line 13: What does `"*"` mean here? Any version? Latest?

plugins/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[tool.poetry.dev-dependencies]
+pytest = "*"
+
+# [tool.poetry.scripts]
+# scala_test = "scala_test.main:main"
```
lazaroclapp, line 19: Why commented out lines?

plugins/scala_test/README.md
```diff
@@ -0,0 +1,26 @@
+# `scalatest` Migration Plugin 
+
```
lazaroclapp, line 2: This needs a description/explanation of what this is, before the Usage instructions.

Would also be a good point to note if this is a WIP or already functional and for which cases.

plugins/scala_test/main.py
```diff
@@ -0,0 +1,20 @@
+import argparse
+from update_imports import update_imports
+
+def _parse_args():
+    parser = argparse.ArgumentParser(description="Migrates scala tests!!!")
```
lazaroclapp, line 5: Migrates them to what or from what? Also, longer term, do we need a parameter for a target version which affects which mappings we use?

plugins/scala_test/recipes.py
```diff
@@ -0,0 +1,83 @@
...
+        rule_graph=rule_graph,
+        substitutions={"search_heuristic": f"{search_heuristic}"},
+        dry_run=dry_run
+    )
+    
+    return execute_piranha(args)
+    
+
+
+def replace_import_rules_edges(
```
lazaroclapp, line 36:
```suggestion
def replace_import_rules_and_edges(
```

Otherwise it reads as generating the "rules' edges" and it might be surprising that it also returns a list of rules.

Btw, these methods and `update_imports` could use docs. Maybe even those in `test_update_imports` too, but I leave that up to you.

plugins/scala_test/recipes.py
```diff
@@ -0,0 +1,83 @@
...
+def replace_imports(
+    target_new_types: dict[str, str], search_heuristic: str, path_to_codebase: str,
+    dry_run = False
+):
+    find_relevant_files = Rule(
+        name="find_relevant_files",
+        query="((identifier) @x (#eq? @x \"@search_heuristic\"))",
+        holes={"search_heuristic"},
+    )
+    e1 = OutgoingEdges("find_relevant_files", to=[f"update_import"], scope="File")
```
lazaroclapp, line 12: Interesting. This runs every rule in the "update_import" group if "find_relevant_files" matches, right?

plugins/scala_test/tests/resources/input/sample.scala
```diff
@@ -0,0 +1,7 @@
+package com.scala.piranha
+
+import com.uber.michelangelo.AbstractSparkSuite
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types.{DoubleType, StringType, StructField, StructType}
+import org.scalatest.{BeforeAndAfter, Matchers}
+import org.scalatest.mock.MockitoSugar
```
lazaroclapp, line 7: Shouldn't we have a case where an import like `import pkg1.pkg2.{A, B}` is replaced as `import pkg1.pkg2.{C, B}` and also the simple `import pkg3.pkg4.D` -> `import pkg5.pkg6.E` case?

plugins/scala_test/tests/test_update_imports.py
```diff
@@ -0,0 +1,40 @@
...
+from os import listdir
+
+from update_imports import update_imports
+# from update_imports import update_imports
+
+def test_update_imports():
+    summary = update_imports("plugins/scala_test/tests/resources/input/", dry_run=True)
+    assert is_as_expected("plugins/scala_test/tests/resources/", summary)
+
+def is_as_expected(path_to_scenario, output_summary):
```
lazaroclapp, line 14: Wonder if there is a clean way to avoid the duplication between this code and the [top level test harness logic](https://github.com/uber/piranha/blob/b16fa19a340cba45a368adda64c90bf3468e2293/tests/tests.py#L198). Maybe a shared test utilities library? Not a big deal, but if every plugin will have it's own copy of this code that might be a pain when you need to update something.

lazaroclapp (commented):
Do we want to call this `scala_test`? Looking at the directory path, I'd expect that to be some testing for Piranha, not a tool to migrate tests. Maybe `scala_test_migrator`?

plugins/scala_test/recipes.py
ketkarameya: done! Credits - CoPilot

plugins/scala_test/tests/resources/input/sample.scala
ketkarameya: hmmm. Actually the solution u suggest looks clean when the before and after type have a significant overlap in their qualified name. Else we have to "infer" the level to split the type name. 

From: `a.b.c.D` to `a.e.f.g.H`.  
Before: `import a.b.c.{D, E}` 
After: `import a.{b.c.E, e.f.H}`
or add some extra logic to heuristically decide when it is not a good idea to add a nested import, and add a simple import at those times.
-------------

From: `a.b.c.D` to `a.b.c.H`.  
Before: `import a.b.c.{D, E}` 
After: `import a.b.c.{H, E}`

-------
( actually @raviagarwal7 suggested adding simple import and keep the rewrite logic less complicated. We believe that we can use Scala linters to re-org the imports like we want to)

plugins/scala_test/tests/test_update_imports.py
ketkarameya: I agree. I wanted to that . I will eventually extract out a `commons` . 
This could moved to `commons/test_utilities`
I believe that `replace_imports` could also be a part of `commons/scala`

ketkarameya (commented):
Addressed comments

plugins/scala_test/tests/resources/input/sample.scala
lazaroclapp: Not sure I understand this. I am not saying the rewrite done in these tests is incorrect, but I think there are missing cases given the rewrite rules you added. Specifically the `update_simple_import_{...}` rules. My question here is about test coverage for the rules/logic added.

plugins/pyproject.toml
lazaroclapp: Do we want to be silently latest? Could we have this just be in sync with the current released Piranha version? (Also, `pytest` below should probably be set to a concrete library and we should manually keep the dep up to date, no?). Basically, just in terms of reproducibility I am wary of dependencies without a explicit version.

plugins/scala_test/recipes.py
```diff
@@ -1,42 +1,69 @@
...
-def replace_import_rules_edges(
+def replace_import_rules_and_edges(
     target_qualified_type_name: str, new_qualified_type_name: str
 ) -> (list[Rule], list[OutgoingEdges]):
-
+    """This function generates the rules and edges to replace the imports of the target type with the new type.
+    It supports both simple and nested imports. While the simple imports are replaced directly, the nested imports are deleted and the new type is imported (as a simple non-nested import).
+    Assume that the target type is "a.b.c.d" and the new type is "x.y.z". Then the following rules are generated:
+    import a.b.c.d -> import x.y.z
+    import a.b.c.{d, e} -> import x.y.z \n import a.b.c.{d}
```
lazaroclapp, line 65:
```suggestion
    import a.b.c.{d, e} -> import x.y.z \n import a.b.c.{e}
```

I believe you meant `e` here, since `d` is the one getting replaced...

beyeu107:
Rules LGTM

PR #573, opened 2023-08-17
Bug Fix: Avoid over-deletion when looking for associated comma

ketkarameya:
Previously, we were walking up the parent to find the associated trailing/leading comma. 
But we did not have any safety checks guarding this comma deletions (like we have for comments).

Lets say in the below scenario we delete ` isFlgEnabled := exp.BoolValue(staleFlagConst)` :
```
 t.Run("message", func(t *Foobar.T) {
         isFlgEnabled := exp.BoolValue(staleFlagConst)
         fmt.Println("some logging statement")
     })
```

Our comment/comma cleanup logic would kick in and cleanup stuff like below:
```
 t.Run("message"
         fmt.Println("some logging statement")
     })
```
Note we deleted the extra `, func(t *Foobar.T) {`

I introduced a simple safety check that ensures that there is no node between the comma and the node intended to be deleted.

### DIFF
--- a/test-resources/go/feature_flag/system_1/const_same_file/input/sample.go
+++ b/test-resources/go/feature_flag/system_1/const_same_file/input/sample.go
@@ -87,6 +87,19 @@ func callerFunc() {
 
 // should not replace the function name
 func isFlagEnabledFunc() bool {
+    // Should not delete this comment and the func below
+    t.Run("message", func(t *Foobar.T) {
+        isFlgEnabled := exp.BoolValue(staleFlagConst)
+        fmt.Println("some logging statement")
+    })
+    
+    // Should not delete this comment and the func below
+    t.Run1(func(t *Foobar.T) {
+        isFlgEnabled1 := exp.BoolValue(staleFlagConst)
+        fmt.Println("some other logging statement")
+    }, "other_message" )
+
+
     isFlagEnabledFunc := exp.BoolValue(staleFlagConst)
 
     if !isFlagEnabledFunc {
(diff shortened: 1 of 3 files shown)


test-resources/go/feature_flag/system_1/const_same_file/input/sample.go
```diff
@@ -87,6 +87,19 @@ func callerFunc() {
...
+    t.Run("message", func(t *Foobar.T) {
+        isFlgEnabled := exp.BoolValue(staleFlagConst)
+        fmt.Println("some logging statement")
+    })
+    
+    // Should not delete this comment and the func below
+    t.Run1(func(t *Foobar.T) {
+        isFlgEnabled1 := exp.BoolValue(staleFlagConst)
+        fmt.Println("some other logging statement")
+    }, "other_message" )
```
lazaroclapp, line 100: In both these examples, the comma is before the deleted node, but the change handles both before and after, right? Any test case case you can think of for the "after" case?

lazaroclapp (commented):
Fix makes sense to me, but see comment about potentially missing test case. If there isn't any case that would fit, then let me know and I'll approve as is :)

test-resources/go/feature_flag/system_1/const_same_file/input/sample.go
ketkarameya: aah I see ur point. I actually had a test case like that, but I had it slightly wrong. I reordered the statements here. 
In our earlier logic we would delete `},` in addition to `isFlgEnabled1 := exp.BoolValue(staleFlagConst)` 

But now we don't, because we see that there is a node which ends between the ending of deleted node and comma

ketkarameya (commented):
addressed comments

test-resources/go/feature_flag/system_1/const_same_file/input/sample.go
lazaroclapp: Ok, but this change makes the comment above outdated, not?

```
// Should not delete this comment and the func below
```

It should now say something about the trailing `},`

ketkarameya: good catch. Addressed.

lazaroclapp (approved):
👍

PR #578, opened 2023-08-22
Fix replace_identifier_with_nil rule

ketkarameya:
Fix the enclosing node from block to function_declaration for 
`replace_identifier_with_nil` rule


danieltrt (approved):
LGTM

PR #580, opened 2023-08-22
tree-sitter-go pointing to a new branch

dvmarcilio:
- branch is now up to date with upstream
- `statement_list` still exposed
- `interpreted_string_literal_basic_content` newly exposed
- https://github.com/uber/tree-sitter-go/commit/f8cffd0af7baaf7bf6062e403efe7c0d06319c41

PR #581, opened 2023-08-22
Update go grammar

ketkarameya:
Updated the go grammar to account for string literals (https://github.com/uber/piranha/pull/580).
 Also checked if the fix solves the piranha bug around parsing string literals when using concrete syntax


danieltrt (approved):
LGTM!

PR #584, opened 2023-08-28
Swift: Fix cleaning of statements after return

Adarsh-NP:
This PR fixes the issues where we were seeing improper cleanup of statements after a return statement was brought in the function scope post an if/guard cleanup. Please refer to tests to understand the cases in more details.


ketkarameya (approved):
LGTM.

PR #589, opened 2023-08-31
Introduce execute piranha strategy for plugins

ketkarameya:
Introduces execute piranha strategy for plugins


plugins/spark_upgrade/execute_piranha.py
```diff
@@ -0,0 +1,117 @@
+#  Copyright (c) 2023 Uber Technologies, Inc.
```
raviagarwal7, line 1: we might want to move it outside in plugins/base so all plugins can use this

raviagarwal7 (approved):

plugins/spark_upgrade/execute_piranha.py
ketkarameya: Since forever, we have maintained one LICENSE file for entire repo and such license comments for each file. :|

PR #610, opened 2023-09-11
Fix empty switch entries and statements

Adarsh-NP:
This PR:
* solves the cases where there are empty switch entries post refactoring
*  Also has a rule to clean empty switch statements left after the aforementioned cleanup. Please refer to tests for more info. 
* Re-organises tests in post_processing_cleanup that includes delete_statements_after_return and aforementioned switch refactoring.


ketkarameya (approved):
LGTM

PR #613, opened 2023-09-13
Fix Variable Inline issues related to parameter cleanup in adhoc cases

Adarsh-NP:
This diff fixes the issues caught when Piranha was run in our swift code. The issues broadly revolved around adhoc cases of variable cleanup like being passed in a function as an argument. Please refer to the new tests added to understand more about these adhoc cases. 

In case CI tests fail, pls rerun after landing #612

### DIFF
--- a/src/cleanup_rules/swift/rules.toml
+++ b/src/cleanup_rules/swift/rules.toml
@@ -769,6 +769,9 @@ not_contains = [
     )""",
 ]
 
+[[rules.filters]]
+not_enclosing_node = "(value_argument_label) @va"
+
 # rule to delete declaration of variable in class scope
 [[rules]]
 name = "delete_field_declaration"
--- a/src/cleanup_rules/swift/edges.toml
+++ b/src/cleanup_rules/swift/edges.toml
@@ -36,8 +36,8 @@ to = ["replace_identifier_with_value", "delete_parent_assignment"]
 scope = "Class"
 from = "delete_field_initialisation_init"
 to = [
-  "replace_self_identifier_with_value",
   "replace_identifier_with_value",
+  "replace_self_identifier_with_value",
   "delete_field_declaration",
 ]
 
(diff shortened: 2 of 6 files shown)


Adarsh-NP:
There is some work to be done in this, pls don't merge it.

Adarsh-NP:
Please don't land this, I am working on another feature that can have this as a dependency. I'll update when it's ready for landing.

src/cleanup_rules/swift/rules.toml
```diff
@@ -769,6 +769,39 @@ not_contains = [
...
+            name: (value_argument_label)
+            value: [
+                (simple_identifier) @revar @valere
+                (navigation_expression
+                    target: (self_expression)
+                    suffix: (navigation_suffix
+                        suffix: (simple_identifier) @revar
+                    )
+                ) @valere
+                (ternary_expression
```
ketkarameya, line 788: do we still need this ternary expression pattern here?

src/cleanup_rules/swift/rules.toml
```diff
@@ -769,6 +769,39 @@ not_contains = [
     )""",
 ]
 
+[[rules.filters]]
+not_enclosing_node = "(value_argument_label) @va"
+
+[[rules]]
```
ketkarameya, line 775: I think we won't need this rule at all right? 

Since now anything under `value_argument_label` does not get transformed (as per the update to the above rule)

ketkarameya (changes requested):

src/cleanup_rules/swift/rules.toml
Adarsh-NP: Yes, but we need this to clean for arguments right?

src/cleanup_rules/swift/rules.toml
Adarsh-NP: Yep, for the cases where arguments are passed on the basis of ternary condition evaluation.

ketkarameya:
ya it will not update the label in `x: x`, but it will update the value.

PR #617, opened 2023-09-20
Add a new edge kind - `ParentIterative`

ketkarameya:
Earlier we had `Parent` and this is how it was processed: 
```
for rule in graph.outgoing(current_rule, PARENT):
   for ancestor in context(current_location):
       Edit edit = get_edit(rule, ancestor)
       if edit : 
          return edit 
```

For `ParentIterative` we do something subtly different. We loop over the ancestors in the outer most loop: 
And its processed something like : 
```
   for ancestor in context(current_location):
       for rule in graph.outgoing(current_rule, PARENT_ITERATIVE):
         Edit edit = get_edit(rule, ancestor)
         if edit : 
          return edit 
```

Why suffix iterative? I thought its keeping to iterative-deepening idea from DFS. 
I am open to other names. 
But this pattern makes it easy to transform builder patterns (As required by Scala upgrade folks)


test-resources/java/spark_conf/input/sample.java
```diff
@@ -0,0 +1,29 @@
...
+      .setSparkHome(sparkHome2)
+      .setExecutorEnv("spark.executor.extraClassPath2", "test2");
+
+    val conf3 = new SparkConf()
+      .setExecutorEnv("spark.executor.extraClassPath3", "test3")
+      .setAppName(appName3)
+      .setMaster(master3)
+      .setSparkHome(sparkHome3);
+
+
```
lazaroclapp, line 26: Are this extra empty lines intentional?

test-resources/java/spark_conf/configurations/rules.toml
```diff
@@ -0,0 +1,50 @@
...
+# express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+[[rules]]
+name = "spark_conf_change_java_scala"
+query = "cs new SparkConf()"
+replace_node = "*"
+replace = 'new SparkSession.builder()'
+
+
```
lazaroclapp, line 18: Minor nit: but let's make the spacing between rules consistently, either one or two empty lines, but not a mix of the two unless it implies a particular grouping.

src/models/source_code_unit.rs
```diff
@@ -280,8 +267,8 @@ impl SourceCodeUnit {
     stack: &mut VecDeque<(CGPattern, InstantiatedRule)>,
   ) {
     for (scope_level, rules) in next_rules_by_scope {
-      // Scope level is not "PArent" or "Global"
-      if ![PARENT, GLOBAL].contains(&scope_level.as_str()) {
+      // Scope level is not "Parent" or "Global"
```
lazaroclapp, line 270:
```suggestion
      // Scope level is not "Parent", "Parent Iterative", or "Global"
```

src/models/source_code_unit.rs
```diff
@@ -240,33 +240,20 @@ impl SourceCodeUnit {
...
-        *current_replace_range.start_byte(),
-        *current_replace_range.end_byte(),
-        rules_store,
-        &next_rules_by_scope[PARENT],
-      ) {
+      if let Some(edit) =
+        self.get_edit_for_ancestors(&current_replace_range, rules_store, &next_rules_by_scope)
+      {
         self.rewrites_mut().push(edit.clone());
-        debug!(
```
lazaroclapp, line 250: Just to confirm, we are intentionally getting rid of this debug log line entirely, right?

lazaroclapp (approved):
Minor nits below, plus the offline request to add the negative test case, but otherwise this LGTM!

PR #632, opened 2023-11-11
Update dependencies to latest and fix some bugs

kageiit:
- Update dependencies in Cargo.toml to latest
- Remove str methods macro because it was being [silently ignored](https://github.com/PyO3/pyo3/issues/2862). Due to [limitations](https://pyo3.rs/v0.20.0/features#multiple-pymethods) of pyo3, it is not recommended to have multiple `#[pymethods]` on impls which makes it tricky to have a macro in all locations it is currently used it. Since the amount of code saved is quite minimal, removed the macro instead. Another option could be to use a custom trait to provide the python str methods


ketkarameya (approved):

ketkarameya:
makes sense

PR #633, opened 2023-11-17
Adding the option to use :[x] instead of @x

danieltrt:
This changes the instantiation method to make it more intuitive when using concrete syntax

It is now possible to

```
match = if (true) :[consq] else :[alt]
replace_node = "*"
replace = :[consq]
```


ketkarameya (commented):
can u please add a test :)

ketkarameya (approved):
Lgtm

PR #655, opened 2024-03-27
Fix 3rd party dependency that's breaking the build

danieltrt:
## Build Failure Resolution

I've encountered a build failure due to a recent update in the `tree-sitter-kotlin` dependency, which has led to conflicting versions of the `tree-sitter` crate.

Previously, we were pointing to the latest version rather than specifying a fixed version (which was updated two weeks ago).

To resolve this issue, I've reverted the dependency to the last known working version, `0.3.1`, which appears to function correctly on my end.

For a detailed version history, please check the [tree-sitter-kotlin releases](https://github.com/fwcd/tree-sitter-kotlin/releases).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ env_logger = "0.10.0"
 tempdir = "0.3"
 serde_json = "1.0.82"
 
-tree-sitter-kotlin = { git = "https://github.com/fwcd/tree-sitter-kotlin.git" }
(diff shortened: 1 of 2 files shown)


ketkarameya (approved):
LGTM

danieltrt:
Seems like there's an orthogonal test failing? I assume some grammar changed

ketkarameya (approved):

PR #658, opened 2024-04-29
Updating Spark Migration Rules

dvmarcilio:
* Adding a Java rule for `SparkSession.builder()` wrapped by `new JavaSparkContext()`
* Adding a Scala rule for `SparkSession.builder() ` wrapped by `Mockito.spy()`
* Adding filters to rules that were on infinite loop
* `glob` for Java and Scala files, then running rules for each file individually. With this, we can wrap each file in a try/catch to prevent the whole execution from stopping when there is a parse error on a single file.
* Fixing tests to point to correct directories 
* Adding an integration test to check that execution terminates when applying all rules at once.

The PR has several commits, but we can squash them when merging to main.


stefanheule (approved):
Looks good to me.

PR #671, opened 2024-06-03
Better concrete syntax semantics 

danieltrt:
## Overview

This pull request introduces two significant changes to the semantics of the concrete syntax matching algorithm. These changes enhance the flexibility and expressiveness of the pattern matching with concrete syntax.

#### Previous Semantics

Previously, template variables `:[var]` could only match entire subtrees. This means it was not possible to match sequences of sibling nodes. For example, if matching `[var] = 0;` against `int x = 0;`, the match would fail because `int` and `x` are sibling nodes and not part of the same subtree.

graphviz-3

#### New Semantics

Now template variables can match multiple sequential siblings at the same level. So matching 
`[var] = 0;` with `int x = 0;` is possible.

graphviz-4

#### Previous Semantics

Previously, the algorithm required that the concrete template match an entire node, and only one node. This restriction limited the pattern matching to single statements rather than sequences of statements. For example:

```
int :[var] = 0;
:[var]++;
```

This template could never match, because it corresponds to two sibling nodes, rather than a single subtree. For example, the snippet 

```cpp
{
  int x = 0;
  x++;
  print(x);
}
```
has this tree representation. 

graphviz-5

Notice that `int x = 0`; and `x++`; are sibling nodes. The template would never match just ONE node.

#### New Semantics

Now, the template can match entire sequential sibling nodes. This allows the template to match sequences of statements.

For example, the following code:

**Code:**
```cpp
// just a regular loop in C++
int some = 0; 
while(some < 100) { 
    float length = 3.14;
    float area = length * length;
    some++; 
}
print(some);
```

matches the following template.

**Template:**
```plaintext
int :[var] = 0; 
while(:[var] < 100) { 
    :[body]
    
    :[var]++; 
}
```

and the alignment is as follow (loosely)

graphviz-6


src/tests/test_piranha_java.rs
```diff
@@ -426,8 +426,8 @@ fn test_dyn_rule() {
   let rule = piranha_rule! {
     name = "match_class",
     query = "cs println(:[xs])",
-    replace_node = "xs",
-    replace = "@xs, 2"
+    replace_node = "*",
+    replace = "println2(@xs, 2)"
```
danieltrt, line 430: Since :[x] can match multiple sequential siblings, this would be an infinite loop if we only rewrote `:[x]`

src/tests/test_piranha_java.rs
```diff
@@ -463,7 +463,10 @@ fn test_multiple_code_bases() {
   // Note that we expect 2 matches because we have 2 code bases, and each code base has 1 match.
   // We also have another codebase `folder_3` but the `paths_to_codebase` does not include it.
   assert_eq!(output_summaries.len(), 2);
-  assert_frequency_for_matches(&output_summaries, &HashMap::from([("match_import", 2)]));
+  assert_frequency_for_matches(&output_summaries, &HashMap::from([("match_import", 4)]));
```
danieltrt, line 466: The snippet

```
package org.piranha.examples;

import java.util.List;
```

```
[0, 0] - [4, 0]
[package_declaration] [0, 0] - [0, 29]
[scoped_identifier] [0, 8] - [0, 28]
    scope: [scoped_identifier] [0, 8] - [0, 19]
        scope: [identifier] [0, 8] - [0, 11]
        name: [identifier] [0, 12] - [0, 19]
   name: [identifier] [0, 20] - [0, 28]
[import_declaration] [2, 0] - [2, 22]
    [scoped_identifier] [2, 7] - [2, 21]
        scope: [scoped_identifier] [2, 7] - [2, 16]
            scope: [identifier] [2, 7] - [2, 11]
            name: [identifier] [2, 12] - [2, 16]
    name: [identifier] [2, 17] - [2, 21]
```

Now has two ways to match. Entire the entire sequential children of the `import_declaration` node, or the last node of the `program` node

src/models/concrete_syntax.rs
```diff
@@ -153,35 +219,49 @@ pub(crate) fn get_matches_for_node(
...
-      if let (mut recursive_matches, true) =
-        get_matches_for_node(&mut tmp_cursor, source_code, &meta_advanced)
-      {
-        // If we already matched this variable, we need to make sure that the match is the same. Otherwise, we were unsuccessful.
-        // No other way of unrolling exists.
-        if recursive_matches.contains_key(var_name)
-          && recursive_matches[var_name].text.trim() != current_node_code.trim()
-        {
-          return (HashMap::new(), false);
+      let mut should_match = find_next_sibling(&mut tmp_cursor);
```
danieltrt, line 223: determine whether the next recursive call to `get_matches_for_node` should match any node. If there are no "siblings" left, then we should not match anything

src/models/concrete_syntax.rs
```diff
@@ -82,20 +85,70 @@ pub(crate) fn get_all_matches_for_concrete_syntax(
...
+      let mut tmp_cursor = cursor.clone();
+      let (mapping, mut matched, indx) =
+        get_matches_for_node(&mut tmp_cursor, source_code, meta, true, the_node);
+
+      if matched {
+        // Determine the last matched node. Remember, we are matching subsequences of children [n ... k]
+        let mut last_node = the_node.child(the_node.child_count() - 1);
+        if let Some(last_node_index) = indx {
+          last_node = the_node.child(last_node_index);
+          matched = matched && (last_node_index != child_incr || the_node.child_count() == 1);
```
danieltrt, line 138: A bit convoluted and should be refactored.

Here we say, we match if:

1. we were able to align the sequence with the templates
2. the sequence is of length > 1, UNLESS the node only has one child

src/models/concrete_syntax.rs
```diff
@@ -117,99 +177,168 @@ fn find_next_sibling(cursor: &mut TreeCursor) {
...
   }
-  (HashMap::new(), false)
+  (HashMap::new(), None)
+}
+
+fn find_last_matched_node(cursor: &mut TreeCursor, parent_node: &Node) -> Option<usize> {
+  parent_node
+    .children(&mut parent_node.walk())
+    .enumerate()
+    .find_map(|(i, child)| {
```
ketkarameya, line 330: use `filter` instead of `if` inside `find_map(...)`

src/models/matches.rs
```diff
@@ -321,6 +321,15 @@ impl Range {
       end_point: position_for_offset(source_code.as_bytes(), mtch.end()),
     }
   }
+
+  pub(crate) fn from_siblings(left: tree_sitter::Range, right: tree_sitter::Range) -> Self {
```
ketkarameya, line 325: is this function like merging two ranges?

src/models/concrete_syntax.rs
```diff
@@ -82,134 +87,272 @@ pub(crate) fn get_all_matches_for_concrete_syntax(
...
+///
+/// 1. Initialize cursor to the first child and iterate through siblings.
+/// 2. Use `get_matches_for_subsequence_of_nodes` to attempt matching the template against a sequence of subtree starting at each sibling.
+/// 3. If a match is found, determine the range of matched nodes subtrees (i.e., [2nd,..., 4th], and return the match mapping, and range.
+/// 4. If no match is found, return an empty mapping, and None for range.
+fn match_sequential_siblings(
+  cursor: &mut TreeCursor, source_code: &[u8], cs: &ConcreteSyntax,
+) -> Option<MatchResult> {
+  let parent_node = cursor.node();
+  let mut child_incr = 0;
```
ketkarameya, line 125: could be renamed to `child_seq_match_start`

ketkarameya (approved):

stefanheule (approved):

PR #678, opened 2024-07-10
Release Workflow: multiple py versions and arm

dvmarcilio:
Using a matrix strategy in the "Release Polyglot Piranha" workflows jobs to change python versions within [3.8; 3.12].
Moreover, adding more jobs to release for linux and macos ARM.

Successful execution of the workflow: https://github.com/uber/piranha/actions/runs/9909648923
Successful release of wheels in TestPyPi: https://test.pypi.org/project/polyglot-piranha-test/0.3.253a0/#files

### DIFF
--- a/.github/workflows/polyglot_release.yml
+++ b/.github/workflows/polyglot_release.yml
@@ -3,10 +3,11 @@ on:
   workflow_dispatch:
 jobs:
   build_ubuntu:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        os: [ubuntu-latest, ubuntu-20.04]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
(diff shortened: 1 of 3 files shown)


.github/workflows/polyglot_release.yml
```diff
@@ -2,81 +2,49 @@ name: Release Polyglot Piranha
 on:
   workflow_dispatch:
 jobs:
-  build_ubuntu:
-    runs-on: ubuntu-latest
+  release:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, ubuntu-20.04, macos-latest]
```
stefanheule, line 8: how about having a target here instead of the OS? because that's what we are really trying to have a matrix for: `linux-x86, linux-arm, macos-arm`.

.github/workflows/polyglot_release.yml
```diff
@@ -2,81 +2,49 @@ name: Release Polyglot Piranha
...
+        rustup target add x86_64-apple-darwin
+        rustup target add aarch64-apple-darwin
+    - name: pip install --upgrade maturin
       run: |
         pip install --upgrade maturin
+    - name: Build wheel with Maturin linux x86
+      if: ${{ matrix.os == 'ubuntu-20.04' }}
+      run: |
         maturin build --release -o dist
         maturin build --sdist -o dist
```
stefanheule, line 38: Lets maybe have a separate step for the source distribution? It doesn't seem like it belongs with Linux x86 necessarily.

If we do that, we can combine all matrin steps (other than src) into one.

stefanheule (commented):
Looks great, thank you very much!

.github/workflows/polyglot_release.yml
```diff
@@ -1,82 +1,45 @@
...
+        echo "linker = \"aarch64-linux-gnu-gcc\"" >> .cargo/config.toml
+    - name: Setup rustup target macOS-arm
+      if: ${{ matrix.target == 'macos-arm' }}
+      run: |
+        rustup target add x86_64-apple-darwin
+        rustup target add aarch64-apple-darwin
+    - name: Build wheel with Maturin ${{ matrix.target }}
+      run: |
+        maturin build --release -o dist --target ${{ fromJSON('{"linux-x86":"x86_64-unknown-linux-gnu","linux-arm":"aarch64-unknown-linux-gnu","macos-arm":"universal2-apple-darwin"}')[matrix.target] }} -i ${{ matrix.python-version }}
+    - name: Build source distribution
```
stefanheule, line 38: Maybe say "Build source distribution (only once)" or add a comment next to the if to say why we are doing this only once.

stefanheule (approved):

PR #682, opened 2024-07-29
linux-arm release: install g++

dvmarcilio:
Fix for [error](https://github.com/uber/piranha/actions/runs/10113433225/job/27969800161#step:7:298) in the release workflow after we added the YAML dependency.
Successfully reproduced and fixed when running `maturin build --release` on a local Debian machine.


yuxincs (approved):
Nice find!

PR #688, opened 2024-09-02
Use language names

stefanheule:


polyglot_piranha.pyi
```diff
@@ -13,7 +13,7 @@ from __future__ import annotations
 from typing import List, Optional, Literal
 
 # Languages that Piranha supports (see ./src/models/language.rs)
-PiranhaLanguage = Literal["java", "kt", "kotlin", "go", "py", "swift", "ts", "tsx", "thrift", "strings", "scm", "scala", "rb", "yaml", "yml"]
+PiranhaLanguage = Literal["java", "kt", "kotlin", "go", "python", "swift", "typescript", "tsx", "thrift", "strings", "scm", "scala", "ruby", "yaml", "yml"]
```
yuxincs, line 16: Quick question: is this a safe change? 

Looking at the names here it feels that we are probably also using this as file extensions? 

Having `kt` and `kotlin` here also seems to agree with that theory. Should we include `py` and `ts` here as well?

src/models/default_configs.rs
```diff
@@ -25,15 +25,15 @@ pub const JAVA_CS: &str = "java_cs";
 pub const KOTLIN: &str = "kt";
 pub const KOTLIN_ALIAS: &str = "kotlin";
```
yuxincs, line 26: Continuing the discussion above, here we have an alias for kotlin, can we look at the uses of this and see if we need `PYTHON_ALIAS` and `TYPESCRIPT_ALIAS`?

yuxincs (approved):
Two questions due to my unfamiliarity with the codebase :)

But stamping in case you think it is actually not a problem

stefanheule:
Discussed offline.

polyglot_piranha.pyi
danieltrt: a bit of the blue, but I added the this alias because it would otherwise break something. I don't remember exactly what

PR #694, opened 2024-09-09
Kotlin cleanup rules to support `!true &&` and `!false ||` expressions

daveight:
I noticed an issue when processing boolean expressions:
`if (!feature.isEnabled(FEATURE_A) && condition != "constant")`

Piranha refactors it like this:
`if (!condition != "constant")` which is syntactically and logically incorrect

I implemented two additional rules to fix it:

1) simplify not true in conjunction expression

`!true && ...` => the whole boolean expression will be refactored to `false`

2) simplify not false in disjunction expression

`!false || ... ` => the whole boolean expression will be refactored to `true`

Looking forward to a feedback / review. Thanks!

### DIFF
--- a/src/cleanup_rules/kt/rules.toml
+++ b/src/cleanup_rules/kt/rules.toml
@@ -204,6 +204,25 @@ replace = "true"
 replace_node = "prefix_expression"
 is_seed_rule = false
 
+# Before :
+#  !false ||
+# After :
+#  true
+#
+[[rules]]
+groups = ["boolean_expression_simplify"]
+name = "simplify_not_false_disjunction"
+query = """
+(
+(prefix_expression (disjunction_expression (_) @nested_expression)) @prefix_expression
+(#match? @nested_expression "^false$")
+(#match? @prefix_expression "!.*")
+)
+"""
+replace = "true"
+replace_node = "prefix_expression"
+is_seed_rule = false
+
 # Before : 
 #  !true
 # After :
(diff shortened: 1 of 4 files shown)


danieltrt:
Hi @daveight 

The problem here is not piranha, but rather the tree sitter grammar of Kotlin. That entire expression is getting parsed this way
Screenshot 2024-10-10 at 5 32 32 PM

daveight:
Hello, I totally agree with you. 

However for my project introducing this `workaround` rule would be a fastest way to solve the problem, however the problem is that I cannot configure this rule externally, because in order this to make correctly - the rule should be executed in a specific order.

Is there a way to configure this order in Piranha for an external rule?
Thanks

danieltrt:
I think you just inject these custom rules at run-time instead:


```python
from polyglot_piranha import execute_piranha, PiranhaArguments, Rule, RuleGraph, OutgoingEdges

# Original code snippet
code = """
class Sample {
   fun test_fn() {
    if (!feature.isEnabled(FEATURE_A) && condition != "constant") {
        // if branch
    } else {
        // else branch
    }
   }
   
}
"""

r1 = Rule(
    name="replace_method",
    query="""cs !:[x].isEnabled(@stale_flag_name) && :[rest+]  
    """, # if :[x].isEnabled(@stale_flag_name) is now permanently true
    replace_node="*",
    replace="false",
    holes={"stale_flag_name"},
    is_seed_rule=True
)

edge = OutgoingEdges("replace_method", to=["boolean_literal_cleanup"], scope="Parent")

# Create Piranha arguments
piranha_arguments = PiranhaArguments(
    code_snippet=code,
    language="kt",
    rule_graph=RuleGraph(rules=[r1], edges=[edge],),
    substitutions={"stale_flag_name": "FEATURE_A"}

)

piranha_summary = execute_piranha(piranha_arguments)
print(piranha_summary[0].content)
```

danieltrt:
@daveight Did it work for you?

daveight:
Hello @danieltrt thanks for the reply. Is there a reference to learn this syntax? I am using tree-sitter queries, but this syntax I am not aware of. Thanks.

danieltrt:
Closing this since it has been fixed since 0.3.33

PR #703, opened 2024-10-18
add csharp support

GTB3NW:
Initial csharp support, I'm hoping to contribute further support in the future

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,11 +8,11 @@ include = ["pyproject.toml", "src/"]
 exclude = ["legacy"]
 license-file = "LICENSE"
 categories = [
-  "structural find-replace",
-  "find-replace",
-  "structural seearch",
-  "structural search and replace",
-  "tree-sitter",
+    "structural find-replace",
+    "find-replace",
+    "structural seearch",
+    "structural search and replace",
+    "tree-sitter",
 ]
 
 [[bin]]
@@ -68,6 +68,8 @@ pyo3 = "0.20.0"
 pyo3-log = "0.9.0"
 glob = "0.3.1"
 lazy_static = "1.4.0"
+# newer versions of "tree-sitter-c-sharp" require that we bump "tree-sitter" version
+tree-sitter-c-sharp = "0.20.0"
 
 [features]
 extension-module = ["pyo3/extension-module"]
--- a/POLYGLOT_README.md
+++ b/POLYGLOT_README.md
@@ -113,7 +113,7 @@ Polyglot Piranha can be used as a python library or as a command line tool.
 
 `pip install polyglot-piranha`
 
-Currently, we support one simple API (`execute_piranha`), a simple python wrapper around Polyglot Piranha's CLI. 
+Currently, we support one simple API (`execute_piranha`), a simple python wrapper around Polyglot Piranha's CLI.
 We believe this makes it easy to incorporate Piranha in *"pipelining"*.
 
 <h4> <code>execute_piranha</code></h4>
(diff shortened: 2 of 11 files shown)


danieltrt:
Hey! @GTB3NW 

Thanks for the contribution! Is there any specific use case you’re targeting with csharp? Adding language support is generally comprised of two steps: (1) Adding the actual tree sitter grammar, and (2) adding the tree sitter rules for the cleanups in that language (and corresponding tests). (see src/cleanup_rules)

GTB3NW:
> Hey! @GTB3NW
> 
> Thanks for the contribution! Is there any specific use case you’re targeting with csharp? Adding language support is generally comprised of two steps: (1) Adding the actual tree sitter grammar, and (2) adding the tree sitter rules for the cleanups in that language (and corresponding tests). (see src/cleanup_rules)

@danieltrt no problem!

I'd eventually love to start using this for codemods. I've had some good usage of ast-grep but I think pirhana is more suitable for building out a good suite of rules.

I could look at adding feature flag rules first. Is that a requirement for PR approval? I was just looking to layer the PR's and add some initial support and a test for the language and then go from there if that's okay?

ketkarameya (approved):
Awesome. Thanks for the contribution.

danieltrt:
No, I was just making sure you wanted to do feature flag cleanup, we would need those rules too!

GTB3NW:
> No, I was just making sure you wanted to do feature flag cleanup, we would need those rules too!

I'll give it a shot! @danieltrt

danieltrt:
Ah, can you share the code you're trying out? To declare a rule as global, you need to add the keyword `is_seed_rule = TRUE`.
We have a research paper that describes at a high level how Piranha works: https://dl.acm.org/doi/pdf/10.1145/3656429 

Screenshot 2024-10-22 at 11 49 46 AM

You should be able to declare a rule as global / seed in the TOML file

stefanheule:
Thank you for your contribution! Just as a heads up, we won't be able to merge this PR as we do not use C# at Uber. We've updated the readme to make this more clear. Happy to keep this PR open in case you or others want to continue referring to it.

GTB3NW:
> Ah, can you share the code you're trying out? To declare a rule as global, you need to add the keyword `is_seed_rule = TRUE`. We have a research paper that describes at a high level how Piranha works: https://dl.acm.org/doi/pdf/10.1145/3656429
> Screenshot 2024-10-22 at 11 49 46 AM
> 
> You should be able to declare a rule as global / seed in the TOML file

I think I was missing the context a little. I managed to get it working by specifying a rules.toml in the test and it seems like it treats that as a seed rule which can then trigger other rules. Some of the tests I was looking at didn't have a rules.toml in the structure, but they were not using the macro's and instead define some seed rules.

GTB3NW:
> Thank you for your contribution! Just as a heads up, we won't be able to merge this PR as we do not use C# at Uber. We've updated the readme to make this more clear. Happy to keep this PR open in case you or others want to continue referring to it.

There are more references - https://github.com/uber/piranha/blob/master/POLYGLOT_README.md#languages-supported

I know you're just the messenger, but it's really sad to read this. Clearly at one point this wasn't the case and it's Uber's moral duty to be explicit about its commitment to maintenance - "we will only accept contributions to the core & any of these listed languages".

I'd rather fork and accept contributions than have a lonely PR here. If I do end up forking, would you update the language support with a link to a fork with "extended and unsupported languages" with a link to it?

stefanheule:
Thank you, we've updated those references (and a few others that we found) too. I've also added a link to forks in the README.md, so feel free to fork the repository and add C# to that fork.

GTB3NW:
> Thank you, we've updated those references (and a few others that we found) too. I've also added a link to forks in the README.md, so feel free to fork the repository and add C# to that fork.

I appreciate it thank you!

PR #714, opened 2025-02-03
Add a new argument for custom internal rules

norhh:


dvmarcilio (approved):
LGTM. thanks for updating it

PR #720, opened 2025-05-23
Add delete_comments Flag to Rules

danieltrt:
Allows rules to optionally preserve comments during code transformations. Defaults to current behavior (deletion), but can be disabled per rule

### DIFF
--- a/polyglot_piranha.pyi
+++ b/polyglot_piranha.pyi
@@ -227,6 +227,8 @@ class Rule:
     "Filters to test before applying a rule"
     is_seed_rule: bool
     "Marks a rule as a seed rule"
+    delete_comments: bool
+    "Marks a rule as a seed rule"
 
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


polyglot_piranha.pyi
```diff
@@ -227,6 +227,8 @@ class Rule:
     "Filters to test before applying a rule"
     is_seed_rule: bool
     "Marks a rule as a seed rule"
+    delete_comments: bool
+    "Marks a rule as a seed rule"
```
danieltrt, line 231:
```suggestion
    "Make comment deleting optional"
```

src/models/unit_tests/rule_test.rs
```diff
@@ -381,3 +381,66 @@ fn test_satisfies_filter_not_enclosing_node_negative() {
...
+            // Then
+            assertTrue(\"This is a test\", result);
+          }
+        }";
+
+  let mut rule_store = RuleStore::default();
+
+  let args = PiranhaArgumentsBuilder::default()
+    .paths_to_codebase(vec![UNUSED_CODE_PATH.to_string()])
+    .language(PiranhaLanguage::from(JAVA))
```
norhh, line 416: Can you run this test with `cleanup_comments` set to `True`

PR #723, opened 2025-05-28
Changes the kotlin grammar dependency to my version

danieltrt:
Kotlin's OS grammar parser all expressions wrong. 
For example -a + b gets parsed to `-(a+b)`. This PR changes the grammar dependency to my version, which fixes that problem

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ serde_json = "1.0.82"
 ctrlc = "3.4"
 
 # TODO: Update if we upgrade tree-sitter to >=0.21
-tree-sitter-kotlin = "=0.3.5"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -47,7 +47,7 @@ serde_json = "1.0.82"
 ctrlc = "3.4"
 
 # TODO: Update if we upgrade tree-sitter to >=0.21
-tree-sitter-kotlin = "=0.3.5"
+tree-sitter-kotlin = { git = "https://github.com/danieltrt/tree-sitter-kotlin", rev = "8eb82b294900b66965ecfe39cdcbb8b16c99ff71" }
```
yuxincs, line 50: Perhaps adding a TODO here documenting why we have to do this, and call for upstreaming it when we upgrade tree-sitter >=0.23

yuxincs (approved):

yuxincs (approved):

PR #724, opened 2025-05-28
API changes for deciding whether or not to keep a comment at the rule level

danieltrt:
API changes for deciding whether or not to keep a comment at the rule level.
Now instead of booleans, we take a set of regex

### DIFF
--- a/src/models/matches.rs
+++ b/src/models/matches.rs
@@ -20,6 +20,8 @@ use log::trace;
 use pyo3::prelude::{pyclass, pymethods};
 use serde_derive::{Deserialize, Serialize};
 use tree_sitter::Node;
+use regex::Regex;
+use std::collections::HashSet;
 
 use super::{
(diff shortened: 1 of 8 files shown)


yuxincs (approved):

src/models/matches.rs
```diff
@@ -201,13 +203,16 @@ impl Match {
...
-        {
-          // Add the comment to the associated matches
-          self.associated_comments.push(sibling.range().into());
-          current_node = sibling;
-          continue; // Continue the inner loop (i.e. evaluate next sibling)
+        } else if self._is_comment_safe_to_delete(&sibling, node, piranha_arguments, trailing) {
+          if !keep_comment_regexes.iter().any(|pattern| {
+            Regex::new(pattern)
+              .map(|re| re.is_match(sibling.utf8_text(code.as_bytes()).unwrap()))
+              .unwrap_or(false)
```
yuxincs, line 210: I'm just wondering if we should pre-compile the regexes instead of doing it every time. But I don't know rust enough to say whether that's worth it :)

PR #726, opened 2025-05-30
Releasing w/ Debian and select between PyPI or TestPyPI

dvmarcilio:
Adds:
- debian container for our linux release (to be tested once merged)
- workflow_dispatch to easily select between releasing to PyPI or TestPyPI

### DIFF
--- a/.github/workflows/polyglot_release.yml
+++ b/.github/workflows/polyglot_release.yml
@@ -3,15 +3,21 @@ on:
   push:
     tags:
       - '*'
+  workflow_dispatch:
+    inputs:
+      repository:
+        description: 'PyPI repository (pypi or testpypi)'
+        required: false
+        default: 'pypi'
 jobs:
   release:
     strategy:
       matrix:
         target: [linux-x86, linux-arm, macos-arm]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
-    # We need 20.04 due to glibc not being new enough on Debian/devpod
-    # It has glibc 2.28+ (manylinux_2_28), which is compatible with Debian 10: https://github.com/pypa/manylinux/tree/8e297ac3813be400a9a244a2828c18d3dd7e6969?tab=readme-ov-file#manylinux_2_28-almalinux-8-based
-    runs-on: ${{ fromJSON('{"linux-x86":"ubuntu-22.04","linux-arm":"ubuntu-latest","macos-arm":"macos-latest"}')[matrix.target] }}
+    runs-on: ${{ fromJSON('{"linux-x86":"ubuntu-latest","linux-arm":"ubuntu-latest","macos-arm":"macos-latest"}')[matrix.target] }}
+    # We need to pin a debian version for linux due to glibc not being enough on Debian used by CI and Devpods.
+    container: ${{ contains(matrix.target, 'linux-') && 'debian:11' || '' }}
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/polyglot_release.yml
```diff
@@ -3,15 +3,20 @@ on:
...
 jobs:
   release:
     strategy:
       matrix:
         target: [linux-x86, linux-arm, macos-arm]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
-    # We need 20.04 due to glibc not being new enough on Debian/devpod
-    # It has glibc 2.28+ (manylinux_2_28), which is compatible with Debian 10: https://github.com/pypa/manylinux/tree/8e297ac3813be400a9a244a2828c18d3dd7e6969?tab=readme-ov-file#manylinux_2_28-almalinux-8-based
     runs-on: ${{ fromJSON('{"linux-x86":"ubuntu-22.04","linux-arm":"ubuntu-latest","macos-arm":"macos-latest"}')[matrix.target] }}
+    container: ${{ contains(matrix.target, 'linux-') && 'debian:stretch' || '' }}
```
yuxincs, line 19: stretch is debian 9 IIRC, let's use debian 11?

.github/workflows/polyglot_release.yml
```diff
@@ -3,15 +3,20 @@ on:
...
+        default: 'pypi'
 jobs:
   release:
     strategy:
       matrix:
         target: [linux-x86, linux-arm, macos-arm]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
-    # We need 20.04 due to glibc not being new enough on Debian/devpod
-    # It has glibc 2.28+ (manylinux_2_28), which is compatible with Debian 10: https://github.com/pypa/manylinux/tree/8e297ac3813be400a9a244a2828c18d3dd7e6969?tab=readme-ov-file#manylinux_2_28-almalinux-8-based
     runs-on: ${{ fromJSON('{"linux-x86":"ubuntu-22.04","linux-arm":"ubuntu-latest","macos-arm":"macos-latest"}')[matrix.target] }}
```
yuxincs, line 18: If we are using container anyways, should we always use ubuntu-latest?

.github/workflows/polyglot_release.yml
yuxincs: Also, perhaps keep the comments here stating why we need to pin a debian version? (we had that, somehow missed it when doing OS upgrades)

.github/workflows/polyglot_release.yml
```diff
@@ -3,15 +3,20 @@ on:
...
+      repository:
+        description: 'PyPI repository (pypi or testpypi)'
+        required: false
+        default: 'pypi'
 jobs:
   release:
     strategy:
       matrix:
         target: [linux-x86, linux-arm, macos-arm]
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
```
yuxincs, line 17: python 3.8 is EoL'ed, perhaps we can take this opportunity to remove it and add 3.13? https://devguide.python.org/versions/

this can either be bundled in this PR or in a follow-up :)

.github/workflows/polyglot_release.yml
dvmarcilio: yes! updated

.github/workflows/polyglot_release.yml
dvmarcilio: good catch. also added back the comments

.github/workflows/polyglot_release.yml
dvmarcilio: will do this in the next PR once the release is working properly

yuxincs (approved):

PR #731, opened 2025-06-03
Include `polyglot_piranha.pyi` file in the source distribution

yuxincs:
This would make source distribution fully-typed as well even when users are building from source instead of getting prebuilt wheels.


danieltrt (approved):
cool!

PR #735, opened 2025-06-19
Add asterisk (*) support for zero-or-more matching in concrete syntax 

danieltrt:
## Summary
Added support for `:[var*]` pattern matching in concrete syntax templates, allowing zero-or-more node matching similar to regex quantifiers.

## Changes
- Added `MatchMode` enum to replace boolean parameters for cleaner API
- Implemented `ZeroPlus` mode in `handle_template_variable_matching` 
- Added unit tests for zero-or-more matching behavior

## Implementation Details
The asterisk implementation first attempts to match zero nodes, then falls back to one-or-more matching if zero match fails. This ensures `:[var*]` behaves as expected:
- `class Example { }` matches `class :[name] { :[body*] }` with empty body
- `class Example { int x = 1; }` matches same pattern with body containing the field

Like before, concrete syntax continues to be lazy during matching; (otherwise there is a state space explosion)


src/models/unit_tests/concrete_syntax_test.rs
```diff
@@ -180,3 +180,27 @@ fn test_match_anything() {
...
   );
 }
+
+#[test]
+fn test_asterisk_zero_or_more() {
+  // Test asterisk (zero or more) matching
+  run_test(
+    "class Example { }",
+    "class :[name] { :[body*] }",
+    1,
```
danieltrt, line 190:
```suggestion
    2,
```

norhh (approved):

PR #764, opened 2025-07-07
Add concrete syntax to the playground

danieltrt:
Add tree-sitter playground integration with concrete syntax support
    
- Add concrete syntax pattern matching to tree-sitter playground
- Implement WASM bindings for concrete syntax in browser
- Add GitHub Actions workflow for playground deployment
- Update build system to handle concrete syntax integration


dvmarcilio (approved):

crates/concrete-syntax/src/lib.rs
```diff
@@ -233,29 +257,3 @@ pub mod wasm {
     }
     "#;
 }
-
```
danieltrt, line 236: dead code

playground/tree-sitter/build.py
```diff
@@ -138,9 +183,13 @@ def copy_wasm_to_assets(wasm_file: Path, lang_name: str, assets_dir: Path) -> Pa
...
     with template_path.open("r") as inp, output_path.open("w") as out:
         content = inp.read()
-        languages = [
-            f'<option value="{lang}">{lang.title()}</option>' for lang in _LANGUAGES
-        ]
+        languages = []
+        for lang in _LANGUAGES:
+            # Set python as the default language
+            if lang == "python":
+                languages.append(f'<option value="{lang}" selected>{lang.title()}</option>')
```
yuxincs, line 190: hmmm, not opposing this, but why? IIRC we don't even support python in the playground

playground/tree-sitter/build.py
```diff
@@ -29,6 +29,51 @@ def run_command(cmd: List[str], cwd: Path = None) -> subprocess.CompletedProcess
...
+    print("Building concrete syntax WASM package...")
+    
+    # Check if WASM files need to be built (simple existence check)
+    wasm_files = [
+        pkg_web_dir / "concrete_syntax.js",
+        pkg_web_dir / "concrete_syntax_bg.wasm"
+    ]
+    
+    if all(f.exists() for f in wasm_files):
+        print("✓ Concrete syntax WASM files already exist, skipping build...")
```
yuxincs, line 46: Is this because we want to leverage cache? But it seems to have taken out in GH actions.

We should actually check the hash of the source files + Cargo locks to ensure we are not using an old version, but that might not be worth it, perhaps we just simply always build it?

crates/core/src/models/matches.rs
```diff
@@ -552,37 +543,22 @@ impl From<concrete_syntax::models::matches::Range> for Range {
   }
 }
 
-impl From<Range> for concrete_syntax::models::matches::Range {
-  fn from(range: Range) -> Self {
```
danieltrt, line 556: dead code

playground/tree-sitter/build.py
```diff
@@ -29,6 +29,51 @@ def run_command(cmd: List[str], cwd: Path = None) -> subprocess.CompletedProcess
...
+        ], cwd=concrete_syntax_dir)
+        
+        # Generate WASM bindings
+        run_command([
+            "wasm-pack", "build",
+            "--target", "web",
+            "--out-dir", "pkg-web", 
+            "--no-default-features",
+            "--features", "wasm"
+        ], cwd=concrete_syntax_dir)
```
yuxincs, line 65: Why not just call the `build-wasm.sh` file, is it because we want to control `--out-dir`?

playground/tree-sitter/build.py
```diff
@@ -175,8 +224,32 @@ def main():
     print("Instantiating index.html.template to dist directory...")
     dist_dir.mkdir(parents=True, exist_ok=True)
     instantiate_index_html(_HERE / "index.html.template", dist_dir / "index.html")
+    
+    # Copy concrete syntax integration files
+    print("Copying concrete syntax integration files...")
+    concrete_syntax_files = ["concrete-syntax.css", "concrete-syntax.js"]
```
yuxincs, line 230: Sorry, can we rename this file to `concrete-syntax-playground.js` or something similar? 

We have `concrete_syntax.js` which is a wrapper around the WASM, here we have another `concrete-syntax.js` which is gluing the logic to playground :)

yuxincs (approved):

PR #768, opened 2025-07-18
Add a Directory Recursive scope to Piranha

danieltrt:
DirectoryRecursive rules are global rules that essentially only apply to the directory from which they were triggered (and respective directory tree).

Added a new test case for this new scope

### DIFF
--- a/crates/core/src/models/rule.rs
+++ b/crates/core/src/models/rule.rs
@@ -13,7 +13,14 @@ Copyright (c) 2023 Uber Technologies, Inc.
 
 use std::collections::{HashMap, HashSet};
 use std::hash::{Hash, Hasher};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+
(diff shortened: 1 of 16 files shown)


crates/core/src/models/rule.rs
```diff
@@ -307,15 +314,23 @@ impl InstantiatedRule {
...
         if let Some(file_dir) = file_path.parent() {
           file_dir == scope_dir
         } else {
           false
         }
       }
+      DirectoryScope::DirectoryRecursive(scope_dir) => {
+        // DirectoryRecursive scope - check same directory or subdirectories
+        if let Some(file_dir) = file_path.parent() {
+          file_dir == scope_dir || file_dir.starts_with(scope_dir)
```
dvmarcilio, line 329: should we use a Path-like abstraction to check that the sub directories have `scope_dir` as parent? `starts_with()` seems a bit fragile and we can certainly come up with some counter examples where it would fail.

PR #772, opened 2025-07-23
Add .type constraint to cs

danieltrt:
This PR adds a new constraint type to the concrete syntax, letting users restrict the types of nodes they match against. This should lead to significant performance improvements at match time.

- Added three new E2E tests for the type matchers
- `root` is a special keyword for referring to the entire match, so we can constrain not just capture nodes but the full match itself

### DIFF
--- a/crates/core/src/models/filter.rs
+++ b/crates/core/src/models/filter.rs
@@ -17,6 +17,7 @@ use derive_builder::Builder;
 use getset::Getters;
 
 use itertools::Itertools;
+use log::debug;
 use pyo3::prelude::{pyclass, pymethods};
 
 use serde_derive::Deserialize;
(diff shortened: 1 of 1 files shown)


crates/concrete-syntax/src/models/concrete_syntax/interpreter.rs
```diff
@@ -470,6 +483,7 @@ fn create_empty_captured_node() -> CapturedNode {
       end_point: crate::models::matches::Point { row: 0, column: 0 },
     },
     text: String::new(),
+    node_type: "empty".to_string(),
```
yuxincs, line 486: Perhaps make this more verbose like `<empty-node-type-that-should-never-be-matched>`?

`empty` feels too generic and if we see this when debugging we might think it's a valid node type.

yuxincs (approved):

PR #779, opened 2025-10-06
Fixing bug after previous fix was not complete

danieltrt:
The previous [bug fix](https://github.com/uber/piranha/pull/778) was partial. It would work for single application of rules, but not until fix point because I forgot to add the edit to the queue. I've modified the test to reflect the correct behavior and fixed the bug.

### DIFF
--- a/tests/test_piranha.py
+++ b/tests/test_piranha.py
@@ -391,6 +391,8 @@ def test_match_only_parent_rule_bug_fix():
 if (obj.isLocEnabled() || x > 0) {
     // do something
     x = obj.isLocEnabled();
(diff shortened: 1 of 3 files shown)


tests/test_piranha.py
```diff
@@ -419,6 +422,7 @@ def test_match_only_parent_rule_bug_fix():
         name="logic_triggered_only_after_match_only_parent_rule",
         query='cs if (other_check) { }',
         replace_node="*",
+        is_seed_rule=False
```
danieltrt, line 425: This rule was triggering because `is_seed_rule` is True by default

